PR #4, opened 2019-09-04
Business Units initial commit

averbuks:
Initial commit for `organization-bootstrap` example `business-units`


organization-bootstrap/business-units/README.md
```diff
@@ -0,0 +1,102 @@
+# Business-units based organizational sample
+
+This sample creates an organizational layout with two folder levels, where the first level is usually mapped to one business unit (infra, data, analythics) and the second level represents enviroments (prod, test). It also sets up all prerequisites for automation (GCS state buckets, service accounts, etc.), and the correct roles on those to enforce separation of duties at the environment level.
```
ludoo: Typo on `analytics`.

organization-bootstrap/business-units/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+This layout is well suited for small and medium-sized infrastructures managed by a smal set of teams, where the complexity in application resource ownership and access roles is mostly dealt with at the project level, and/or in the individual services (GKE, Cloud SQL, etc.). Its simplicity also makes it a good starting point for more complex or specialized layouts.
```
ludoo: Typo on `small`

ludoo: We should change this paragraph to more closely match this sample, and remove the duplication with the following paragraph.

organization-bootstrap/business-units/main.tf
```diff
@@ -0,0 +1,162 @@
...
+
+# Business unit 1
+
+module "business-unit-1-folders-tree" {
+  source = "./modules/business-unit-folders-tree"
+
+  billing_account_id            = var.billing_account_id
+  tf_project_id                 = module.project-tf.project_id
+  top_level_folder_name         = var.business_unit_1_name
+  second_level_folders_names    = var.business_unit_1_envs
```
ludoo: I would prefer using a single variable for second level folders (environments), so they are the same for any business unit. This is how I have always seen this implemented for my customers.

organization-bootstrap/business-units/main.tf
```diff
@@ -0,0 +1,162 @@
...
+  generate_service_account_keys = var.generate_service_account_keys
+  gcs_location                  = var.gcs_location
+  organization_id               = var.organization_id
```
ludoo: I don't really like the way service accounts and bucket are managed implicitly by the module. It removes the documentation aspect which is really really important, and makes them more fragile. It also does not really match the actual use cases, where you have per-environment service accounts and buckets which are then used across all second-level folders.

Can we rework this so that the internal module is as lightweight as possible, just a simple wrapper to manage the folder hierarchy, and all the rest happens explicitly at the top level?

organization-bootstrap/business-units/main.tf
```diff
@@ -0,0 +1,162 @@
...
+module "bq-audit-export" {
+  source                   = "terraform-google-modules/log-export/google//modules/bigquery"
+  version                  = "3.0.0"
+  project_id               = module.project-audit.project_id
+  dataset_name             = "logs_audit_${replace(var.business_unit_1_name, "-", "_")}"
+  log_sink_writer_identity = module.log-sink-audit.writer_identity
+}
+
+# Audit log sink for business unit 1
+# set the organization as parent to export audit logs for all business units
```
ludoo: Audit log sink should be on the environment, not the bu. No idea really on how to do it unless the log module adds multiple support, so in this example it's probably best to switch to a single org-level export.

organization-bootstrap/business-units/modules/business-unit-folders-tree/README.md
```diff
@@ -0,0 +1,43 @@
+# Two-levels folders tree
```
ludoo: See comments above on how this module should be refactored imho.

PR #5, opened 2019-09-05
Organization sample: environments (Almanac bootstrap port)

ludoo:
Initial PR for comments


organization-bootstrap/environments/README.md
```diff
@@ -0,0 +1,55 @@
...
+![High-level diagram](diagram.png "High-level diagram")
+
+Refer to the [section-level README](../README.md) for general considerations about this type of samples, and usage instructions.
+
+## Managed resources and services
+
+This sample creates several distinct groups of resources:
+
+- one folder per environment
+- one top-level project to hold Terraform-related resources
```
morgante: I'd encourage putting these shared services into a "shared" folder.

organization-bootstrap/README.md
```diff
@@ -0,0 +1,46 @@
...
+- on the organization Organization Administrator, if Shared VPC needs to be managed by the automation service accounts
+
+### State
+
+This type of stage creates the prerequisites for Terraform automation including the GCS bucket used for its own remote state, so some care needs to be used when running it for the first time, when its GCS bucket has not yet been created. This is the sequence of steps to be followed to enable remote state management:
+
+- when running `apply` for the first time, the `backend.tf` file needs to be commented so local state is used
+- after the first `apply` has completed successfully, the comments in the `backend.tf` file need to be removed, and the GCS bucket name from the `bootstrap_tf_gcs_bucket` output added as a value for the `bucket` attribute
+  - in the `backedn.tf` file directly, or better
+  - in a separate `backend-config.hcl` file that will then be passed in the `-backend-config` option to every `init` call
```
morgante: I'd strongly discourage this approach, it's all too easy to forget to include this one time and thereby corrupt everything. Much better to be explicit and include it in backend.tf. Curious on why we'd recommend this?

organization-bootstrap/environments/main.tf
```diff
@@ -0,0 +1,159 @@
+# Copyright 2019 Google LLC
```
morgante: I'd encourage decoupling this file into projects.tf, folders.tf, export.tf, etc.

organization-bootstrap/environments/README.md
ludoo: All the actual examples I've had so far of using this with customers have only ever used a single project for shared resources, and having it at the top level makes it very clear that it's not tied to a single environment (or a single business unit in the 2-level example we'll add soon).
That said, I can see the merit of having a Shared folder if multiple shared projects are needed. My preference would be to leave as is as it reflects actual installs and is simpler, and add a comment in `main.tf` to suggest using a folder for more complex needs.

organization-bootstrap/README.md
ludoo: Explicit meaning you'd include the bucket name? And how do you then create the bucket?

We've always used this approach and never had an issue in the ~10 actual installs we've had so far (and the countless test installs we do for testing, prototyping, etc). I like it because you pay a minimal price once, and then are set with everything documented in code.

organization-bootstrap/README.md
```diff
@@ -26,16 +26,9 @@ This type of automation stage needs very specific IAM roles on the root node (or
...
 
-- when running `apply` for the first time, the `backend.tf` file needs to be commented so local state is used
-- after the first `apply` has completed successfully, the comments in the `backend.tf` file need to be removed, and the GCS bucket name from the `bootstrap_tf_gcs_bucket` output added as a value for the `bucket` attribute
-  - in the `backedn.tf` file directly, or better
-  - in a separate `backend-config.hcl` file that will then be passed in the `-backend-config` option to every `init` call
-- after setting the bucket name, re-run `init` with the above option if using a backend config file, then
-- run `apply` so that state is moved from the local file to the remote bucket
-
-The steps above only need to be performed once, after that the chicken-and-eggs problem is solved and state is remote for all subsequent runs.
+After the first successful `terraform apply`, edit `backend.tf.sample` and set the bucket name to the one shown in the `bootstrap_tf_gcs_bucket` output, then save and rename the file to `backend.tf`. Once that is done, run `terraform apply` again to transfer local state to the remote GCS bucket. From then on, state will be remote.
```
morgante: Instead of renaming the file, should create a copy (to make staying in sync with upstream easier).

organization-bootstrap/environments/main.tf
```diff
@@ -141,6 +141,9 @@ module "log-sink-audit" {
 ###############################################################################
 
 # shared resources project
+# you might want to create a 'Shared' folder if you need multiple projects for
```
morgante: Could we be prescriptive and default to this? It's not a big harm IMO.

organization-bootstrap/README.md
ludoo: done

organization-bootstrap/environments/main.tf
ludoo: I would prefer keeping this as simple as possible, and having it match he customer installs we did in the past. Is this something you strongly feel for?

morgante: I'd strongly prefer to have consistency. This 90% matches how we do things typically already, so I think it's valuable to just use it as a default.

ludoo: Ok, lemme then make that change (I also need to change the diagram), and switch to using comment for the simpler setup.

organization-bootstrap/environments/main.tf
ludoo: Thanks, but especially for simple environments like this, I much prefer having everything together to allow easy parsing.

organization-bootstrap/environments/main.tf
ludoo: After giving it a bit of thought, if a Shared folder is needed, and shared resources go beyond a simple project, then the best way to implement it is by adding a "shared" item to the environments list, and treating shared resources as one extra environment (with their own service account, bucket, Terraform setup, and pipeline).

If that is not needed (and it never was in all our automation engagements), a simple top-level project managed in the bootstrap stage is best, typically providing a simple way to manage GCR, KMS and a handful of buckets.

So I'm not adding an extra folder, but I'm detailing the above in the README.

PR #6, opened 2019-09-07
Turn on ci, add linting and boilerplate steps

ludoo:
Preliminary linting step.


ludoo:
Linting step uses the Docker tflint image.

ludoo:
Yes, let's experiment with it here for a while, taking it upstream if it's solid will be very easy. Thanks for chiming in so quickly!

ludoo:
Added a simple step to check boilerplate (and spotted 3 files missing it). The script is simple, but it reflects what our presubmit checks usually do, so it should be good enough (and less prone to breakage).

PR #7, opened 2019-09-10
Org environments module tests

ludoo:
To run tests manually, create a venv, install requirements, and launch pytest:

```
virtualenv -p python3 /some/path/venv
. /some/path/venv/activate
pip install tests/requirements.txt
pytest
```


ludoo:
This initial version should now be complete, with a cloud build config. Let's try to trigger it...

PR #13, opened 2019-10-13
Initial Shared VPC example

ludoo:
Alex, this has grown more than I anticipated, but I think it should be ready to go in as an initial version.

I added a set of test resources, with one instance in the host and GKE service project, and a COS MySQL instance in the GCE project that allows showing custom firewall rules, and how the COS MySQL module works with KMS.

The README is not hyper detailed, but should be enough for an initial version. I used gcpdraw for the diagram (the source is committe here along with the code), then tweaked its output slightly by hand

Let me know your thoughts, I will be OOO tomorrow and will only be in front of a screen late afternoon. If you have the time, could you try running it from scratch? I only ran it incrementally for the last few cycles, so it might need a few fixes. I will add tests if no major issues surface in your review.


infrastructure/shared-vpc/README.md
```diff
@@ -0,0 +1,61 @@
...
+- one internal Cloud DNS zone in the host project
+- one Cloud KMS keyring with one key in the host project
+- IAM roles to wire all the above resource together
+- one test instance in each project, with their associated DNS records
+
+## Test resources
+
+A set of test resources are included for convenience, as they facilitate experimenting with different networking configurations (firewall rules, external connectivity via VPN, etc.). They are encapsulated in the `test-resources.tf` file, and can be safely removed as a single unit.
+
+SSH access to instances is configured via [OS Login](https://cloud.google.com/compute/docs/oslogin/), except for the GKE project instance since [GKE nodes do not support OS Login](https://cloud.google.com/compute/docs/instances/managing-instance-access#limitations). To access the GKe instance, use a SSH key set at the project or instance level. External access is allowed via the default SSH rule created by the firewall module, and corresponding `ssh` tags on the instances.
```
averbuks: I think it should be `an SSH` instead of `a SSH`

infrastructure/shared-vpc/main.tf
```diff
@@ -0,0 +1,187 @@
...
+  private_visibility_config_networks = [module.net-vpc-host.network_self_link]
+  record_names                       = ["localhost"]
+  record_data                        = [{ rrdatas = "127.0.0.1", type = "A" }]
+}
+
+################################################################################
+#                                     KMS                                      #
+################################################################################
+
+module "host-kms" {
```
averbuks: When I'm trying to destroy the example:
```
Error: Instance cannot be destroyed

  on .terraform/modules/host-kms/terraform-google-modules-terraform-google-kms-4237838/main.tf line 27:
  27: resource "google_kms_crypto_key" "key" {

Resource module.host-kms.google_kms_crypto_key.key[0] has
lifecycle.prevent_destroy set, but the plan calls for this resource to be
destroyed. To avoid this error and continue with the plan, either disable
lifecycle.prevent_destroy or reduce the scope of the plan using the -target
flag.
```
I can see that `google-kms` module does not have it as an option, the questions are:
1. how to destroy?
2. Should we have some notes in readme about that?

infrastructure/shared-vpc/README.md
ludoo: Hah, I read it as "a Secure Shell key" while you read it as "an S-S-H key". Not sure which is the correct form. :)

infrastructure/shared-vpc/main.tf
ludoo: Thanks for catching this. I'll submit a PR to the KMS module to add support for setting `prevent_destroy` on keys. I seem to remember we have the same variable in the GCS module for buckets.

PR #14, opened 2019-10-13
Initial Hub and Spoke (VPNs) example (does not include tests, diagram and readme atm).

averbuks:
This is an initial PR for `net-hub-and-spoke` example based on Cloud VPNs
This example creates:
 - 3 VPCs: hub, spoke-1, spoke-2
 - 6 subnets: 2 subnets per VPC
 - VPNs and CloudRouters to connect Spoke VPCs to Hub VPC 
 - Private DNS Zone, Dummy Forwarding DNS Zone in Hub project and one peering DNS Zone per Spoke project.
 - Test resources: VM + VM dns entry for each spoke and hub subnets. 

It would be great to have any ideas/thoughts/concerns, especially about:
 - Should we create hub/spoke projects as well? 
 - Should we have Forwarding DNS Zone? 
 - Should we break down the `main.tf` to specific files like routers.tf, dns-zones.tf, vpns.tf?
 - Any other input appreciated.


ludoo:
Initial feedback is it looks really good. :) I wouldn't split into different files, as IMHO it makes finding resources harder. My approach is to treat TF modules similar to Python modules: up to a certain size, there's no issue keeping things together in a single file, unless you have a reason to (eg test resources).

infrastructure/net-hub-and-spoke/dns.tf
```diff
@@ -0,0 +1,68 @@
...
+  version = "~> 2.0"
+
+  project_id = var.hub_project_id
+  type       = "forwarding"
+  name       = "${var.forwarding_dns_zone_name}-hub-forwarding"
+  domain     = var.forwarding_dns_zone_domain
+
+  private_visibility_config_networks = [module.vpc-hub.network_self_link]
+  target_name_server_addresses       = var.forwarding_zone_server_addresses
+}
```
ludoo: Cam we add a comment here saying that if forwarding is needed from the spokes, additional peering zones should created?

infrastructure/net-hub-and-spoke/variables.tf
```diff
@@ -28,9 +28,9 @@ variable "prefix" {
   description = "Prefix for VPC names."
 }
 
-variable "hub_custom_route_advertisement" {
+variable "spoke_to_spoke_route_advertisement" {
```
ludoo: maybe `enable_cross_spoke_routing`? but this is also fine

infrastructure/net-hub-and-spoke/dns.tf
averbuks: It's addressed by adding peerings from spokes to hub forwarding zone.

ludoo: I mean, it needs additional peerings to the fwding zone, right? we might just add a comment saying so

PR #15, opened 2019-10-20
Minimal rewording of top-level README, initial infrastructure README

ludoo:
Infrastructure README needs more work, but at least i's not empty.


ludoo:
Reframe READMEs

PR #17, opened 2019-10-28
Foundations architecture diagrams

averbuks:
I was trying to create some "GCP Style" diagrams for foundations examples. I'm not sure that it's better than what we have right now, any feedback is appreciate.


ludoo:
I'm not sure either: the old ones are clearer, these are more instinctively recognizable as GCP. Maybe try another iteration on these to try and capture some of the clarity of the originals? Can you share the slides file with the diagrams?

PR #24, opened 2019-11-28
BigQuery Module

juliocc:
Opinionated BigQuery module with support for datasets, tables, views and dataset-level permissions.


ludoo:
Hey Julio, can you add type hints to variable definitions?

PR #26, opened 2020-01-10
Update foundations modules versions

ludoo:
Update folders, log sinks, service accounts versions to avoid errors with recent Terraform versions.


ludoo:
Tests for foundations are ok while infrastructure tests have errors, will address in a separate PR.

PR #29, opened 2020-01-14
VPN HA module

averbuks:
VPN HA module which supports
 - GCP to GCP HA VPN
 - GCP to On-prem HA VPN

### DIFF
--- a/modules/net-vpn-ha/README.md
+++ b/modules/net-vpn-ha/README.md
@@ -1,0 +1,157 @@
+# Cloud VPN Dynamic Module
+
+## Example
+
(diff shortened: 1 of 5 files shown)


ludoo:
Aleks, let's wait and merge all modules and samples together when they are all ready. WDYT?

averbuks:
> d samples together when they are all ready. WDYT?

Sure, that's why the PR is to `development`, but we can do straight to master once they are all ready.

averbuks:
Ohh, I did the PR to master, it's by mistake, intention was to create it to development branch

PR #51, opened 2020-03-20
Split COS to a generic module

juliocc:
This introduces a generic COS module and updates the compute-vm-cos-coredns to use it

### DIFF
--- a/modules/compute-vm-cos-coredns/main.tf
+++ b/modules/compute-vm-cos-coredns/main.tf
@@ -14,49 +14,9 @@
  * limitations under the License.
  */
 
-locals {
-  files = {
-    for path, data in var.files :
-    path => merge(data, {
-      attributes = (
-        data.attributes == null
-        ? { owner = "root", permissions = "0644" }
-        : data.attributes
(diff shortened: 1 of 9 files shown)


ludoo:
Ok, I am a bit ambivalent on this, so let me explain how I see it and then we can brainstorm.

In most cases, people will pass a pre-made cloud config for a number of different reasons:
- the container needs specific options, or pre/post scripts so the systemd unit is very specialized
- there is more than one container, so more than one systemd unit
- etc

In those cases, having variables for tcp ports, etc. is more a bother (it complicates the interface) than an advantage. People are already passing in the right config, so the only thing they  might need to leverage is a handful of template variables for things the module decides internally (full hostname, things like that).

In the above scenario, it probably makes sense to have a set of pre-made configs for coredns, lighttpd, mysql, squid, etc. that the user can invoke with one variable. Any further configuration from the user is accessible via files (with the possible exception of volumes, which can leverage a specific cloud init section).

Does the above make sense? What am I not seeing that would need the tcp etc. variables?

PR #52, opened 2020-03-20
business-units foundation example

averbuks:
- bu example update
- project module nonauth to additive renaming
- small fix in net-vpc module readme


ludoo:
Aleks, sent you #53 with my proposed changes for this.

PR #59, opened 2020-04-22
Update outputs for foundations/environments example

averbuks:
Outputs have been uncommented and updated


ludoo:
Nice catch, this completely slipped my radar. :)

PR #60, opened 2020-04-23
Add peering route configuration support for private clusters

ludoo:
Aleks, could you take a look?


ludoo:
my fault, I was passing the network self url :)

PR #61, opened 2020-04-25
Internal load balancer module

ludoo:
This is a first shot at an ILB module. It has a few caveats but seems to be fairly concise and couples well with the `compute-vm` module.

@averbuks @juliocc let me know what you think


ludoo:
Switched this PR to draft as I'm splitting out MIGs from `compute-vm` in a separate `compute-mig` module. This will give MIGs more flexibility, add support for optional autoscalers and healtch checks, and untie templates from MIGs allowing to manage versions.

PR #68, opened 2020-05-06
Adds clarification to gke-cluster module for node_config

geojaz:
I ended up in the `gke-cluster` module from [Shared VPC/GKE](https://github.com/terraform-google-modules/cloud-foundation-fabric/tree/master/infrastructure/shared-vpc-gke) because I wanted to configure various aspects of the node-pool. A note like this would have saved me having to figure out why we don't take node_config params to the `gke-cluster` module.


ludoo:
Makes sense, apologies for the confusion and thanks for the clarifying comment!

PR #70, opened 2020-05-10
Envoy with Traffic Director cloud-config

drebes:
This is a Terraform/COS implementation of [Configuring a single Compute Engine VM for Traffic Director](https://cloud.google.com/traffic-director/docs/set-up-gce-vms?hl=hr#per_host_config) using the fabric framework. The main difference here is that while there they intercept node-internal traffic, we intercept traffic sent to the node externally via port 80. This does not include the TD resources, the plan is to use this on an end-to-end example that exposes a fleet of these instances behind a L7 XLB to use TD advanced routing features later on. Potentially this could also replace the Squid example at one point.

I've generalized the COS cloud-config generation, not sure if including the symlink is the appropriate way for a module to reference another one.


ludoo:
Thanks for this. I'm going to merge, then fix the small check violations. Let me know if you think the generic module should be also used as a basis for the nginx and squid modules.

PR #74, opened 2020-05-12
adds project_id to external gateway

drebes:
External gateway is not using the provided project_id, but defaulting to the provider one. This fails if the provider does not specify a project ID.


ludoo:
Thanks for spotting this!

PR #76, opened 2020-05-14
Source repository module

juliocc:
Initial version of a module for managing Source Repositories


ludoo:
Can you also refer it from the top-level and module folder READMEs? And note it in the CHANGELOG. :+1:

PR #78, opened 2020-05-22
Correct the name used to reference the folders module

taylorza:


ludoo:
Thanks for spotting this!

PR #79, opened 2020-05-22
Support Cloud KMS on Compute-VM module +  Fixes

lcaggio:
* Support Cloud KMS for boot and attached disks
 * Fix instance group example
 * Fix instance template example

To support KMS on the boot disk I had to first create the disk and than use it as boot disk. Please let me know if you see a better way to support it.


ludoo:
The `boot_disk` block supports a `kms_key_self_link` attribute, I would much prefer keeping the module as-is without introducing another resource.

Let's pause for a second and try to design variables, what I would like is trying to isolate all key attributes in a single top-level variable: for regular users it would default to `null`, when encryption is needed it would be filled in and support both CMEK and CSEK.

Tests are failing BTW :)

lcaggio:
Refactored with a single top level variable. Tests not failing anymore :-)

ludoo:
Lorenzo, whenever variables or outputs change the README table needs to be updated by running the `tfdoc` tool in the top-level `tools` folder. I will do it now so don't worry.

PR #82, opened 2020-05-30
Better service account outputs in project module

ludoo:
This change rationalizes the way robot and default service accounts are exposed in the project module's outputs:

- all service accounts are exposed via a single `service_accounts` output
- the general/legacy service robot is exposed in `service_accounts.cloud_services`
- per-service robots are exposed in `service_accounts.robots`, eg `service_accounts.robots.container-engine`
- default service accounts are exposed in `service_accounts.default`, eg `service_accounts.default.compute`

Adding extra robot service accounts is done via the `local.service_accounts_robot_services` variable in the `service_accounts.tf` file which has been added in this PR.


ludoo:
@drebes given that you're OOO for a while I'm merging this. :) Feel free to bug me about it of course when you come back.

PR #83, opened 2020-05-31
Refactor service account module outputs

ludoo:
This updates a few leftovers from the old versions in the service account module, and hopefully makes outputs more resilient to corner cases like a service account being deleted outside of Terraform, where old outputs tended to raise errors.

The changes should be backward-compatible.


juliocc:
LGTM

PR #84, opened 2020-06-01
Added PD CSI Driver Addon support

tvvignesh:
Google supports CSI Driver for PD (https://cloud.google.com/kubernetes-engine/docs/how-to/gce-pd-csi-driver) which is also supported by TF Google Beta Provider (https://www.terraform.io/docs/providers/google/r/container_cluster.html#gce_persistent_disk_csi_driver_config)


ludoo:
Thanks for this PR!

PR #85, opened 2020-06-01
KMS module refactoring

ludoo:
This improves the KMS module in several ways

- keyring is now optional so that keys can be managed in an existing keyring
- keyring IAM
- specifying key purpose
- key labels
- outputs depend on relevant IAM bindings

It also removes the dual "protected" / "unprotected" key resources which were triggering spurious recreation in dependent resources, as reported by Lorenzo. A notice has been added in the README.

It goes without saying that this version is mostly incompatible with the previous one.

### DIFF
--- a/modules/kms/README.md
+++ b/modules/kms/README.md
@@ -1,6 +1,6 @@
 # Google KMS Module
 
-This module allows creating and managing KMS keys and IAM bindings at both the keyring and key level. An existing keyring can be used, or a new one can be created and managed by the module if needed.
+This module allows creating and managing KMS crypto keys and IAM bindings at both the keyring and crypto key level. An existing keyring can be used, or a new one can be created and managed by the module if needed.
 
 When using an existing keyring be mindful about applying IAM bindings, as all bindings used by this module are authoritative, and you might inadvertently override bindings managed by the keyring creator.
 
@@ -26,7 +26,7 @@ module "kms" {
 }
 ```
 
-### Keyring creation and key rotation and IAM roles
+### Keyring creation and crypto key rotation and IAM roles
 
 ```hcl
 module "kms" {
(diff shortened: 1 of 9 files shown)


modules/kms/README.md
```diff
@@ -1,40 +1,71 @@
 # Google KMS Module
 
-Simple Cloud KMS module that allows managing a keyring, zero or more keys in the keyring, and IAM role bindings on individual keys.
+This module allows creating and managing KMS keys and IAM bindings at both the keyring and key level. An existing keyring can be used, or a new one can be created and managed by the module if needed.
```
lcaggio, line 3: nit: I would change "keys" to "crypto keys" here and in other point of the module. I know that this also not straight forward in GCP documentation neither, but it would reduce confusion between crypto keys and keys version and it would be aligned with the TF resource.

modules/kms/README.md
```diff
@@ -1,40 +1,71 @@
...
 
-The `protected` flag in the `key_attributes` variable sets the `prevent_destroy` lifecycle argument on an a per-key basis.
+When using an existing keyring be mindful about applying IAM bindings, as all bindings used by this module are authoritative, and you might inadvertently override bindings managed by the keyring creator.
+
+## Protecting against destroy
+
+In this module **no lifecycle blocks are set on resources to prevent destroy**, in order to allow for experimentation and testing where rapid `apply`/`destroy` cycles are needed. If you plan on using this module to manage non-development resources, **clone it and uncomment the lifecycle blocks** found in `main.tf`.
```
lcaggio, line 9: I would mention that data / VM encrypted with a disabled/destroyed key won't be accessible once the crypto key / key version is deleted.

modules/kms/README.md
ludoo: done

modules/kms/README.md
ludoo: I would assume whoever uses KMS understands what happens once you delete a key? :)

PR #87, opened 2020-06-04
Add Cloud KMS  support to GCS module

lcaggio:
Add Cloud KMS  support to GCS module


ludoo:
@lcaggio I made a couple changes hope you won't mind. I will merge this, feel free to disagree and log an issue or ping me if you think it should be made differently. :)

PR #99, opened 2020-06-23
Make project creation optional in project module

ludoo:
This allows using the project module to configure and customize pre-existing projects, without having to import them manually in state. It introduces count in the project resource, which works as far as I can understand but might need some extra testing especially with destroy.

This also adds a variable to optionally configure project API activation, useful in the situations above where you're referencing a pre-existing project, and need to activate extra APIs but don't want them to disable any depend APIs that were previously enabled.


ludoo:
@averbuks @juliocc would you mind taking a look, as I would not want to introduce bugs in this module, which is used pretty much everywhere in our examples.

ludoo:
mergin as review requests got no traction :)

PR #100, opened 2020-06-23
Cloud Endpoints module

lcaggio:
Add Cloud Endpoint module.


modules/endpoint/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+  default     = []
+}
+
+variable "iam_members" {
+  description = "Authoritative for a given role. Updates the IAM policy to grant a role to a list of members. Other roles within the IAM policy for the instance are preserved."
+  type        = map(list(string))
+  default     = {}
+}
+
+variable "openapi_config" {
```
ludoo, line 35: I assume you can only specify a single API definition file, and it's not optional. If it's like that, would it make sense to change to a single variable

```hcl
variable "endpoint_config" {
  type = object({
    spec = string
    path = string
 }) 
```

Where `spec` can be either `openapi` or `grpc`. Or even simpler

```hcl
variable "endpoint_config" {
  type = object({
    grpc = bool
    path = string
 }) 
```

Where default is to use OpenAPI.

modules/endpoint/main.tf
```diff
@@ -0,0 +1,46 @@
...
+  project              = var.project_id
+  service_name         = var.service_name
+
+  openapi_config       = var.openapi_config != null ? file(var.openapi_config) : null
+  grpc_config          = var.grpc_config != null ? file(var.grpc_config) : null
+
+  protoc_output_base64 = var.protoc_output_base64 != null ? base64encode(file(var.protoc_output_base64)) : null
+}
+
+resource "google_project_service" "default" {
```
ludoo, line 33: this should be activated in the project module, non of our modules deal with service activation

modules/endpoint/main.tf
```diff
@@ -0,0 +1,46 @@
...
+  project                    = var.project_id
+  service                    = google_endpoints_service.default.service_name
+  disable_on_destroy         = true
+  disable_dependent_services = true
+}
+
+resource "google_endpoints_service_iam_binding" "default" {
+  for_each     = local.iam_roles_bindings
+
```
ludoo, line 42: blank line ;)

PR #101, opened 2020-06-24
Add support for vpc routes to compute-vm

ludoo:
This adds support for defining VPC routes in the `compute-vm` module. The rationale is large architectures where you need to split out separate parts in different root modules, and instances are managed in a separate root module from one of the VPC they use.

TL;DR if you have multiple VPCs and appliances connecting them, this might be allow to parcel out resources in separate root modules


juliocc:
@ludoo do you still want to go ahead with this?

PR #102, opened 2020-06-28
Cloud functions module

ludoo:


modules/cloud-function/README.md
```diff
@@ -0,0 +1,160 @@
...
+  }
+  trigger_config = {
+    event = "google.pubsub.topic.publish"
+    resource = local.my-topic
+    retry = null
+  }
+}
+```
+
+### Anonymous HTTP access
```
lcaggio, line 52: nit: I see above that you suggest handling IAM on the container project. BTW, since we handle in the module the Anonymous scenario, would you add an example to enable it for a specific user/service account?

modules/cloud-function/README.md
```diff
@@ -0,0 +1,160 @@
...
+  bundle_config = {
+    source_dir = "my-cf-source-folder
+    output_path = "bundle.zip"
+  }
+}
+```
+
+### Service account management
+
+To use a custom service account managed by the module, set `service_account_create` to `true` and leave `service_account` set to the `null` value (the default).
```
lcaggio, line 96: nit: set to the `null` value (the default) -> set to `null` (default value)

modules/cloud-function/README.md
```diff
@@ -0,0 +1,160 @@
+# Cloud Function Module
+
+Cloud Function management, with support for IAM roles and optional bucket creation.
+
+The GCS object used for deployment uses a hash of the bundle zip contents in its name, which ensures change tracking and avoids recreating the function if the GCS object is deleted and needs recreating.
+
+## TODO
+
+- [ ] add support for source repository
```
lcaggio, line 9: It would be nice to add support to vpc_connector. It can be done on a future iteration.

modules/cloud-function/variables.tf
```diff
@@ -0,0 +1,128 @@
...
+
+variable "project_id" {
+  description = "Project id used for all resources."
+  type        = string
+}
+
+variable "region" {
+  description = "Region used for all resources."
+  type        = string
+  default     = "us-central1"
```
lcaggio, line 100: nit: I would make "europe-west1" as default, to be consistent with all other modules default values.

modules/cloud-function/variables.tf
```diff
@@ -0,0 +1,128 @@
...
+  description = "Function trigger configuration. Leave null for HTTP trigger."
+  type = object({
+    event    = string
+    resource = string
+    retry    = bool
+  })
+  default = null
+}
+
+# variable "source_repository_url" {
```
lcaggio, line 125: Add TODO or remove this commented variable

modules/cloud-function/README.md
ludoo: Ah yes, I planned on adding support for ingress settings and vpc connector, then forgot. Let's merge this and I'll add in a second PR. Thanks for spotting this. :)

ludoo:
Thanks Lorenzo! I completely forgot to add the ingress / egress settings even though I planned on adding them, good catch! I will do a second PR soon.

PR #103, opened 2020-06-30
Invert the list for role/member mapping so that member is the key

vanessabodard-voi:
This PR is to propose a change in the syntax for the input of the module project for additive roles, by reversing the mapping on the member and not on the role.
 

If I want to check the permissions for user A :
Currently : I need to check all of the keys "role", and see if each of them contains user A.
Example with the current syntax :
```
module "project" {
  source         = "./modules/project"
  name           = var.project_id
  project_create = false

  iam_additive_roles = [
    "roles/editor",
    "roles/viewer",
    "roles/storage.objectCreator",
    "roles/pubsub.subscriber",
    "roles/storage.objectAdmin"
  ]

  iam_additive_members = {
    "roles/editor" = [
      "group:usergroup_gcp_admin@lemonadeinc.io"
    ],
    "roles/viewer" = [
      "group:usergroup_watermlon_experimentation@lemonadeinc.io"
    ],
    "roles/storage.objectViewer" = [
      "group:engineering@lemonadeinc.io"
    ],
    "roles/pubsub.suscriber" = [
      "group:engineering@lemonadeinc.io"
    ],
  }
}

```

By reversing : I directly look for the user A entry and have directly the list of all of the roles this user have.
See the "Minimal example with IAM additive roles" section added in the Readme


vanessabodard-voi:
Thx for the feedback @ludoo 
I have updated with iam_additive_members and the test suite in the last commit.

For the other modules, I found the following modules to be concerned by a change:
- folder-units
- folder
- kms
- net-vpc
- organization
- pubsub
- secret-manager
- service-directory
- source-repository

This list is meant to be exhaustive, so I could be wrong by forgetting some. Any feedback welcome.
I will make changes time-opportunistically.

ludoo:
Looks good! Please ping me when you manage to tackle all modules, or let me know if you find out you don't have time so I can do a few. :)

juliocc:
I can help too if needed :)

ludoo:
You probably already discovered it, but you can just point `tools/tfdoc/tfdoc.py` to the module folder to update the README (provided you have installed its Python requirements).

@juliocc let's split modules three way (Vanessa, you, me) so we get this done, WDYT?

vanessabodard-voi:
@ludoo , thx for the tip.
Yes, we can totally split the load.
I have numbered the list of modules, we can assign the following way:
- I take the modules that are multiple of 3,
- @juliocc takes the modules that are multiple of 3 +1
- @ludoo  takes the modules that are multiple of 3 +2

ludoo:
Let me make a different proposal as some modules are more complex than others and I'm more familiar with them (which is another way of saying that if they break, it will be my fault). :)

*vanessa*

- folder-units
- folder
- pubsub

*julio*

- secret-manager
- service-directory
- source-repository

*ludo*

- net-vpc
- kms
- organization

Ideally, we should all work in the same branch. What is the best way of doing that in your opinion?

vanessabodard-voi:
Do you have access to my forked repo [https://github.com/voi-oss/cloud-foundation-fabric](https://github.com/voi-oss/cloud-foundation-fabric)  ?

ludoo:
Nope:

```
ERROR: Permission to voi-oss/cloud-foundation-fabric.git denied to ludoo.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

vanessabodard-voi:
I invited both of you with write access on the fork (@ludoo and @juliocc ).

vanessabodard-voi:
I made a mistake with the previous list: none of the modules have additive bindings, except for organization (for which I commited the change some minutes ago).

Most of the modules only offer authoritative bindings, not additive.

juliocc:
That's right. None of mine have additive bindings yet. Do we want to add them?

ludoo:
Super! Can you just fix the merge conflict in your branch? Guess we're good to go then. :)

Did you also check if we're using additive roles in the end to end examples, or do you need us to check?

PR #107, opened 2020-07-02
Reserved External IP-Compute VM

sumeetoc:
This PR includes the resolution to a bug which happens to be while creating a compute engine with reserved External IP


ludoo:
Thanks for the fix!

PR #108, opened 2020-07-02
Shared VPC- net address

sumeetoc:
With two simple modifications, net-address module now supports reservation of Internal IP addresses in a Shared VPC architecture.

Modification:
1. In an existing variable(internal_addresses), another entity has been added "host_project". Accepted values "null" or "host_project_id"
2. Based on the entity module decides which project to use

### DIFF
--- a/modules/net-address/main.tf
+++ b/modules/net-address/main.tf
@@ -37,7 +37,7 @@ resource "google_compute_address" "internal" {
   description  = "Terraform managed."
   address_type = "INTERNAL"
   region       = each.value.region
-  subnetwork   = each.value.subnetwork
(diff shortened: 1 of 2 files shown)


ludoo:
What happens with the current module if you pass the netwrk self link? it should already work for shared vpc afaik

sumeetoc:
> What happens with the current module if you pass the shared vpc self link in the network variable? it should already work for shared vpc afaik

That is correct, it works! Seems like I missed few pointers while initial testing...

I'll go ahead and close my PR

ludoo:
Thanks for cross-checking. The point you raise is still valid and probably worth mentioning in the module's README file. Feel free to send a PR, or I'll try to get around to doing it today or tomorrow.

PR #110, opened 2020-07-04
Cloud Asset feed operations example

ludoo:
This PR adds a new top-level folder for operations examples, and the first example on using Cloud Asset feeds for instance tag policy enforcement.


ludoo:
@juliocc mind taking a look if you have a few minutes?

PR #126, opened 2020-08-28
Add the option to not create a DNS managed zone

vanessabodard-voi:
This is code to make it optional to create the managed zone in the DNS module.
With this, it would be possible to split-up the record-sets in multiple modules.


ludoo (approved):
Thanks for this (and the extra fix to the org module)!

ludoo:
Small ask: can you also update the top-level changelog?

PR #127, opened 2020-08-28
Add alias IP support in `compute-vm`

juliocc:


juliocc:
@ludoo can you review the interface? I tried to use the same names as the underlying blocks in [the GCP provider](https://www.terraform.io/docs/providers/google/r/compute_instance.html#alias_ip_range) but maybe we can come up with better names.

ludoo:
Hey! Just one ask: before releasing can you mark the change as incompatible in the CHANGELOG? Thanks for taking care of this! :)

ludoo (approved):

PR #131, opened 2020-09-01
Add support for internal address purpose (and possibly refactor variables)

ludoo:
This adds support for specifying internal address purposes, which is [only supported from version 3.28.0](https://github.com/hashicorp/terraform-provider-google-beta/blob/master/CHANGELOG.md#3280-june-29-2020) of the `google-beta` provider, so this requirement is also set in the module.

The module's interface has always been a bit ugly, and it's even uglier now that a new variable has been added. I would love breaking compatibility and putting all internal IP configuration options in a single `internal_address_config` variable. Upside is a cleaner interface more similar to other modules, downside is one needs to set all three configuration items (`null` can be used of course) even if a single one needs to be specified.

WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ All notable changes to this project will be documented in this file.
 
 - fix error in `project` module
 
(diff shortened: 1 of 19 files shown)


juliocc:
I like the internal_address_config idea but that will indeed break the interface. Your call ;)

In any case the PR LGTM.

PR #132, opened 2020-09-01
Make VPN Gateway creation optional for the net-vpn-ha module

averbuks:


modules/net-vpn-ha/README.md
```diff
@@ -146,16 +146,18 @@ module "vpn_ha" {
...
 | *tunnels* | VPN tunnel configurations, bgp_peer_options is usually null. | <code title="map&#40;object&#40;&#123;&#10;bgp_peer &#61; object&#40;&#123;&#10;address &#61; string&#10;asn     &#61; number&#10;&#125;&#41;&#10;bgp_peer_options &#61; object&#40;&#123;&#10;advertise_groups    &#61; list&#40;string&#41;&#10;advertise_ip_ranges &#61; map&#40;string&#41;&#10;advertise_mode      &#61; string&#10;route_priority      &#61; number&#10;&#125;&#41;&#10;bgp_session_range               &#61; string&#10;ike_version                     &#61; number&#10;vpn_gateway_interface           &#61; number&#10;peer_external_gateway_interface &#61; number&#10;shared_secret                   &#61; string&#10;&#125;&#41;&#41;">map(object({...}))</code> |  | <code title="">{}</code> |
+| *vpn_gateway* | HA VPN Gateway Self Link for using an existing HA VPN Gateway, leave empty if `vpn_gateway_create` is set to `true`. | <code title="">string</code> |  | <code title="">null</code> |
+| *vpn_gateway_create* | Create HA VPN Gateway. | <code title="">bool</code> |  | <code title="">true</code> |
```
ludoo, line 152: Aleks, I think we should keep only one of `vpn_gateway` and `vpn_create`, WDYT?

averbuks: Do you mean using `vpn_create` instead of `vpn_gateway_create`?. To my mind it's going to be confusing, since we are still creating VPN (tunnels), but using an existing gateway

ludoo: nono, I mean use `vpn_gateway` as a flag, and skip create entirely.

ludoo: @averbuks :)

averbuks: Ou yeah, that's how I started, but then I testes a scenario with a few VPNs in the same TF profile, so if the gateway is created by one of them, another cannot use it due to dynamicity and count, so double apply with a target required. I'm trying to avoid that, WDYT?

ludoo: Makes complete sense :)

ludoo (approved):

PR #133, opened 2020-09-01
Add retention policy in gcs module

vanessabodard-voi:
This PR is to add retention policy for the bucket.


ludoo:
Super, many thanks for this!

PR #134, opened 2020-09-03
Add bucket logging

vanessabodard-voi:
This PR is to add logging for buckets.

### DIFF
--- a/modules/gcs/main.tf
+++ b/modules/gcs/main.tf
@@ -38,7 +38,7 @@ locals {
   )
   kms_keys = { for name in var.names : name => lookup(var.encryption_keys, name, null) }
   retention_policy = { for name in var.names : name => lookup(var.retention_policies, name, null) }
-  logging = { for name in var.names : name => lookup(var.logging, name, null) }
+  logging_config = { for name in var.names : name => lookup(var.logging_config, name, null) }
 }
 
 resource "google_storage_bucket" "buckets" {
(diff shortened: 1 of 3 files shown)


modules/gcs/main.tf
```diff
@@ -72,6 +73,14 @@ resource "google_storage_bucket" "buckets" {
       is_locked = lookup(local.retention_policy[each.key], "is_locked", false)
     }
   }
+
+  dynamic logging {
+    for_each = local.logging[each.key] == null ? [] : [""]
+    content {
+      log_bucket = local.logging[each.key]["log_bucket"]
+      log_object_prefix = lookup(local.logging[each.key], "log_object_prefix", "${local.prefix}${lower(each.key)}" )
```
vanessabodard-voi, line 81: Not sure about this : maybe it is better to not set log_object_prefix (and then it will be computed to the bucket name), rather than putting the bucket name if log_object_prefix is not set. But I don't know what is the syntax to do this.

modules/gcs/variables.tf
```diff
@@ -89,3 +89,9 @@ variable "retention_policies" {
   type        = map(map(string))
   default     = {}
 }
+
+variable "logging" {
+  description = "Per-bucket logging."
+  type        = map(map(string))
```
ludoo, line 95: Can you narrow down the type? Probably

```hcl
type = map(object({
  log_bucket = string
  log_object_prefix = string
}))
```

I would also rename the variable to `logging_config` for conformity with other modules.

modules/gcs/main.tf
ludoo: I think this should either be a string if you need a specific prefix, or `null` to fallback to the default behaviour. So if you enforce the variable type (comment below), you just need to pass the value specified by the user, which can be `null`:

```hcl
log_object_prefix = local.logging[each.key]["log_object_prefix"]
```

ludoo (commented):
Looks good, with a couple comments!

modules/gcs/variables.tf
vanessabodard-voi: Renamed.

vanessabodard-voi:
Should I update also the syntax for retention policy ?
It could also be a map(object).

ludoo:
If you have the time, it would be great!

ludoo:
Thanks!

PR #135, opened 2020-09-14
Change bucket_policy_only into uniform_bucket_level_access in GCS module

vanessabodard-voi:
This PR is to reflect the deprecation of bucket_policy_only in favor of uniform_bucket_level_access in the terraform provider (https://github.com/hashicorp/terraform-provider-google/releases/tag/v3.38.0).

To stay consistent with the default behavior of the module, the default for uniform_bucket_level_access is set to `true` to override the default in storage API which is `false`.
I thought it was wise, but I could be wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 - add support for logging and better type for the `retention_policies` variable in `gcs` module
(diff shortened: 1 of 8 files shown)


ludoo:
Vanessa, thanks a lot for this. Please be patient, I was swamped but I'll try to approve today!

ludoo:
@vanessabodard-voi can you also update `modules/folders-unit/main.tf` on line 98, and add the change to CHANGELOG marking as incompatible? Then I'll merge, thanks a lot for this!

PR #140, opened 2020-09-23
provider version update

sumeetoc:
google-beta provider version 3.28.0 can no more be fetched by Terraform, updated the version to latest


ludoo:
Thanks!

PR #146, opened 2020-09-26
Updated the main.tf for bigquery-datasets for creating views

chinmaydorlikar:
Added the "depends_on" attribute for views to be created only after all the tables are created to avoid failure.
Please check terraform-google-modules#145
This should help the user avoid the error "The table xxxxxx does not exist when creating a view.


ludoo:
Thanks a lot for catching this. I will fix the (unrelated) failing test tomorrow and merge.

PR #153, opened 2020-11-06
Add support for Terraform 0.13 features

juliocc:
This is a major refactor adding support for terraform features introduced in version 0.13. 

The main changes include:
* Modules only create a single resource. Creating multiple resources in the module is no longer needed since Terraform 0.13 added for_each/count support to modules.
* Removed all `*_members` and `*_roles` variables and merged them into a single `iam` variable of the form { ROLE => [MEMBERS] }.
* Added basic validations to some variables.
* Added more tests


ludoo:
wooooohoooooo :)

ludoo (approved):

PR #154, opened 2020-11-06
Add lint step to check module READMEs

juliocc:


ludoo:
Sent you a PR with a slight refactor, sorry could not resist [#155 ]

Issue #62, opened 2020-04-28
Add support for auto-creating service account in GKE nodepool module


ludoo:
The GKE nodepool module should offer the option to auto-create the service account used for nodes, similarly to what the Compute VM module is doing. It should of course adopt the same interface for variables.

PR #156, opened 2020-11-07
Add support for internal service account to GKE nodepool module

ludoo:
This change adds support for an optional internally-manages service account to the GKE nodepool module like we have for the GCE module, fixing #62.

It also shortens the node configuration variables by renaming their prefix from `node_config_` to `node_` for readability, breaking compatibility with previous versions.

A module test has also been added to validate the service account / scopes change.


ludoo:
TH svpc/gke example probably needs fixing....

juliocc:
Nice. LGTM

juliocc (approved):

PR #157, opened 2020-11-07
Make examples in READMEs runnable and testable

juliocc:
This PR parses the READMEs for all the modules and runs them as tests. This ensures consistent examples that reflect the actual usage of the module.


ludoo:
wow!

juliocc:
61 new tests ;)

ludoo:
We need a new trigger, the CB run times out ;)

ludoo:
Can we add a trigger that filters to only kick in for changes in README files?

ludoo (approved):
this is actually great!

PR #159, opened 2020-11-07
Parallel tests

juliocc:
This PR changes the fixtures that run terraform to copy the fixture contents to a temporary directory before running init. This way we can run terraform tests in parallel avoiding lock conflicts.


ludoo:
@juliocc I'm closing this and we both got nowhere with parallel tests :(

PR #160, opened 2020-11-09
Revert iam_additive behaviour

ludoo:
The `iam_additive` variable behaviour has been reverted so it's keyed by roles, to make it more resilient with dynamic values. A new `iam_additive_members` has been added to preserve the previous use case.


modules/organization/main.tf
```diff
@@ -16,13 +16,17 @@
...
 locals {
   iam_additive_pairs = flatten([
-    for member, roles in var.iam_additive : [
-      for role in roles :
-      { role = role, member = member }
+    for role, members in var.iam_additive : [
+      for member in members : { role = role, member = member }
+    ]
+  ])
+  iam_additive_member_pairs = flatten([
```
juliocc, line 23: isn't [transpose()](https://www.terraform.io/docs/configuration/functions/transpose.html) what you want?

modules/project/main.tf
```diff
@@ -16,13 +16,17 @@
...
 locals {
   iam_additive_pairs = flatten([
-    for member, roles in var.iam_additive : [
-      for role in roles :
-      { role = role, member = member }
+    for role, members in var.iam_additive : [
+      for member in members : { role = role, member = member }
+    ]
+  ])
+  iam_additive_member_pairs = flatten([
```
juliocc, line 23: idem

juliocc (commented):
PTAL at comments but LGTM.

ludoo:
As discussed offline ;)

PR #166, opened 2020-11-16
Add boot_disk_kms_key to node_config

nikever:
Add the possibility to use a Customer Managed Encryption Key to encrypt the boot disk attached to each node in the node pool.

Node config reference: 
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#node_config

Cloud KMS reference: 
https://cloud.google.com/compute/docs/disks/customer-managed-encryption


ludoo:
the failing lint test is unrelated (firewall module, and it's my fault) ;)

ludoo:
Thanks a lot for the contribution, merging now.

PR #169, opened 2020-11-17
Added support for forwarding path attribute in Cloud DNS

manuuuel:
I updated the target server ipv4 list to a list of objects (with `ipv4_address` and `forwarding_path` attributes).

Example of usage:

```
forwarders = [
    {
      ipv4_address    = "10.103.195.124"
      forwarding_path = "private"
    },
   {
      ipv4_address    = "10.236.195.230"
      forwarding_path = "default"
    }
]
```


juliocc:
Nice. Can you make sure all tests pass before we merge?

ludoo:
I would prefer to use a map `ip address => path` instead of a list of maps. Less to type, more readable.

ludoo:
Closing as #171 does it slightly differently and also fixes examples and tests.

PR #173, opened 2020-11-23
Hierarchical firewall

### LINKED ISSUE #121: Add support for hierarchical firewalls
Probably in the org/folder modules.

https://www.terraform.io/docs/providers/google/r/compute_organization_security_policy_association.html https://www.terraform.io/docs/providers/google/r/compute_organization_security_policy.html https://www.terraform.io/docs/providers/google/r/compute_organization_security_policy_rule.html

juliocc:
Support for hierarchical firewall policies. Fixes #121.

Still WIP. There's a lot of duplicated code to create the policies in the `folder` and `organization` modules. Not sure if we should create a `firewall-policy` module and put the common code there, or just leave like it is now. Comments and feedback?


ludoo:
@drebes would also appreciate your opinions ;)

ludoo:
I'm not too worried about code duplication, we do the same with health checks and I always debated splitting out the module, but that would mean losing module independency as you then would always need two modules for a single feature. I'm +1 on this, and we can revisit later if we find a sane way of avoiding duplication.
Unless we want to explore using symlinks in modules, to refer to common files across modules, which could be an option. But again, we can brainstorm on this after this is merged.

PR #177, opened 2020-11-25
monitoring is automatically enabled when monitoring.writer is enabled

lgandras:


ludoo:
Can you also change the test in `tests/modules/gke_nodepool/test_plan.py:43`?

PR #178, opened 2020-12-04
Logging sinks and exclusions

### LINKED ISSUE #161: Add examples to log sink module documentation
Use the log sink configuration in the foundation examples as a starting point.

juliocc:
Add support for creating logging sinks and logging exclusions in the `project`, `folder` and `organization` modules

Fixes #161 and #163

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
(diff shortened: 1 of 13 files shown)


ludoo (approved):
Super nice. I guess dropping the existing sink module and refactoring examples to use what's added here, can be done in a separate PR.

Only trivial doubt: would `iam` be better than `grant`? I' fine with either of course.

ludoo:
@juliocc I'm merging so we can start working on removing the old module from the e2e examples. :)

PR #183, opened 2020-12-22
Iigeregi dns onprem - two regions

iigeregi:


juliocc:
Nice! can you make sure all tests pass before we merge? Ping me if you need help running the tests

@averbuks @ludoo can you take a look?

networking/onprem-google-access-dns/main.tf copy
```diff
@@ -0,0 +1,332 @@
+/**
```
ludoo, line 1: can you remove this?

networking/onprem-google-access-dns/variables.tf
```diff
@@ -18,16 +18,19 @@ variable "bgp_asn" {
   description = "BGP ASNs."
```
ludoo, line 18: I would redesign variables, using a single "regions" variable where key is the region name, values an object with parameters for each region (ASN, IP range, etc.).

Then parameterize main.tf so we create resources for 1-n regions.

ludoo (commented):
This is super nice. As I mentioned in a comment on the variables, I would parameterize this to accept 1-n regions.

And as Julio said, please fix tests. :) Ping us if you need help with those.

ludoo:
I'm closing this since Aleks sent a new revised PR, let me know if it needs to be reopened.

PR #184, opened 2021-01-08
Cloud DNS and Shared VPC

aurelienlegrand:
This PR is a module to allow application teams to have their own Cloud DNS configuration.
It does the following:

- Creates a GCP project (that will be dedicated to Cloud DNS) for each application team
- Creates an empty VPC in this project, enables required APIs
- Creates a Cloud DNS private zone for each application team
- Creates a Cloud DNS peering zone in the host project to each application team


juliocc:
Hew @aurelienlegrand, this is a really nice example.

Looking at the module itself, it seems it only creates a project, a VPC and a few Cloud DNS resources. Do you think it would make sense to convert the module to a e2e example and reuse the existing modules for those resources. Check out the examples in the networking or foundations folders.

@ludoo what do you think?

BTW ping me if you need help fixing those tests.

ludoo:
> Looking at the module itself, it seems it only creates a project, a VPC and a few Cloud DNS resources. Do you think it would make sense to convert the module to a e2e example and reuse the existing modules for those resources. Check out the examples in the networking or foundations folders.

I was chatting with Aurelien, and we were discussing the same thing. I think we're all aligned on moving this to the cloud operations folder, and switching to using modules internally. :+1:

ludoo:
Merging into a local branch so we can help adding tests, and with final tweaks. @aurelienlegrand can you then reopen a PR to merge from local branch to master?

PR #189, opened 2021-01-21
Fix link path from readme to modules / serverless - Cloud Functions

next-modnavarro:


ludoo:
Modesto, thanks for the fix!

PR #190, opened 2021-01-28
Adjusting second region for on-prem-in-a-box for testing more use cases

averbuks:


ludoo:
Wow, I'll have a look ASAP.

ludoo:
Aleks, I think you need to rebase this, I see mentions in the test errors of module names we no longer use.

ludoo:
@averbuks can we drop the initial PR that started this? BTW I fixed tests and as soon as they finish on my machine, will commit here.

ludoo (approved):

Issue #193, opened 2021-01-31
Add support for regional PDs


ludoo:
Support regional PD in `compute-vm` module.

PR #194, opened 2021-02-01
Add support for rePD and existing disk attachment to compute VM

ludoo:


ludoo:
Disk interface has changed, so some tests will probably fail. Other than that, this should be ready.

ludoo:
Fixes #193

PR #196, opened 2021-02-13
Changed parent folder from root_node to shared-folder into the audit-project example code.

danielmarzini:
Fix for the parent folder of the audit-project. it was set to root-node instead of shared-folder as indicated into the overview diagram.


ludoo:
Thanks for spoting this. :)

PR #201, opened 2021-02-24
Add support for essential contacts

juliocc:


ludoo (approved):
Nice, I did not even knew that was a thing. :)

Curious as to why dynamic block names are now quoted, is it a new syntax warning?

juliocc:
> Curious as to why dynamic block names are now quoted, is it a new syntax warning?
No idea. That's `terraform fmt` doing its thing, seems to be a [recent change](https://github.com/hashicorp/terraform/commit/ff0dbd621553e54fab0a55de564bd1a8b023dd69).

PR #208, opened 2021-03-14
Added FW-Yaml module for distributed-firewall network example.

averbuks:
Tests are on the way.

### DIFF
--- a/modules/net-vpc-firewall-yaml/README.md
+++ b/modules/net-vpc-firewall-yaml/README.md
@@ -92,7 +92,7 @@ lb-health-checks:
   - 35.191.0.0/16
   - 130.211.0.0/22
 
-# deny all eggress
+# deny all egress
 deny-all:
   deny:
   - ports: []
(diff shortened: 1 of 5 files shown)


ludoo:
@juliocc we need to find a way to make example tests work for this ;)

ludoo:
@averbuks one thing I'm a bit wary of with this is rule name clashes within the same project: what about using a format like `[vpc name]-[leaf directory name]-[name for the rule yaml block]`?

averbuks:
> @averbuks one thing I'm a bit wary of with this is rule name clashes within the same project: what about using a format like `[vpc name]-[leaf directory name]-[file name]-[direction]-[name for the rule yaml block]`?
 
@ludoo good catch, I will incorporate vpc-name, I also like taking directory names, not sure about file name though.

averbuks:
@ludoo I adjusted VPC name to the FW rule name, but decided to skip taking folder names for now. I would definitely do that if I'm a customer (so I can define a strict folder structure and naming), but let's keep it more generic for now. Someone can just define one yaml file in the same folder where main.tf resides and say `config_folder="./"`, so we have nothing to derive. 
WDYT?

ludoo (approved):
:+1:

PR #213, opened 2021-03-20
Add userinfo-email scope to GCE instance service account

ludoo:


juliocc:
LGTM

Just curious, when is userinfo.email needed?

ludoo:
From memory, it allows logging the service account email in audit logs etc, instead of the instance id.

Issue #219, opened 2021-04-01
Clarify vpc output key formats


LucaPrete:
The [current vpc output documentation](https://github.com/terraform-google-modules/cloud-foundation-fabric/tree/master/modules/net-vpc#outputs) doesn't specify the format of the key to pull data out of the subnet_regions, subnet_secondary_ranges and subnet_self_links maps.
It would be good for users to know the exact key.

ludoo:
Feel free to send a PR for the module's README Luca, there are examples in other modules, but still worth having in `net-vpc`. Good catch.

PR #220, opened 2021-04-01
[#219] Specify VPC module key output format in docs

LucaPrete:


ludoo:
Luca, the in/out tables are generated via `tools/tfdoc.py`, you cannot edit them by hand. I would just put this in the descriptive part of the README.

Issue #221, opened 2021-04-06
Module gke-cluster misses master_global_access_config option


LucaPrete:
The module cke-cluster misses the option private_cluster_config.master_global_access_config which allows to reach private master endpoints from other regions from where the masters live

ludoo:
Thanks for flagging, we did not yet update the module. Want to send a PR?

PR #222, opened 2021-04-06
[#221] Add master_global_access_config to gke module

LucaPrete:


ludoo:
Luca, can you check failing tests?

PR #224, opened 2021-04-07
[#223] Remove unused IAM policies from shared-vpc-gke example

LucaPrete:


LucaPrete:
@juliocc @ludoo I don't find a sense to those IAM bindings, but please let me know if I'm missing something :)

juliocc:
This is my understanding:
- roles/container.developer: required from the bastion to deploy to/query the cluster
- roles/logging.logWriter, roles/monitoring.metricWriter, roles/stackdriver.resourceMetadata.writer: required by the nodes to send metrics/logs to Cloud Ops

@ludoo can you confirm

ludoo:
> * roles/container.developer: required from the bastion to deploy to/query the cluster

yes

> * roles/logging.logWriter, roles/monitoring.metricWriter, roles/stackdriver.resourceMetadata.writer: required by the nodes to send metrics/logs to Cloud Ops

yes

ludoo:
But Luca's point is that those are just cloned from the gke service project, and rference the gke node SA, in the gce project so they can be safely removed.

LucaPrete:
Yes, indeed @juliocc I see the same roles configured in the GKE project, where insteadI think they make sense.
I also brought up a cluster without those roles and I see everything working correctly.

Issue #227, opened 2021-04-08
Adding short names for the GCS nomenclature instead of the complete region


chinmaydorlikar:
I have observed that in the case of regional buckets, the complete region is included in the name of the bucket 
(prefix-us-central1-suffix), this is due to the default logic that if the bucket is multi-regional then the regional prefix is simply us or eu which makes sense. But in the case of regional buckets, can we use short names of the regions?
This can certainly help:
`module "utils" {`
`  source  = "terraform-google-modules/utils/google"`
`  version = "~> 0.3"`
`  region  = "us-central1"`
`}`
Anything that's already stopping us from doing so?

ludoo:
As per comments in the PR.

PR #228, opened 2021-04-08
 Adding short names for the GCS nomenclature instead of the complete region 

chinmaydorlikar:
#227 Added the local variable for location_prefix for regional buckets which changes us-central1 to usc1 and likewise using terraform-google-modules/terraform-google-utils and retaining the original logic if the bucket is multi-regional.


ludoo:
Hey, thanks for the PR. Submodules are really, really an antipattern, and one of our main tenets is having no external dependencies in the core modules. I also don't see a reason to use short names, especially since most times you already have proper region names used for GCE; etc.

chinmaydorlikar:
Yeah no that makes complete sense. My idea was to beautify the nomenclature by sort of shortening the names as they appear. But yeah introducing an external dependency does complicate things. Thanks for the clarification!

PR #229, opened 2021-04-10
Add support for group-based IAM to resource management modules

ludoo:
This integrates customizations we're making for my current customer, where IAM at the resource hierarchy level is tightly coupled with groups, which are vetted by the security team as a unit, together with their associated roles and point of application (org, folder, project).

Directly supporting this model allows much better readability in the resulting Terraform code, and emphasises the documentation part of IaC, which is one of the core features we aim to support in Fabric.

The feature is implemented via authoritative bindings, as an integral part of the security approval is the assurance that no other bindings are applied outside of this process, but is implemented in a lightweight fashion so that it can be used together with the `iam` variable, to support mixed use-cases or exceptions to the rule. It's of course transparent, so it can be ignored if not needed.

Example usage from the org README:

```hcl
module "org" {
  source          = "./modules/organization"
  organization_id = "organizations/1234567890"
  group_iam       = {
    "cloud-owners@example.org" = ["roles/owner", "roles/projectCreator"]
  }
  iam             = {
    "roles/browser" = ["domain:example.org"]
  }
}
```

### DIFF
--- a/modules/folder/README.md
+++ b/modules/folder/README.md
@@ -161,7 +161,6 @@ module "folder2" {
 # tftest:modules=2:resources=6
 ```
 
-
 <!-- BEGIN TFDOC -->
 ## Variables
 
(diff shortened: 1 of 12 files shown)


ludoo:
Example usage from the org README:

```hcl
module "org" {
  source          = "./modules/organization"
  organization_id = "organizations/1234567890"
  group_iam       = {
    "cloud-owners@example.org" = ["roles/owner", "roles/projectCreator"]
  }
  iam             = {
    "roles/browser" = ["domain:example.org"]
  }
}
```

averbuks (approved):
Am I right that customers can skip using `group_iam` and just use `iam` for users, SAs, groups?

ludoo:
> Am I right that customers can skip using `group_iam` and just use `iam` for users, SAs, groups?

Yep, you can use `iam`, `group_iam` or both, the two are merged and used in `for_each` for the same resource.

juliocc:
Nice. LGTM

juliocc (approved):

PR #233, opened 2021-05-06
Make cluster creation optional in Shared VPC example

ludoo:
This allows quicker creation of the svpc example when a GKE cluster is not needed. I will use this in the upcoming OpenShift example.


ludoo:
@juliocc I'm merging this directly as it only changes a var and adds conditionals on a couple modules and resources. I tested it with the var set to both `true` and `false`.

PR #235, opened 2021-05-07
Add Support for Ingress Egress policies

lcaggio:


modules/vpc-sc/main.tf
```diff
@@ -89,6 +93,93 @@ resource "google_access_context_manager_service_perimeter" "standard" {
...
+
+                content {
+                  service_name = try(operations.value.service_name, null)
+
+                  dynamic "method_selectors" {
+                    for_each = try([operations.value.method_selectors], [])
+
+                    content {
+                      method     = try(method_selectors.value.method, null)
+                      permission = try(method_selectors.value.permission, null)
```
drebes, line 126: I was surprised by this. I see it supported on the provider resource, but it is not listed on the docs. Have you tested this?

modules/vpc-sc/main.tf
```diff
@@ -89,6 +93,93 @@ resource "google_access_context_manager_service_perimeter" "standard" {
...
+        content {
+          dynamic "egress_from" {
+            for_each = try(var.egress_policies[egress_policies.value].egress_from != null ? [""] : [], [])
+
+            content {
+              identity_type = try(var.egress_policies[egress_policies.value].egress_from.identity_type, null)
+              identities    = try(var.egress_policies[egress_policies.value].egress_from.identities, null)
+            }
+          }
+          dynamic "egress_to" {
```
drebes, line 109: Likewise, maybe skip the dynamic block.

modules/vpc-sc/main.tf
```diff
@@ -89,6 +93,93 @@ resource "google_access_context_manager_service_perimeter" "standard" {
...
+              resources = try(var.egress_policies[egress_policies.value].egress_to.resources, null)
+
+              dynamic "operations" {
+                for_each = try(var.egress_policies[egress_policies.value].egress_to.operations, [])
+
+                content {
+                  service_name = try(operations.value.service_name, null)
+
+                  dynamic "method_selectors" {
+                    for_each = try([operations.value.method_selectors], [])
```
drebes, line 122: Have you tried an example with two `method_selectors`? For me this only works if I remove the [] from here and add it to the variable definition.

modules/vpc-sc/main.tf
drebes: Found it on the docs, cool!

modules/vpc-sc/main.tf
```diff
@@ -89,6 +93,93 @@ resource "google_access_context_manager_service_perimeter" "standard" {
           allowed_services   = each.value.enforced_config.vpc_accessible_services
         }
       }
+
+      dynamic "egress_policies" {
+        for_each = try(local.perimeter_egress_policies_enforced[each.key] != null ? local.perimeter_egress_policies_enforced[each.key] : [], [])
+
+        content {
+          dynamic "egress_from" {
```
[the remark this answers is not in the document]
lcaggio: Those are not required over API, I did a test and it works. I'll keep it, we can change the behaviour in a future iteration.

modules/vpc-sc/main.tf
lcaggio: Those are not required over API, I did a test and it works. I'll keep it, we can change the behaviour in a future iteration

modules/vpc-sc/main.tf
lcaggio: you are right, fixed

drebes (approved):
LGTM

PR #241, opened 2021-05-17
Openshift changes

ludoo:
This PR does two things
- set the Shared VPC in the DNS private zone of the service project, similar to what we do for Cloud DNS in GKE
- set custom hostnames for bootstrap and master instances, since the FQDN is used to set the node name


ludoo:
@LucaPrete

PR #244, opened 2021-05-19
Added terraform.tfvars.sample

kumar-dhanagopal:
... to enable users to easily customize the configuration based on their requirements.

### DIFF
--- a/networking/hub-and-spoke-vpn/terraform.tfvars.sample
+++ b/networking/hub-and-spoke-vpn/terraform.tfvars.sample
@@ -1,0 +1,51 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


juliocc:
I'm not really convinced of the value of having a sample tfvars file. Anyone somewhat familiar with terraform should be able to create one from variables.tf.

That being said, if we do want to include it, I would leave out all the variables with default values, specially if the value in the sample tfvars file is the same as the default.

kumar-dhanagopal:
I agree that many users might be able to put together a tfvars file using variables.tf as a starting point. But I believe users will appreciate the convenience of a sample tfvars that they can quickly edit. 

I see your point about leaving out variables that have defaults already set in variables.tf. But again, when users start from a sample tfvars that includes all the declared variables, they won't need to hand-code any variables in the tfvars file, which is a convenience that some users might appreciate. I do recognize that this approach entails more maintenance overhead (for the code owners) to update 2 files when adding or changing variables.

ludoo:
+1 on Julio's coments, I also don't think there should be unnecessary duplication. Code in this repo should adhere to Terraform best practices and do the sensible thing.

kumar-dhanagopal:
Ok, Julio and Ludovico. The only variable that doesn't have a default in `variables.tf` is `project_id`. Providing a sample tfvars file with just one variable doesn't seem useful.

Issue #247, opened 2021-05-21
Call to function "list" in ../../modules/gke-cluster/main.tf failed


kumar-dhanagopal:
I was trying out the example in `.../networking/hub-and-spoke-peering` using Terraform (v0.15.3) from within Cloud Shell.

The `plan` command returned the following error:

```
Error: Error in function call
│
│   on ../../modules/gke-cluster/main.tf line 105, in resource "google_container_cluster" "cluster":
│  105:     for_each = length(var.master_authorized_ranges) == 0 ? [] : list(var.master_authorized_ranges)
│     ├────────────────
│     │ var.master_authorized_ranges is map of string with 3 elements
│
│ Call to function "list" failed: the "list" function was deprecated in Terraform v0.12 and is no longer available; use tolist([ ... ]) syntax to write a literal list.
```
Going by the last line in the error message, it appears that the error occurred because I'm using a version that's higher than 0.12.  But, interestingly, the same example worked fine when I tested from an *off-cloud* host that has Terraform v0.14.11.

ludoo:
We haven't yet verified and aligned code for terraform 0.15...

rosmo:
@kumar-dhanagopal can you try replacing the list with tolist()?

kumar-dhanagopal:
Yes, @rosmo . I was able to create an execution plan and apply it after changing `list()` to `tolist([...])`.

ludoo:
Thanks Taneli for the quick fix!

PR #248, opened 2021-05-22
Update to Terraform 0.15

ludoo:
This only fixes errors raised by tests, some sensitive output issues might not have been caught here and will be fixed as they surface during regular usage.


ludoo:
Fixes #247

PR #251, opened 2021-05-30
Data Foundations: reorder first level README and folder structure

ludoo:
This is a first pass for a proposal to simplify the file/folder structure, and make this example a better fit for Fabric. I also streamlined the top-level README.

@yorambenyaacov please check this out, and if it can work what other changes need to be made. Then we can start looking at each of the separate stages, and start reviewing the actual terraform resources.


yorambenyaacov (approved):
LGTM

PR #258, opened 2021-06-05
Refactor environment

ludoo:
@yorambenyaacov I refactored the first step, and there's quite a few things where you need to work on for the other steps:

- you are not using modules in the way they are designed, eg you use two project modules where you could use a single one, you compute the prefix while the same is already done internally, etc
- don't use random resources, we don't use them anywhere
- try to design variables for compactness, and to be ignored whenever possible for common use cases
- modules should refer to the source in the cloned repo
- variables should all live in a `variables.tf` file

The end result if you follow our design best practices, is that this step is super compact. Which makes me think that, once the other steps are refactored, you don't need three separate stages, but a single Terraform setup. Each step can then be a single `.tf` file in the same directory, and everything gets applied in a single go. Which would make this compliant with all our other examples.

Please look at this and let me know, but let me say I don't see many other ways of doing this, if you want it to live in this repo. :)


ludoo:
@lcaggio FYI

lcaggio:
First, thanks Ludo for taking the time to review this. I see great value to keep this in the Fabric repo, we would do all needed to have the data foundation here :-) 

I see your point, let me share few point based on experience with customers behind the way we structured the code to identify the best way to refactor phase 2 (./datasource/).

* 2 steps: This is to reflect the fact that team who is creating projects is not the one responsible to create resources. Having 2 steps you can interact with the 2 responsible teams easier and quickier.
* multiple files / variables: This is to address feedback we had from some customer to create sub-modules (I do not like that approach, but I know there are different approaches around). Having separate input / output / tf files let the user easily create the module. We can address this with visual blocks (the one with `###`). 

Having said that, I/we will work to refactor the second phase to fit with the repo ;-)

ludoo:
Thanks Lorenzo!

On the submodules / split variables: I think submodules would be really overkill, and go against the way we do things here, which is modify the original Fabric modules, or just use them as a source of copy/paste, and create customized one tailored to your needs. In that sense, I think splitting files would point customers in the wrong direction. It would also make it harder for "normal" users to compile tfvars files as they would need to open and check a ton of variable files, often with just one or two variables. And make it impossible to format variables tables in README files using `tfdoc`.

As for using Fabric modules, I would ask @yorambenyaacov to get familiar with them: as an example, using the `project` module twice, first to create the project and then for IAM, is really really unnecessary. Dependencies are also in most cases tracked by our modules outputs, so it's something you don't need to manually specify.

Let me know if this can be merged, or if you prefer to do your own refactor along the same lines!

yorambenyaacov:
@ludoo thanks for your feedback. Our first assumptions were different but the changes you are asking for looks OK. We will take care of it and do the required changes.

yorambenyaacov (approved):
LGTM

PR #260, opened 2021-06-10
Refactor resources step

ludoo:
@yorambenyaacov this refactors the resources step.

There were quite a few things to fix:

- consistency, not only with our other examples but also in naming variables in this module, etc
- proper use of modules (please, please, try to get familiar with how they work before writing examples)
- unnecessary cmplexity (eg in locals)
- Terraform syntax gotchas

FYI @lcaggio


yorambenyaacov:
LGTM

PR #264, opened 2021-06-14
Add Service Identity for Secret Manager

lcaggio:


ludoo (approved):

ludoo:
Thanks Lorenzo, looks super good. I would just add the comment we discussed on the new service identity resource, and let's hope we nailed the dependency issue in outputs. :)

PR #270, opened 2021-06-24
Create pubsub service identity if service is enabled

lcaggio:
Create pubsub service identity if service is enabled.

### DIFF
--- a/modules/project/service_accounts.tf
+++ b/modules/project/service_accounts.tf
@@ -39,6 +39,10 @@ locals {
     for service, name in local.service_accounts_robot_services :
     service => "${service == "bq" ? "bq" : "service"}-${local.project.number}@${name}.iam.gserviceaccount.com"
   }
+  jit_services = [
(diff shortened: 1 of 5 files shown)


modules/project/service_accounts.tf
```diff
@@ -54,10 +58,10 @@ data "google_bigquery_default_service_account" "bq_sa" {
 }
 
 # Secret Manager SA created just in time, we need to trigger the creation.
-resource "google_project_service_identity" "sm_sa" {
+resource "google_project_service_identity" "jit_si" {
+  for_each   = { for service in var.services : service => service if contains(local.jit_services, service) }
```
juliocc, line 62: what about
``` setintersection(var.services, local.jit_services)```

IMO that's s little easier to read

lcaggio: nice!

juliocc (approved):
LGTM

PR #271, opened 2021-06-25
Add support for CMEK keys in Data Foundation end to end example

lcaggio:
Add support for CMEK keys in Data Foundation end to end example.


ludoo:
Lore, did not have time for a review, but are keys completely optional? What happens if the user does not set the corresponding variable?

yorambenyaacov:
project-datamart in 01-environment is missing service_encryption_key_ids for bigtable

data-solutions/data-platform-foundations/01-environment/variables.tf
```diff
@@ -55,3 +55,17 @@ variable "service_account_names" {
...
+variable "service_encryption_key_ids" {
+  description = "Cloud KMS encryption key in {LOCATION => [KEY_URL]} format. Keys belong to existing project."
+  type = object({
+    regional      = string
+    multiregional = string
+    global        = string
+  })
+  default = {
+    regional      = null
+    multiregional = null
```
yorambenyaacov, line 68: the code is using only multiregional.
1. Do we need the other options?
2. If yes (I assume we need it), "how to use" example can be nice.

lcaggio: Regional key not in use at the moment, removed.

PR #275, opened 2021-07-08
Data Foundation: add vpc-sc support

lcaggio:
Implemented VPC-SC support. Since VPC-SC is handled at org level, no VPC-SC configuration handled in this script. The script take the VPC-SC perimeter configuration as prerequisite.


data-solutions/data-platform-foundations/01-environment/README.md
```diff
@@ -44,6 +44,7 @@ Once done testing, you can clean up resources by running `terraform destroy`.
...
 | *service_account_names* | Override this variable if you need non-standard names. | <code title="object&#40;&#123;&#10;main &#61; string&#10;&#125;&#41;">object({...})</code> |  | <code title="&#123;&#10;main &#61; &#34;data-platform-main&#34;&#10;&#125;">...</code> |
 | *service_encryption_key_ids* | Cloud KMS encryption key in {LOCATION => [KEY_URL]} format. Keys belong to existing project. | <code title="object&#40;&#123;&#10;multiregional &#61; string&#10;global        &#61; string&#10;&#125;&#41;">object({...})</code> |  | <code title="&#123;&#10;multiregional &#61; null&#10;global        &#61; null&#10;&#125;">...</code> |
+| *service_perimeter_standard* | VPC Service control standard perimeter name in the form of 'accessPolicies/ACCESS_POLICY_NAME/servicePerimeters/PERIMETER_NAME'. All projects will be added to the perimeter in enforced mode. | <code title="">string</code> |  | <code title="">null</code> |
```
ludoo, line 47: Where does the access policy come from here? Is there a way to offer optional creation? I would also explain this in the README, and provide quick info and examples.

data-solutions/data-platform-foundations/01-environment/main.tf
```diff
@@ -38,6 +38,9 @@ module "project-datamart" {
     bq      = [var.service_encryption_key_ids.multiregional]
     storage = [var.service_encryption_key_ids.multiregional]
   }
+  # If used, remember to uncomment 'lifecycle' block in the
```
ludoo, line 41: Is there a better way of doing this? At a minimum, it should be clearly detailed in the README.

ludoo (changes requested):

data-solutions/data-platform-foundations/01-environment/README.md
lcaggio: I would live the policy and perimeter creation as a prerequisite for this step (for demo purpose I would add in the future a zero step to create prerequisites like: VPC-SC perimeter, CMEK keyrings and key, ...). Added a note in the README with a command to retrieve the list of existing perimeter names.

data-solutions/data-platform-foundations/01-environment/main.tf
lcaggio: At the moment the default is to handle parimeter<->project mapping in the vpc-sc module. At the moment I do not see a better way to address this, added a comment in the README. In the future we may change the default behavior and set the default in the other way around.

ludoo (approved):

PR #277, opened 2021-07-18
Update ILBaNH example (all protocols, symmetric hashing, multi-zone)

ludoo:
Changes the ILB as next hop example so that

  - the README file mentions recent support for all protocols and symmetric hashing, and links to the official documentation
  - sets of instances are now created in two zones to better reflect real-world scenarios
  - gw configuration does not use `netplan` anymore to make sure network configuration is present at first boot
  - VM anf forwarding rules addresses, and basic commands to SSH to instances and monitor backed health are provided in the Terraform outputs


juliocc (approved):
LGTM

PR #280, opened 2021-07-20
Simplify new CF from onprem via PSC example

ludoo:
This simplifies the above example, to make it more similar to other examples that can use an existing project, like the ones in the cloud operations folder. It also adds support for PSC addresses to the net-address module.


juliocc:
LGTP

juliocc (approved):

networking/private-cloud-function-from-onprem/variables.tf
```diff
@@ -73,8 +26,34 @@ variable "ip_ranges" {
   }
 }
 
+variable "name" {
+  description = "Name used for new resources."
+  type        = string
+  default     = "psc-onprem"
```
andgandolfi, line 32: I suggest changing the default name, otherwise with the current value you would end up with VPCs named `psc-onprem-hub` and `psc-onprem-onprem` which I find a bit confusing

networking/private-cloud-function-from-onprem/main.tf
```diff
@@ -14,80 +14,60 @@
...
+  name       = "${var.name}-onprem"
   subnets = [
     {
       ip_cidr_range      = var.ip_ranges.onprem
-      name               = "${local.prefix}onprem"
+      name               = "${var.name}-onprem"
       region             = var.region
       secondary_ip_range = {}
     }
   ]
```
andgandolfi, line 54: I'd remove Private Google Access from this network, otherwise the function would be reachable even without the VPN

andgandolfi (commented):
Just a couple of points. Thanks Ludo for refactoring this!

Issue #211, opened 2021-03-18
Allow using gVNIC in compute-vm


juliocc:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#nic_type

PR #284, opened 2021-07-25
Allow using gVNIC in compute-vm (#211)

lgandras:
This only adds the parameter `nic_type` to compute-vm. I would consider creating a module for compute images in order to catch the issue that certain flags are already enabled in public images when enabling `GVNIC` as a guest os feature.

### DIFF
--- a/modules/compute-vm/README.md
+++ b/modules/compute-vm/README.md
@@ -24,7 +24,6 @@ module "simple-vm-example" {
     subnetwork = var.subnet.self_link
     nat        = false
     addresses  = null
-    alias_ips  = null
   }]
   service_account_create = true
 }
@@ -54,7 +53,6 @@ module "simple-vm-example" {
     subnetwork = var.subnet.self_link
     nat        = false
     addresses  = null
-    alias_ips  = null
   }]
   attached_disks = [{
     name        = "repd-1"
@@ -85,7 +83,6 @@ module "simple-vm-example" {
     subnetwork = var.subnet.self_link
     nat        = false
     addresses  = null
-    alias_ips  = null
   }]
   attached_disks = [{
     name        = "repd"
@@ -119,7 +116,6 @@ module "kms-vm-example" {
     subnetwork = var.subnet.self_link
     nat        = false
     addresses  = null
-    alias_ips  = null
   }]
   attached_disks = [
     {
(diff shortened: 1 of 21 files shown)


lgandras:
In order to merge underlying enabled `guest_os_features` of public images with the ones of the custom image, we need a patch: https://github.com/hashicorp/terraform-provider-google/issues/9630

modules/compute-vm/README.md
```diff
@@ -171,6 +175,60 @@ module "vm-with-alias-ips" {
...
+  }]
+  service_account_create = true
+  instance_count         = 3
+}
+# tftest:modules=1:resources=4
+```
+
+### Using gVNIC
+
+This example shows how to enable [gVNIC](https://cloud.google.com/compute/docs/networking/using-gvnic) on your VM by customizing a `cos` image. Given that gVNIC needs to be enabled as an instance configuration and as an guest os configuration, you'll need to supply a bootable disk with `guest_os_features=GVNIC`. `SEV_CAPABLE`, `UEFI_COMPATIBLE` and `VIRTIO_SCSI_MULTIQUEUE` are enabled implicitl in the `cos` image (this example).
```
juliocc, line 188: s/implicitl/implicitly/

juliocc (changes requested):

modules/compute-vm/README.md
juliocc: and s/as an guest os/a a guest OS/

ludoo (dismissed):
I'm not sure the `network_interfaces` variable is the right place for this, as it will force users to specify an extra bool every time. I have the same issue with alias ips and addresses BTW.

@juliocc WDYT of an extra variable, something like `network_options`, where users can specify just the overrides they need?

juliocc:
I was thinking the same thing.  I've been waiting for [optional attributes](https://www.terraform.io/docs/language/expressions/type-constraints.html#experimental-optional-object-type-attributes) to make the interface simpler but it's been stuck in experimental for a while.

IMO `network_interfaces` is the right place but I agree that it's becoming cumbersome to pass in a bunch of nulls. My problem with an override like `network_options` is that once optional attributes become stable, we'll probably end up reverting to the current interface.

ludoo:
> I was thinking the same thing. I've been waiting for [optional attributes](https://www.terraform.io/docs/language/expressions/type-constraints.html#experimental-optional-object-type-attributes) to make the interface simpler but it's been stuck in experimental for a while.

Once/if that lands, we'll have to go over each module I guess... :)

> IMO `network_interfaces` is the right place but I agree that it's becoming cumbersome to pass in a bunch of nulls. My problem with an override like `network_options` is that once optional attributes become stable, we'll probably end up reverting to the current interface.

Well, I would much rather have to do it twice (the second time with all other modules), than overload network interfaces for an unknown amount of time. WDYT? I can give it a shot and we can discuss on a PR.

juliocc:
SGTM. Lets give it a try.

juliocc:
@lgandras optional is nice but it's an experimental feature. I think the idea was to move the secondary ranges and nic type to a new variable.

lgandras:
You're right. Since `network_options` is a separate variable and `network_interfaces` is a list, should we make `network_options` a map indexed by interface name?

lgandras:
And sorry, by interface name I mean the interface index.

juliocc:
I see two options

Either
```hcl
module "simple-vm-example" {
  // ...
  network_interfaces = [{
    network    = var.vpc.self_link
    subnetwork = var.subnet.self_link
    nat        = false
    addresses  = null
  }]
  network_interface_options = {
    0 = {
      alias_ips = null,
      nic_type = "GVNIC"
    }
  }
}
```

or 
```hcl
module "simple-vm-example" {
  // ...
  network_interfaces = [{
    network    = var.vpc.self_link
    subnetwork = var.subnet.self_link
    nat        = false
    addresses  = null
  }]
  network_interface_type = {
    0 = "GVNIC"
  }
  network_interface_alias_ips = {
    0 = ...
  }
}
```

And maybe even `addresses` can be moved out of `network_interfaces` since it's not used that often.

@ludoo wdyt?

ludoo:
> @ludoo wdyt?

The first one limits proliferation of variables, but is less convenient if we force users to specify all parameters to change a single one. Not really sure which one is best, but I know I would like to move `addresses` out of the `network_interfaces` part.

juliocc:
I think the first option is a reasonable compromise at least until optional() becomes stable

lgandras:
Regarding the following error:
```
E       tftest.TerraformTestError: Error running command plan: 1  
E       Error: error retrieving image information: Get "https://compute.googleapis.com/compute/v1/projects/cos-cloud/global/images/family/cos-89-lts?alt=json&prettyPrint=false": private key should be a PEM or plain PKCS1 or PKCS8; parse error: asn1: syntax error: sequence truncated
E       
E         with data.google_compute_image.cos,
E         on main.tf line 1, in data "google_compute_image" "cos":
E          1: data "google_compute_image" "cos" {
```
The example in the README has a data resource. This one needs credentials to talk to GCP. I assume the error is because github is trying to authenticate using `.github/workflows/fake-key.json`. Should I just hardcode the `module.vm-with-gvnic.boot_disk.image` rather than referencing it from a data source?

juliocc:
yeah, just set `source-image` to `cos-89-lts/cos-cloud`.

juliocc (approved):

PR #285, opened 2021-07-26
Decentralized firewall management example added.

averbuks:


ludoo:
Hey Aleks, can you describe what this does? :)

ludoo:
@sruffilli FYI

averbuks:
> Hey Aleks, can you describe what this does? :)

Hey Ludo, it's a simple example showing how fw-yaml module can be used to organise decentralized fwl management. Example itself doesn't do a lot, but showing how an example of firewall factory repository structure. 
Module was updated as well to support multiple config directories.

ludoo:
> Hey Ludo, it's a simple example showing how fw-yaml module can be used to organise decentralized fwl management. Example itself doesn't do a lot, but showing how an example of firewall factory repository structure.

Awesome! I looped in Simone as he was also planning to add an example, based on a similar approach he started using with customers. I really really like the approach of using descriptive files to manage repetitive resources.

sruffilli:
Hey Aleks - I was heavily inspired by your approach and I have my own set of modules based on the same config mechanism.
I have a few additional features which I will be glad to contribute back to your module: 

- Using the directory structure to feed informations (e.g. directory names -> VPC, file names -> rules description)
- Template variables used to have a single, reusable definition of repetitive sources/targets (e.g. $ssh_server mapping to a list of cidrs or service accounts)

@ludoo as far as I'm concerned I'd rather contribute via separate PRs. Thoughts?

ludoo:
> @ludoo as far as I'm concerned I'd rather contribute via separate PRs. Thoughts?

:+1: I would really love a set of modules in their own folder (similar to what we're already doing for the COS modules), that showcase how to manage repetitive resources like fw rules, subnets, projects, etc.

averbuks:
> Hey Aleks - I was heavily inspired by your approach and I have my own set of modules based on the same config mechanism.
> I have a few additional features which I will be glad to contribute back to your module:
> 
> * Using the directory structure to feed informations (e.g. directory names -> VPC, file names -> rules description)
> * Template variables used to have a single, reusable definition of repetitive sources/targets (e.g. $ssh_server mapping to a list of cidrs or service accounts)
> 
> @ludoo as far as I'm concerned I'd rather contribute via separate PRs. Thoughts?

Awesome, yeah I would expect this as a separate PR. 
Some thoughts on your ideas:
 - feeding some info from the folder structure: I thought about that, but for this FW module you may have as many nested folders as you want, so you are flexible to organise rules for you team as you want it. It makes it difficult to feed values out, I also wanted to keep modules as simple as possible, of course customers would add more here based on their req. 
 - template variables is very good, wondering if we can implement it without overwhelming complexity.

ludoo (changes requested):
Aleks, can you add a simple module/resource count test, like for all the other end-to-end examples?

averbuks:
> Aleks, can you add a simple module/resource count test, like for all the other end-to-end examples?

I will

modules/net-vpc-firewall-yaml/README.md
```diff
@@ -4,47 +4,56 @@ This module allows creation and management of different types of firewall rules
...
-  source      = "./modules/net-vpc-firewall-yaml"
-  project_id  = "my-dev-project"
-  network     = "my-dev-network"
-  config_path = "./dev"
+  source = "./modules/net-vpc-firewall-yaml"
+  
+  project_id         = "my-dev-project"
+  network            = "my-dev-network"
+  config_directories = [
+    "./prod",
```
ludoo, line 35: Should this be `./dev`?

modules/net-vpc-firewall-yaml/README.md
```diff
@@ -4,47 +4,56 @@ This module allows creation and management of different types of firewall rules
 
 Yaml abstraction for FW rules can simplify users onboarding and also makes rules definition simpler and clearer comparing to HCL.
 
-Nested folder structure for yaml configurations is supported, which allows better and structured code management. 
+Nested folder structure for yaml configurations is supported, which allows better and structured code management for multiple teams and environments.
```
ludoo, line 7: This is nice, but I wonder if the `common` folder is taking the place of hierarchical firewall rules. Would it make sense to add one sentence explaining the rationale of using VPC rules for common instead of hierarchical, maybe citing some limitations of hierarchical firewalls?

ludoo (approved):
There might be one typo in the example (`./prod` instead of `./dev`), and I would maybe add a clarification on common rules vs hierarchical. Other than that :+1:

modules/net-vpc-firewall-yaml/README.md
averbuks: Fixed the readme, good catch. I will think about hierarchical rules

averbuks: @ludoo I have added a bit more of context around dep/prod/common folders with FW rules, also noted hierarchical FW rules.

modules/net-vpc-firewall-yaml/README.md
averbuks: Good point, done

averbuks: It's done in the example, not in the module itself

modules/net-vpc-firewall-yaml/README.md
averbuks: Noted hierarchical FW rules in the example readme, let me know if we want to duplicate it in the module readme as well.

PR #287, opened 2021-07-28
Add new variable for flexibility between project id & name

sly92:
Is this change independant enough ?

### DIFF
--- a/modules/project/main.tf
+++ b/modules/project/main.tf
@@ -74,7 +74,7 @@ locals {
     ]
   ])
 
-  name = var.descriptive_name != null ? var.descriptive_name : var.name
+  descriptive_name = var.descriptive_name != null ? var.descriptive_name : "${local.prefix}${var.name}"
 }
 
 
@@ -88,7 +88,7 @@ resource "google_project" "project" {
   org_id              = local.parent_type == "organizations" ? local.parent_id : null
   folder_id           = local.parent_type == "folders" ? local.parent_id : null
   project_id          = "${local.prefix}${var.name}"
-  name                = "${local.prefix}${local.name}"
+  name                = "${local.descriptive_name}"
   billing_account     = var.billing_account
   auto_create_network = var.auto_create_network
   labels              = var.labels
--- a/modules/project/main.tf
+++ b/modules/project/main.tf
@@ -74,7 +74,7 @@ locals {
     ]
   ])
 
-  temp_project_id = var.project_id != null ? var.project_id : var.name
+  name = var.descriptive_name != null ? var.descriptive_name : var.name
 }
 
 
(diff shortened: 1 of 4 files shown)


ludoo (changes requested):
we never use random ids in our modules for a variety of reasons, let's please keep this PR focuse don splitting id and name

sly92:
Ok I understand

ludoo:
Thanks for the change. I would much prefer not changing the main variable used for project id, as that would require changing every end to end example that uses the project module.

I would just add an optional `descriptive_name` variable, and use that to set the project name if not null.

sly92:
That is what I thought. Thanks I will do it

modules/project/main.tf
```diff
@@ -85,7 +88,7 @@ resource "google_project" "project" {
   org_id              = local.parent_type == "organizations" ? local.parent_id : null
   folder_id           = local.parent_type == "folders" ? local.parent_id : null
   project_id          = "${local.prefix}${var.name}"
-  name                = "${local.prefix}${var.name}"
+  name                = "${local.prefix}${local.name}"
```
ludoo, line 91: I would skip the prefix if user passes in `descriptive_name`.

modules/project/outputs.tf
```diff
@@ -16,7 +16,7 @@
 
 output "project_id" {
   description = "Project id."
-  value       = "${local.prefix}${var.name}"
+  value       = "${local.prefix}${local.temp_project_id}"
```
ludoo, line 19: This should not be changed as project ID will stay the same as before this PR.

ludoo (changes requested):

sly92 (commented):
Adding new local variable to allow having "project name" différent from the "project id"

sly92:
Do you know why the check failed ?

sly92:
yes but what specifically ? 
do you need me to change something?

juliocc:
The README.md for the project module is out-of date. You have to run the tfdoc script (in the tools folder) to update variables and ouputs.

sly92:
I've just created a CLA. 
Do you need me to push again ?

ludoo (approved):

sly92:
Having said that, thanks for all your helps

ludoo:
@sly92 it also needs a CLA fro your company address (sla*****.ext@or*****.com), apparently.

ludoo:
Ignore, the CLA bot is asleep, all looks good on our side. Thanks a lot for your contribution!

sly92:
Sorry the useless space, can I revert and correct this ?

ludoo:
what would you want to correct?

sly92:
just remove the useless linebreak

PR #290, opened 2021-08-02
Proposal, goreleaser integration as release tool

danielmarzini:
This PR is meant to integrate [goreleaser](https://goreleaser.com/) as releaser tool. 
I've already configured its configuration file, an example of what a release could appear with its integration would be like the screenshot attached.

Screenshot 2021-08-02 at 10 23 23

With goreleaser, the release process would change by just adding a new simple command, ex:
```
$  git tag -a v0.1.0 -m "First release"
$  goreleaser release --rm-dist 

```
By using the configuration attached into this PR, gorelease will create a digest of all the PR merged into the current release and will automatically create tar.gz and zip file for it.

@ludoo @juliocc wdyt about this? 

----- UPDATE

1. Added a new CloudBuild conf file (cloudbuild.release.yml) it can be bound to be triggered ONLY upon a new tag is pushed.

2. The CloudBuild conf DO REQUIRE to be completed during the setup of the trigger. The field versionName is projects/#######/secrets/github_token/versions/latest. It needs to be adjusted to reflect the secrets manager path that will contains the [GITHUB_TOKEN allowed to do the release](https://goreleaser.com/quick-start/).


ludoo:
Can we hook it up to Cloud Build to automatically release on specific tags?

danielmarzini:
> Can we hook it up to Cloud Build to automatically release on specific tags?

I saw that there's the cloud build integration, let me have a check on it.

danielmarzini:
@ludoo I do confirm that we can bind the Cloud Build integration with a "push to tag" trigger which invokes GoReleaser. I'll do some tests and I'll be back with the conclusion.

danielmarzini:
@ludoo I've finished the Cloud Build integration and updated this PR description to better indicate the process. wdyt?

rjerrems:
Just a drive by comment, most of the other CFT repos use this which is
pretty neat once you get your commits to conform to the standard
https://github.com/googleapis/release-please

bharathkkb:
Drive by comment as well +1ing release please as goreleaser is mainly for shipping go binaries.
release-please should be enabled for the whole org and all you need is a release-please yml like [this](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/.github/release-please.yml)

Advantages include 
- you don't need a GITHUB_TOKEN
- you don't need to tag manually, as [release PR](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/938) will be created which upon merge will create tag and release
- if you want to start tracking usage metadata via `provider_meta`, it can automatically update it

ludoo:
I'm actually for the KISS principle, and keeping the current fabric release process manual. @juliocc WDYT?

ludoo:
After chatting with Julio, we bot agree that

- releasing takes maybe 20s once or twice per month, so it's not a big deal
- we both dislike the chatty verbose CFT bot, so that's out of the question
- adding a tool would introduce friction, as we would need to either remember how it works, or manage its automation (which would eventually break, need upgrades, etc.)

The current workflow is working perfectly, let's close this and remember it if/when we become so large and complex, that we need to manage the bureaucracy via tooling.

Thanks anyway Daniel! ;)

rjerrems:
FTR the bot really stays out of the way most of the time and makes
releasing much easier having done releases manually before its
introduction, but you all do what works for you.

PR #291, opened 2021-08-04
Create VM Instances from machine images

dsivigli:
This module creates a list of VM from images specified in yaml files.


juliocc:
@dsivigli can you rebase your branch to master?

dsivigli:
> Lookig at the module I'm not sure what's the benefit of it. It's mostly feeding values from a set of yamls into `google_compute_instance_from_machine_image`.
> 
> An idea: would it make sense to extend `compute-vm` to support machine images?

I was thinking it could be used in a region migration contest with the list of machine images and new VPC and subnet provided in configuration files that could be generated from a script.

ludoo:
I'm also a bit unsure about the use case, I've never seen it, and I never had a customer extensively use machine images.

ludoo:
Domenica, let's close this for now and chat offline if you really feel this is a valid use case. Maybe let's sync in the future before working on a PR. ;)

PR #294, opened 2021-08-10
Compute service account email statically

juliocc:
Generate the service account output statically based on the service
account name and parent project. This allows, among other things, to use
service accounts as map keys (e.g. to be used in the `iam` argument in
other modules).

With this  PR, the example below works as expected:
```terraform
module "project" {
  source          = "./modules/project"
  billing_account = "XXXX"
  name            = "my-project"
  parent          = "folders/YYYY"
}

module "sa1" {
  source     = "./modules/iam-service-account"
  project_id = module.project.project_id
  name       = "newsa1"
}

module "folder" {
  source = "./modules/folder"
  parent = "folders/ZZZZ"
  name   = "new folder"
  iam = {
    (module.sa1.iam_email) = ["roles/owner"]
  }
}
```

### DIFF
--- a/modules/iam-service-account/main.tf
+++ b/modules/iam-service-account/main.tf
@@ -56,8 +56,10 @@ locals {
     ? google_service_account_key.key["1"]
     : map("", null)
   , {})
-  prefix             = var.prefix != null ? "${var.prefix}-" : ""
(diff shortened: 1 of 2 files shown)


ludoo (approved):
nice!!!

PR #297, opened 2021-08-12
Billing budget module

juliocc:


ludoo:
This is nice, I'm only unsure about modules wrapping a single resource, as that might lead us to having a ton of modules that just wrap provider resources.

Can we make it so the module also optionally manages internally the notification targets, at least for the most common use cases, similarly to what we do for VM service accounts, and other stuff?

ludoo (approved):

PR #301, opened 2021-09-01
Fix 'scheduled-asset-inventory-export-bq' example

lcaggio:


ludoo:
@lcaggio can you fix tests?

ludoo (approved):

PR #314, opened 2021-10-02
Refactor compute-vm to remove multiple instance support

ludoo:
This aligns the `compute-vm` module to most other modules, removing multiple instance support and allowing simple use of `for_each` at the module level.

Other end to end examples and modules have also been updated to use this version.

The main advantage of this change, apart from consistency and module-level `for_each`, is to drastically simplify the module's internals, thus allowing us to evolve it (e.g. with new PD features when they come out) and customers to fork it in a much easier way.

The change drawbacks are

- IG support works fine, but only really allows using 1 IG per zone (with `for_each`) with a single instance in it; not a big issue IMHO but let me know if you think otherwise
- the `internal_ip` output now triggers dynamic value error if used, e.g. to create DNS records; this is super weird as we did not have the issue before, and the new code is logically identical to the old; if you find a reason plz let me know, otherwise I think this is an acceptable drawback


modules/compute-vm/variables.tf
```diff
@@ -66,11 +64,17 @@ variable "boot_disk" {
   })
   default = {
     image = "projects/debian-cloud/global/images/family/debian-10"
```
juliocc, line 66: maybe update to debian 11?

juliocc (approved):
just one comment but LGTM

modules/compute-vm/variables.tf
ludoo: good point, done

PR #315, opened 2021-10-04
Improve firewall module

juliocc:
- Simplified code structure to a single resource block for custom rules
- Added new output with all the google_compute_firewall resources
- Allow using named ranges in custom firewall rules


modules/net-vpc-firewall/variables.tf
```diff
@@ -14,44 +14,16 @@
...
-  description = "List of IP CIDR ranges for tag-based HTTP rule, defaults to the health checkers ranges."
-  type        = list(string)
-  default     = ["35.191.0.0/16", "130.211.0.0/22", "209.85.152.0/22", "209.85.204.0/22"]
-}
-
-variable "https_source_ranges" {
-  description = "List of IP CIDR ranges for tag-based HTTPS rule, defaults to the health checkers ranges."
-  type        = list(string)
-  default     = ["35.191.0.0/16", "130.211.0.0/22", "209.85.152.0/22", "209.85.204.0/22"]
+variable "admin_ranges_enabled" {
```
ludoo, line 23: Can we remove this and use an empty list to prevent creation?

ludoo (approved):
we can maybe nuke the extra boolean var for admin ranges, WDYT?

modules/net-vpc-firewall/variables.tf
juliocc: We're already doing that

ludoo: then why do we still have that variable? :)

juliocc:
> we can maybe nuke the extra boolean var for admin ranges, WDYT?

Done

modules/net-vpc-firewall/variables.tf
juliocc: no idea, backwards compatibility ;)

PR #317, opened 2021-10-04
Make dns module resilient to dynamic values

ludoo:
This changes the type of the `recordsets` variable in the `dns` module, so that it can accept dynamic values in its `records` attribute, for each record.

It also implements some "magic" in records so that

- a fully qualified name like `foobar.example.org.` is passed in as-is (new behaviour)
- an empty name is passed in as the zone domain
- anything else gets the zone domain appended


modules/dns/main.tf
```diff
@@ -152,10 +153,18 @@ resource "google_dns_record_set" "cloud-static-records" {
...
   managed_zone = var.name
-  name         = each.value.name != "" ? "${each.value.name}.${var.domain}" : var.domain
-  type         = each.value.type
-  ttl          = each.value.ttl
-  rrdatas      = each.value.records
+  name = (
+    each.value.name == ""
+    ? var.domain
+    : (
+      substr(strrev(each.value.name), 0, 1) == "."
```
juliocc, line 160: substr(each.value, -1, 1)

ludoo:
nit: we might want to change `records` to `rrdata` to align with the underlying resource

PR #331, opened 2021-10-18
Feature/cloudsql backup configuration

caiottavares:
Hey all, 

I've added the additional variables to include all backup settings for cloudsql instance. Not sure if you wanted to tweak anything else or add extra validations, so please take a look and let me know.

Thank you!


juliocc:
I think you named the attribute to `retained_count` but are using `retention_count`. Let me know if you need help running the test suite locally so you can take a look at the error

caiotavaresdito:
> Caio, for the README vars/outputs tables it's easier if you generate them with the `tools/tfdoc.py` tool. Just pass it the path of the module that needs updating. (apologies if I'm repeating this from previous PRs)

Hi @ludoo thank you and no need to apologize. I've been using the tfdoc.py, maybe my commit message wasn't clear as it sounds like I changed it manually lol :D

PR #335, opened 2021-10-20
Added GCS notification support

caiottavares:
I've added the GCS notification support to the GCS module.
Please let me know your thoughts.

### DIFF
--- a/modules/gcs/README.md
+++ b/modules/gcs/README.md
@@ -104,7 +104,7 @@ module "bucket-gcs-notification" {
     custom_attributes = {}
   }
 }
-# tftest:modules=1:resources=2
+# tftest:modules=1:resources=4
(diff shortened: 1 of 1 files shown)


caiotavaresdito:
> The tests are failing because of the `google_storage_project_service_account` data source. We tend to avoid data sources inside modules because they break tests.
> 
> I think the best option is to assume (and document) that the SA exists before the module is called. This is not such a big deal because the `project` module already forces the [creation of it if the storage API is enabled](https://github.com/terraform-google-modules/cloud-foundation-fabric/blob/master/modules/project/service_accounts.tf#L49).

You're right @juliocc but what would you suggest in order to pass the tests? Should I add the note in the readme that the service account must exist? Thank you!

PR #336, opened 2021-10-21
Added daily tagging GitHub Action

rosmo:
Creates new tags in format `daily_2021.10.21`, but only if there were any action in the master in the last day.


rosmo:
> /gcbrun why didn't it run by itself?

It did, but it was missing the license boilerplate. Added now.

PR #351, opened 2021-10-31
Added cloud-run module

apichick:


modules/cloud-run/README.md
```diff
@@ -0,0 +1,166 @@
...
+    ports = null
+    resources = null
+    volume_mounts = null
+  }]
+  traffic = {
+    "blue" = 25
+    "green" = 75
+  }
+}
+# tftest:skip
```
ludoo, line 32: Can you make sure documentation tests run?

modules/cloud-run/README.md
```diff
@@ -0,0 +1,166 @@
...
+  ]
+}
+# tftest:skip
+```
+
+### Eventarc trigger (Audit logs)
+
+This deploys a Cloud Run service that will be triggered when specific log events are written to Google Cloud audit logs.
+
+module "cloud_run" {
```
ludoo, line 66: please add quotes, for formatting and example testing

modules/cloud-run/variables.tf
```diff
@@ -0,0 +1,152 @@
...
+    }))
+    volume_mounts = list(object({
+      name       = string
+      mount_path = string
+    }))
+  }))
+}
+
+variable "iam" {
+  description = "IAM bindings for topic in {ROLE => [MEMBERS]} format."
```
ludoo, line 61: topic is probably a copy/paste leftover

ludoo (changes requested):
Thanks a lot for this Miren! Can you enable tests for doc examples by changing the `# tftest:skip` line to the number of modules and resources used in the example?

modules/cloud-run/README.md
```diff
@@ -0,0 +1,166 @@
...
+  source     = "../../modules/cloud-run"
+  project_id = "my-project"
+  name       = "hello"
+  containers = [{
+    image   = "us-docker.pkg.dev/cloudrun/container/hello"
+    command = null
```
juliocc, line 72: Most of the examples only use `image`. Maybe we can have everything except `image` in a `extra_options` field?

juliocc: And we probably want to have one example showing how to use all those fields.

modules/cloud-run/main.tf
```diff
@@ -0,0 +1,212 @@
...
+resource "google_cloud_run_service" "service" {
+  provider = google-beta
+  project  = var.project_id
+  location = var.region
+  name     = "${local.prefix}${var.name}"
+
+  template {
+    spec {
+      dynamic "containers" {
+        for_each = var.containers == null ? [] : var.containers
```
juliocc, line 59: Does this work if you have more than one container?

modules/cloud-run/variables.tf
```diff
@@ -0,0 +1,152 @@
...
+      limits = object({
+        cpu    = string
+        memory = string
+      })
+      requests = object({
+        cpu    = string
+        memory = string
+      })
+    })
+    ports = list(object({
```
juliocc, line 48: Would it make sense to have this as name => {protocol, container_port}?

modules/cloud-run/variables.tf
```diff
@@ -0,0 +1,152 @@
...
+      name           = string
+      protocol       = string
+      container_port = string
+    }))
+    volume_mounts = list(object({
```
juliocc, line 53: Would it make sense to have this as name => mount_path?

juliocc (commented):
very nice. left a few comments

modules/cloud-run/variables.tf
apichick: According to the APi all three are optional that is why I put 3
https://cloud.google.com/run/docs/reference/rest/Shared.Types/Container#ContainerPort

modules/cloud-run/main.tf
apichick: https://cloud.google.com/run/docs/reference/rest/v1/RevisionSpec According to the API you can pass a list, although I am not sure if it works as in the UI they only allow you to specify one container. For compatibility with the API I set it as an array.

modules/cloud-run/variables.tf
apichick: Changing it...

modules/cloud-run/variables.tf
apichick: Yes, changing it...

modules/cloud-run/README.md
apichick: Done

modules/cloud-run/variables.tf
apichick: Done

modules/cloud-run/README.md
apichick: I'm not sure about the extra_options thing, because the moment they need to add any of the other params they are in a similar position with extra options. I have added more examples, have a look.

modules/cloud-run/main.tf
apichick: Fixed

modules/cloud-run/variables.tf
apichick: Changed

modules/cloud-run/README.md
apichick: Done

modules/cloud-run/README.md
ludoo: +1 on Julio's suggestion; maybe only use image / ports / mounts / options?

modules/cloud-run/variables.tf
ludoo: I would skip protocol, as seems to just be a placeholder for future features: `Protocol for port. Must be "TCP". Defaults to "TCP"`.

modules/cloud-run/variables.tf
```diff
@@ -0,0 +1,149 @@
...
+  }))
+  default = null
+}
+
+variable "vpc_connector_config" {
+  description = "VPC connector configuration. Set `create_config` attributes to trigger creation."
+  type = object({
+    egress_settings = string
+    name            = string
+    ip_cidr_range   = string
```
ludoo, line 145: I remember discussions with you the fact the `ip_cidr_range` and `network` might be dynamic, passed in from other modules' outputs, and the plan to split this into two variables to avoid issues.

Can we split this into two variables, somthing like a boolean `vpc_connector_create` which you then use in `count` on the resource, and `vpc_connector_config` which has the actual variables? I would also skip `name` and just use the top-level name.

apichick: split done

modules/cloud-run/variables.tf
apichick: done

modules/cloud-run/README.md
apichick: separated command, args, env and env_from to a separate options var

ludoo (approved):

PR #352, opened 2021-11-01
Project module support for skip_delete flag on google_project resource

christopher-avila-ditoweb:
We have a use case where we want to create short-lived projects with this module but did not want the project to be removed when cleaning up resources. Since the google_project resource had an optional skip_delete flag, it felt to me that maintaining a fork was overkill for such a small change and the cleanest solution was to add it to the repository with a default that would allow the modules current behavior to remain the same while optionally adding the desired functionality.


christopher-avila-ditoweb:
@googlebot I signed it!

modules/project/variables.tf
```diff
@@ -162,6 +162,12 @@ variable "project_create" {
   default     = true
 }
 
+variable "project_skip_delete" {
```
juliocc, line 165: We're already in the project module, so the "project" prefix seems redundant. Can we call it just `skip_delete`?

juliocc (changes requested):
Nice! Left one small comment, otherwise LGTM

juliocc (approved):

juliocc:
Thanks for your contribution!

Issue #354, opened 2021-11-03
Update KMS module to support GCP provider version 4


juliocc:
Version 4 of the GCP provider removed `self_link` field from `google_kms_crypto_key` and `google_kms_key_ring`[1] breaking the `kms` module

[1] [self_link 4.0.0 changes](https://github.com/hashicorp/terraform-provider-google/pull/10424)

PR #355, opened 2021-11-03
Use the same versions file everywhere, pin to tf 1.0+ provider 4.0+

ludoo:


ludoo:
One alternative is to just inject `versions.tf` during tests, tftest already supports it and we would only need to change the pytest fixtures we use for plan.

cloud-operations/dns-fine-grained-iam/versions.tf
```diff
@@ -12,9 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-provider "google" {
-  impersonate_service_account = "data-platform-main@${var.project_ids.services}.iam.gserviceaccount.com"
```
sruffilli, line 16: Is it ok to lose this?

cloud-operations/dns-fine-grained-iam/versions.tf
```diff
@@ -12,9 +12,18 @@
...
+      version = ">= 4.0.0"
+    }
+    google-beta = {
+      source  = "hashicorp/google-beta"
+      version = ">= 4.0.0"
+    }
+  }
 }
-provider "google-beta" {
-  impersonate_service_account = "data-platform-main@${var.project_ids.services}.iam.gserviceaccount.com"
```
sruffilli, line 19: Same as above

cloud-operations/dns-fine-grained-iam/versions.tf
ludoo: ah no, good catch

ludoo:
Fixes #354

juliocc (approved):

PR #358, opened 2021-11-06
Support for consistent function bundles

lgandras:


modules/cloud-function/main.tf
```diff
@@ -137,6 +137,8 @@ data "archive_file" "bundle" {
     ? "/tmp/bundle.zip"
     : var.bundle_config.output_path
   )
+  output_file_mode = var.bundle_config.output_file_mode
```
juliocc, line 140: Do we need a variable for `output_file_mode`. Would it make more sense to just use 0666?

juliocc (approved):

juliocc:
Thanks @lgandras !

PR #359, opened 2021-11-08
Allow custom and predefined roles in bq dataset access rules

### LINKED ISSUE #356: BigQuery dataset: support for role/... access format
Thank you for the modules and the ease of management they provide.

Is there a good reason why roles available in "_access = {}_" are only limited to 'OWNER', 'READER', 'WRITER'?
We can grant other roles to datasets using direct "_google_bigquery_dataset_access_" resource block, like "_roles/bigquery.user_"?

Could the module handle both 'OWNER', 'READER', 'WRITER' and the "roles/..." format by internally converting "roles/bigquery.dataOwner", "roles/bigquery.dataViewer",  "roles/bigquery.dataEditor" to the legacy format and leaving others alone?

I think this is how "terraform-google-modules/bigquery/google" handles this.

Hmmm, may need to ~~merge~~concat members of _LEGACY_ and _roles/..._ for those 3, depending on how it's handled in the module.

juliocc:
Fixes #356


ludoo (approved):
As discussed in chat with Julio.

PR #360, opened 2021-11-14
added depends on topic iam binding

caiotavaresdito:
I've run into scenarios where Terraform attempts to create the gcs notification resource before the topic IAM binding and hence run into 403 issues. I've added an explicit depends on to avoid the scenario.


ludoo (approved):
Thanks!

PR #366, opened 2021-11-23
exclusions for audit in locals in logging_sinks.audit-logs now

skalolazka:
If the lines 132-147 in the foundations/environments/main.tf got uncommented to create audit log sinks, Terraform used to fail since the "exclusions" variable has to be not on the level it's currently specified it - I moved it to locals.tf to logging_sinks.audit-logs.exclusions where it belongs.


juliocc:
Thanks for the PR. 

LGTM but can you fix the linting error in `locals.tf`? I think just aligning the '=' with the line should fix it.

skalolazka:
Hi! Thanks, I've done that, although I somehow can't see if the linting ran through after that or how to trigger it.

juliocc (approved):

skalolazka:
Thank you!!

PR #367, opened 2021-11-23
Added support for Stateful Managed Instance Groups

stenio123:
Supporting both stateful disks for all instances or for specific instances of the Managed Instance Group as documented [here](https://cloud.google.com/compute/docs/instance-groups/configuring-stateful-disks-in-migs)


caiotavaresdito:
Hi @stenio123, thank you for your contribution. It looks like the build fail due Terraform linting (fmt) and also the variables.tf should be sorted alphabetically.

modules/compute-mig/variables.tf
```diff
@@ -66,11 +66,29 @@ variable "location" {
   type        = string
 }
 
+variable "minimal_action" {
```
ludoo, line 69: Can we bundle these two variables in a single `upate_config` variable?

modules/compute-mig/variables.tf
```diff
@@ -66,11 +66,29 @@ variable "location" {
...
+  type        = string # REPLACE | RESTART | REFRESH | NONE
+  default     = "REPLACE"
+}
+
 variable "name" {
   description = "Managed group name."
   type        = string
 }
 
+variable "name_instance_config" {
```
ludoo, line 86: I would bundle all four stateful-related variables in a single `stateful_config` one, with four attributes: `name`, `destroy_state`, `group_policy`, `instance_policy`. I'm not entirely clear how do you select the instance to attach an instance policy too, by positional argument or instance name?

ludoo (changes requested):

stenio123:
Thank you for the quick feedback. Looking into it.

Regarding the tests on tests/examples, does it use the *# tftest:modules=2:resources=2* at the bottom of the test? 
No idea where that came from in the earlier examples :-D

stenio123:
@ludoo I really like your idea of bundling the stateful config variables, however I am running into all sorts of issues managing this complex object... you can check the ongoing discussion here https://discuss.hashicorp.com/t/loop-with-conditional-using-object-known-only-after-apply-oh-my/32639

Is there somewhere in the repo where you might have seen a similar situation that I can use as an example? I am concerned about overcomplicating things unnecessarily. Thanks!

ludoo:
Hey, thanks for spending time on this! This type of error is nasty, basically the rule is keys used in for_each need to be static (i.e. not computed at runtime), while values can be dynamic. I can try and have a look at it tomorrow, don't go crazy on this!

stenio123:
Thank you! I think I got it after the latest reply on the Hashicorp forum, at least all tests are passing :-)

Let me stress test it a bit and clean up everything and will update today after lunch. The only thing is I believe I might still have the error on tests/examples/test_plan.py since I dont really understand where it gets the values from. Will do some digging before pushing.

modules/compute-mig/main.tf
```diff
@@ -152,22 +152,24 @@ locals {
 }
 
 resource "google_compute_per_instance_config" "default" {
-  count = var.stateful_disk_instance == null ? 0 : 1
-  zone  = var.location
+  for_each = try(var.stateful_config.per_instance_config, tomap({}))
```
ludoo, line 155: nit: you can skip `tomap` and only `{}`.

modules/compute-mig/main.tf
```diff
@@ -152,22 +152,24 @@ locals {
...
   # terraform error, solved with locals
   #instance_group_manager           = var.regional ? google_compute_region_instance_group_manager.default : google_compute_instance_group_manager.default
   instance_group_manager           = local.instance_group_manager[0].id
-  name                             = var.name_instance_config
+  name                             = each.key
   project                          = var.project_id
-  minimal_action                   = var.minimal_action
-  most_disruptive_allowed_action   = var.most_disruptive_allowed_action
-  remove_instance_state_on_destroy = var.remove_instance_state_on_destroy
+  minimal_action                   = each.value.update_config.minimal_action
```
ludoo, line 163: Would it make sense to guard against the user passing a `null` value for `update_config`? I skimmed the resource docs, and everything seems to be optional except for `name` and `instance_group_manager`. If that's so, we should wrap in `try`.

ludoo (commented):
Thanks *a lot* for this. I dropped a couple comments, then it might be good to go!

stenio123:
Hello I made the requested changes and regular tests passing. Questions -

1- Could you please advise where I can find "check_documentation.py"? I checked the ci folder and have no idea what linting tests are being made, so dont know how to address the error
2- Related to that, not sure what is expected from "tests\examples"

Thanks!

ludoo:
Sorry, I should have explained earlier. :)

The READMEs are checked against the output of `tools/tfdoc.py`, which creates and injects the vars/outputs tables present at the bottom of each README file. If you're ok with creating a Python virtualenv and installe requirements from the `tools` folder, just run `tfdoc` from the root folder and pass the module's directory as its only argument.

As for examples, they are automatically run as tests, the `tftest` comment in them states the number of resources and modules used in plan.

If you prefer me to do the small fixes above, we can merge this into a branch, and I'll then do them myself before branching to master. No big deal if you prefer this way of course!

modules/compute-mig/main.tf
```diff
@@ -152,23 +152,23 @@ locals {
...
   zone = var.location
   # terraform error, solved with locals
   #instance_group_manager           = var.regional ? google_compute_region_instance_group_manager.default : google_compute_instance_group_manager.default
   instance_group_manager           = local.instance_group_manager[0].id
   name                             = each.key
   project                          = var.project_id
-  minimal_action                   = each.value.update_config.minimal_action
-  most_disruptive_allowed_action   = each.value.update_config.most_disruptive_allowed_action
-  remove_instance_state_on_destroy = each.value.update_config.remove_instance_state_on_destroy
+  minimal_action                   = try(each.value.update_config.minimal_action)
```
ludoo, line 163: I'm reasonably sure you need to explicitly specifiy `null`: `try(..., null)`

ludoo (approved):

PR #378, opened 2021-12-06
net-vpc module: added subnet mini-factory

sruffilli:
This PR integrates a "mini-factory" to drive subnet creation from well-formed YaML files.

### DIFF
--- a/modules/net-vpc/main.tf
+++ b/modules/net-vpc/main.tf
@@ -64,10 +64,11 @@ locals {
       : []
     )
   }
-  subnets = {
+  subnets = merge({
     for subnet in var.subnets :
     "${subnet.region}/${subnet.name}" => subnet
-  }
+  }, local.subnet_data)
+
   subnets_l7ilb = {
     for subnet in var.subnets_l7ilb :
     "${subnet.region}/${subnet.name}" => subnet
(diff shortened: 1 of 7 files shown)


sruffilli:
Still missing tests, don't merge!

modules/net-vpc/main.tf
```diff
@@ -77,6 +78,39 @@ locals {
...
+  }
+
+  subnet_descriptions = merge(var.subnet_descriptions, local.subnet_data_descriptions)
+
+  subnet_data_private_access = {
+    for k, v in local._subnet_data : "${v.region}/${k}" => try(v.private_ip_google_access, true)
+  }
+
+  subnet_private_access = merge(var.subnet_private_access, local.subnet_data_private_access)
+
```
sruffilli, line 113: Missing IAM bindings from YaML

sruffilli:
Ready for review @ludoo

modules/net-vpc/main.tf
```diff
@@ -77,6 +67,58 @@ locals {
...
+    trimsuffix(basename(f), ".yaml") => merge(
+      yamldecode(file("${var.data_folder}/${f}")),
+      {
+        project_id = var.project_id
+        network    = local.network.name
+      }
+    )
+  }
+
+  subnet_data = var.data_folder == null ? {} : {
```
ludoo, line 82: you already check for `var.data_folder` in `_subnet_data`

ludoo: Do we really need a second step? The variables are all there in `_subnet_data` already. I would move the `try` in the resource, and do away with most of these locals.

ludoo (approved):

modules/net-vpc/main.tf
sruffilli: Ack - cleaned check in `subnet_data`

I reckon we do need the second step, as yamldecode doesn't allow us to further manipulate datafields. Might be wrong.

sruffilli: Resolved in a0f1593

PR #379, opened 2021-12-06
SA key uploading and credentials json generation with terraform.

averbuks:
Generationg key pair with openssl and uploading public part to an SA to generate an SA json key
- to avoid SA key storing in TF state
- to avoid SA key downloading to a user laptop or moving from hands to hands
- to allow SA keys with expiration time


cloud-operations/onprem-sa-key-management/README.md
```diff
@@ -0,0 +1,54 @@
+# Generationg and uploading public keys for a service accounts
```
lgandras, line 1: s/Generationg/Generating

averbuks: Thx, addressed

cloud-operations/onprem-sa-key-management/main.tf
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+module "project" {
+  source         = "../../modules/project"
+  name           = var.project_id
+  project_create = var.project_create
+}
```
ludoo, line 21: nit: is there any required service that needs to be activated at the project level?

cloud-operations/onprem-sa-key-management/main.tf
```diff
@@ -0,0 +1,47 @@
...
+  project_id = module.project.project_id
+  name       = "onprem-data-uploader"
+  iam_project_roles = {
+    (module.project.project_id) = [
+      "roles/bigquery.dataOwner",
+      "roles/bigquery.jobUser",
+      "roles/storage.objectAdmin"
+    ]
+  }
+  public_keys_directory = "public-keys/data-uploader/"
```
ludoo, line 34: Aleks, do you think this could be generalized? Instead of hardcoding service account names and paths, use a map variable with sa name => key, roles?

ludoo (approved):
This is great. My only nit is I would love this to be generalized instead of using hardcoded names/paths. It can be changed later of course, but I would at least add a line to the README explaining that the examples uses fixed names/paths.

cloud-operations/onprem-sa-key-management/main.tf
averbuks: I went with a list of objects, pls take a look and let me know if you prefer map of objects (I kinda like list of objects more;))

cloud-operations/onprem-sa-key-management/main.tf
averbuks: just verified that no services needed to create SA/SA keys, IAM is enabled by default I think. I added `services` variable though (if user wants to enable some APIs and test functionality of the SAs)

modules/iam-service-account/main.tf
```diff
@@ -65,6 +65,29 @@ locals {
     ? try(google_service_account.service_account.0, null)
     : try(data.google_service_account.service_account.0, null)
   )
+  service_account_credential_templates = {
+    for file, _ in local.public_keys_data : file => jsonencode(
+      {
+        type : "service_account",
+        project_id : var.project_id,
+        private_key_id : split("/", google_service_account_key.upload_key[file].id)[5]
+        private_key : "REPLASE_ME_WITH_PRIVATE_KEY_DATA"
```
juliocc, line 74: s/REPLASE/REPLACE/

cloud-operations/onprem-sa-key-management/README.md
```diff
@@ -0,0 +1,79 @@
+# Managing on-prem service account keys by uploading public keys
+
+When managing GCP Service Accounts with terraform, it's often a question on **how to avoid Service Account Key in the terraform state?**
+
+This example shows how to manage IAM Service Account Keys by generating a key pair and uploading the public part of the key to GCP, it has the following benefits:
```
juliocc, line 5: This example shows how to manage IAM Service Account Keys by manually generating a key pair and uploading the public part of the key to GCP. It has the following benefits:

cloud-operations/onprem-sa-key-management/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+ - no [passing keys between users](https://cloud.google.com/iam/docs/best-practices-for-managing-service-account-keys#pass-between-users) or systems
+ - no SA key stored in the terraform state (only public part of the key in the state)
```
juliocc, line 8: no SA key stored in the terraform state

PR #380, opened 2021-12-06
Add support for partitioned tables on Organization sinks

lcaggio:
Add support for partitioned tables on Organization sinks


ludoo:
I'm not a huge fan of re-implementing the block in the variable. Can we use a simple bool, or a separate sink options variable?

ludoo (approved):
Lore, can we find a shorter name, something like `bq_partitioned_table`?

PR #388, opened 2021-12-15
Linux sysctls configuration and Kubelet config

sergiotejon:


juliocc:
Hola @sergiotejon! 

Thanks for the patch. LGTM but can you fix the linting issues (variables should be sorted alphabetically) and sign the CLA?

juliocc:
Almost there. You have to run the tfdoc script (in the tools folder) in modules/gke-nodepool

sergiotejon:
Gracias Julio :) .

Once you read CONTRIBUTING.md is when you know what you need to do :P.

modules/gke-nodepool/variables.tf
```diff
@@ -40,6 +40,18 @@ variable "initial_node_count" {
   default     = 1
 }
 
+variable "kubelet_config" {
+  description = "Kubelet configuration."
+  type        = map(string)
```
ludoo, line 45: Would it make sense to use an object type, to ensure there are no typos?

```
type = object({
  cpu_cfs_quota = string
  cpu_cfs_quota_period = string
  cpu_manager_policy = string
})
```

ludoo (approved):

modules/gke-nodepool/variables.tf
juliocc: Only those 3 are [supported](https://cloud.google.com/sdk/gcloud/reference/beta/container/node-pools/create#--system-config-from-file), so probably a good idea. And same thing for sysctls.

sergiotejon: It makes all sense, of course. Uploaded fix.

PR #392, opened 2021-12-15
Only run tests when changing terraform code

juliocc:
I think we can run terraform tests only if we're changing at least one .tf file. 

What do you think?


ludoo:
> I think we can run terraform tests only if we're changing at least one .tf file.
> 
> What do you think?

What if we edit READMEs? I guess we can trigger tests manually in those cases, can we?

juliocc:
> > I think we can run terraform tests only if we're changing at least one .tf file.
> > What do you think?
> 
> What if we edit READMEs? I guess we can trigger tests manually in those cases, can we?

I knew I was missing something ;) 

I can add `**/README.md` to the glob too to include READMEs too.

juliocc:
Dropipng this PR since

PR #393, opened 2021-12-15
Configurable schedule and time_zone for CAI Export Cloud Scheduler

marine675:
Configurable `schedule` and `time_zone` for CAI Export Cloud Scheduler


ludoo:
Aaron, thanks for the PR. When we wrote this example, we felt the schedule was sufficiently simple to configure in the Terraform code. If you strongly feel for having it customizable, can you bundle together both variables in a single `schedule_config` one?

PR #395, opened 2021-12-16
net-address: Added support for PSA ranges

sruffilli:


modules/net-address/README.md
```diff
@@ -47,6 +47,23 @@ module "addresses" {
...
 
+### PSA addresses
+
+```hcl
+module "addresses" {
+  source     = "./modules/net-address"
+  project_id = var.project_id
+  psa_addresses = {
+    cloudsql-mysql = {
+      address       = "10.10.10.0"
```
ludoo, line 58: Would it make sense to use `range` so as to avoid the need for an extra attribute?

ludoo (approved):
LGTM, nit on variable type

modules/net-address/README.md
sruffilli: Makes sense - the reasoning behind not doing that was to keep a the interface consistent with psc_addresses and the resource itself. It's a trivial change, your call.

ludoo: I'm all for simplifying the interface (cf the GKE module for extreme examples of the approach). It's a minor thing, so it's also ok to leave it like this.

PR #396, opened 2021-12-17
New tfdoc version

ludoo:
This backport the new tfdoc to Fabric. It also regenerates tfdoc tables in every README file, and refactors the check_documentation tool.


tools/tfdoc.py
```diff
@@ -14,274 +14,350 @@
...
+- that code has been formatted via `terraform fmt`
+
+The tool supports annotations using the `tfdoc:scope:key value` syntax.
+Annotations are rendered in the optional file table by default, and in the
+outputs and variables tables if the `--extra-fields` flag is used. Currently
+supported annotations are:
+
+- `tfdoc:file:description`
+- `tfdoc:output:consumers`
+- `tfdoc:variable:sorce`
```
juliocc, line 34: s/sorce/source/

tools/tfdoc.py
```diff
@@ -14,274 +14,350 @@
...
+    try:
+      with open(name) as file:
+        body = file.read()
+    except (IOError, OSError) as e:
+      raise SystemExit(f'Cannot read file {name}: {e}')
+    shortname = os.path.basename(name)
+    tags = _extract_tags(body)
+    description = tags.get(
+        'file:description', FILE_DESC_DEFAULTS.get(shortname))
+    modules = set([
```
juliocc, line 168: no need for the list comprehension here. Just set(...) should be enough

tools/tfdoc.py
```diff
@@ -14,274 +14,350 @@
...
+  items.sort(key=lambda i: i.name)
+  yield '| name | description | modules | resources |'
+  yield '|---|---|---|---|'
+  for i in items:
+    modules = resources = ''
+    if i.modules:
+      modules = '<code>%s</code>' % '</code> · <code>'.join(
+          sorted(i.modules))
+    if i.resources:
+      resources = '<code>%s</code>' % '</code> · <code>'.join(
```
juliocc, line 248: interesting...

tools/tfdoc.py
```diff
@@ -14,274 +14,350 @@
...
+  if not items:
     return
-  variables.sort(key=lambda v: v.name)
-  variables.sort(key=lambda v: v.required, reverse=True)
-  yield '| name | description | type | required | default |'
-  yield '|---|---|:---: |:---:|:---:|'
-  row = (
-      '| {name} | {description} | <code title="{type_spec}">{type}</code> '
-      '| {required} | {default} |'
+  items.sort(key=lambda i: i.name)
```
juliocc, line 279: probably not a big deal but you're modifying items. Same thing in `format_outputs`

juliocc (approved):
lgtm. Just a few small comments

tools/tfdoc.py
ludoo: good point :)

tools/tfdoc.py
ludoo: good catch, but I'd say we don't much care as everything is self-contained and items are single use

PR #398, opened 2021-12-21
Fix tfdoc replacement for empty source tables

ludoo:


ludoo:
Force merging as this fixes a dumb mistake on my side that potentially breaks tfdoc replacement.

PR #405, opened 2021-12-31
Org/Folder: Allow for policy association when using rule factory

sruffilli:


modules/folder/outputs.tf
```diff
@@ -24,7 +24,7 @@ output "firewall_policies" {
 output "firewall_policy_id" {
   description = "Map of firewall policy ids created in this folder."
   value = {
-    for name, _ in var.firewall_policies :
+    for name, _ in local.firewall_policies :
     name => google_compute_organization_security_policy.policy[name].id
```
ludoo, line 28: Wouldn't it make sense to cycle on the actual resource?

```hcl
{ for k, v in google_compute_organization_security_policy.policy : k => v.id }
```

ludoo (approved):

modules/folder/outputs.tf
sruffilli: Deferred to the upcoming refactor from Julio

PR #410, opened 2022-01-01
Copyright bump

sruffilli:


sruffilli:
Positive side effect - my linter parsed/fixed all files in the repo in the process.
Notably, tab sized was globally updated from 4 spaces to 2 spaces

PR #418, opened 2022-01-10
Update repo structure

juliocc:
- Move all examples to /examples
- Remove old .ci (Cloud Build) directory
- Update test fixture to allow parallel runs
- Update all internal links
- Add markdown link check to linter


ludoo:
We should also update the top-level README, and any cross-link between examples. And then think about what to do for external linkers (e.g. cloud blog, documentation, etc.).

juliocc:
> We should also update the top-level README, and any cross-link between examples. And then think about what to do for external linkers (e.g. cloud blog, documentation, etc.).

Fixed the internal ones and also added a linter step to make sure there are no broken links.

I'm not sure what to do about incoming links. Any ideas?

ludoo:
> I'm not sure what to do about incoming links. Any ideas?

Not really, I don't we have a solutions and will just have to bite the bullet. For docs we can submit a CL/bug, no idea for GCP blog posts. Our medium articles, we update them ourselves.

BTW we need a top-level README for the examples folder. I can give it a shot, unless you already have something in mind.

ludoo (changes requested):
things to sort out before merging

- list external links and have actions/owners to fix them whenever possible
- new README for examples/

PR #429, opened 2022-01-13
Make project creation optional in gcs to bq example

ludoo:


ludoo:
Next steps: break out resources in separate files, collapse resources into a single project, then write proper examples and tests.

juliocc (approved):

PR #430, opened 2022-01-13
Resource Factory update

sruffilli:


ludoo (approved):
Is this a good time to bring in the project factory module?

sruffilli:
> Is this a good time to bring in the project factory module?
I'd postpone to a new PR since README is still missing. Will TAL at it today.

PR #431, opened 2022-01-14
GLB - Module tests and fixes

LucaPrete:


modules/net-xlb/ssl_certificates.tf
```diff
@@ -15,26 +15,38 @@
...
+  # HC is defined, create also a default HC
+  ssl_certificates_config = (
+    try(var.target_proxy_https_config.ssl_certificates, null) == null
+    || length(coalesce(try(var.target_proxy_https_config.ssl_certificates, null), [])) == 0
+    ? merge(
+      { default = var.ssl_certificates_config_defaults },
+      coalesce(var.ssl_certificates_config, {})
+    )
+    : coalesce(var.ssl_certificates_config, {})
+  )
```
juliocc, line 28: This feels unnecessarily complicated. For example `length(coalesce(try(var.target_proxy_https_config.ssl_certificates, null), [])) == length(try(var.target_proxy_https_config.ssl_certificates, []))`.

And I think you can probably simplify it a bit considering both parts of the ternary operator include `coalesce(var.ssl_certificates_config, {})`. Perhaps something like 
```
ssl_certificates_config = merge(coalesce(var.ssl_certificates_config, {}), {
  default = var.ssl_certificates_config_defaults
})
```

modules/net-xlb/target_proxy.tf
```diff
@@ -19,16 +19,21 @@ locals {
...
-      for cert in try(var.target_proxy_https_config.ssl_certificates, ["default"]) :
+    try(var.target_proxy_https_config.ssl_certificates, null) == null
+    || length(coalesce(try(var.target_proxy_https_config.ssl_certificates, null), [])) == 0
+    ? try(
+      [google_compute_managed_ssl_certificate.managed["default"].id],
+      [google_compute_ssl_certificate.unmanaged["default"].id],
+      null
+    )
+    : [
+      for cert in try(var.target_proxy_https_config.ssl_certificates, []) :
```
juliocc, line 30: same as before.

juliocc (commented):
nice tests. left a comment around some locals

LucaPrete:
Hi @juliocc thanks again for the review and the clarifications today.
The two required blocks have been simplified.
Anyway, I decided to keep -instead- the HC part as it was, given with the simplification we were hitting different errors while executing tests for corner cases.

juliocc:
LGTM. Thanks for the changes.

juliocc (approved):

PR #432, opened 2022-01-14
Project Factory

sruffilli:


sruffilli:
Outputs require some love, but I wouldn't really know what could be interesting to return.
Any taker?

ludoo (approved):

ludoo:
> Outputs require some love, but I wouldn't really know what could be interesting to return. Any taker?

I would merge and refine iteratively, WDYT?

sruffilli:
> I would merge and refine iteratively, WDYT?

sg

PR #441, opened 2022-01-19
replaced firewall_policy_attachments with firewall_policy_association…

apichick:
… in example


juliocc:
Thanks @apichick. Test error is unrelated to your change.

PR #442, opened 2022-01-19
Fix failing tests

juliocc:


juliocc:
@ludoo I think this one fixes the concurrency issue. I did a few tests and it seems to work. Can you give it a try locally?

ludoo:
This is nice. I'm now entirely sure the additional complexity is worth a 25% reduction in test runs (5m40 vs 7m40 for modules), but if you think it is, +1 from me.

We will then need to periodically: a) recheck all providers used across the repo and recompile a versions file, b) bump minimum versions in it. Not a big deal of course, so up to you.

ludoo (approved):

PR #443, opened 2022-01-19
Support service dependencies for crypto key bindings in project module

lcaggio:
Some GCP service rely on other GCP services. Those services need Crypt/Decrypt role on the specified key for several robot accounts.

Example: Composer requires Crypt/Decrypt role on the specified key on the following services:
 - artifactregistry
 - container-engine
 - composer
 - compute
 - pubsub
 - storage

Documentation: https://cloud.google.com/composer/docs/cmek#configuring_cmek_support_for_cloud_composer


ludoo:
Lore, can we add a test case to project module to test for this?

ludoo:
@lcaggio introducing dependent services added the possibility of duplicate keys in the key binding resource if the same service was present as both a first-level, and a dependency of another service.

For example, setting `service_encryption_key_ids` to `{compute=["key1"], dataflow=["key1", "key2"]}` would result in the `compute.key1` index being present twice.

I added `distinct()` to wrap the service/key combinations generated in locals, and now it seems to work. There are two new tests for project module that test use cases with and without dependent services.

ludoo (approved):
Approved after changes (see comments in PR).

PR #448, opened 2022-01-21
fix recreation of resources that depend on project_id on pre-existing…

drebes:
… projects when data source refresh is triggered

### DIFF
--- a/modules/project/main.tf
+++ b/modules/project/main.tf
@@ -47,9 +47,17 @@ locals {
   parent_id   = var.parent == null ? null : split("/", var.parent)[1]
   prefix      = var.prefix == null ? "" : "${var.prefix}-"
   project = (
-    var.project_create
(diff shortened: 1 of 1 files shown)


modules/project/main.tf
```diff
@@ -47,9 +47,17 @@ locals {
   parent_id   = var.parent == null ? null : split("/", var.parent)[1]
   prefix      = var.prefix == null ? "" : "${var.prefix}-"
   project = (
-    var.project_create
-    ? try(google_project.project.0, null)
-    : try(data.google_project.project.0, null)
+    var.project_create ?
```
drebes, line 50: Let me know if you prefer the conditional in each field instead of a single one for the whole object.

juliocc: single conditional is easier to read IMO

juliocc (approved):

PR #450, opened 2022-01-22
Refactor project module, support per-file tags in tfdoc

ludoo:
This refactors and cleans up project module internals for better readability, and implements support via an extra tfdoc tag for options overrides on a per-file basis.

It's now possible to activate the tfdoc files table on a specific module, by adding a `<!-- TFDOC OPTS files:1 -->` tag anywhere in the README file.


ludoo:
Adding support for auto turning off a column in the files table, based on contents (e.g. no modules, don't output that column), as suggested by @juliocc

juliocc (approved):
I love it

PR #452, opened 2022-01-24
Bring back to 0 vm-test counter in net landing

LucaPrete:


ludoo:
Luca, what difference does this make?

LucaPrete:
It's just a "cosmetic" fix.
All other resources have their counter at 0. I think we simply forgot this
suffix at 1.

Il giorno lun 24 gen 2022 alle ore 08:36 Ludovico Magnocavallo <
***@***.***> ha scritto:

ludoo (approved):

PR #453, opened 2022-01-24
Refactor onprem VPN vars

ludoo:
This allows changing type of peer gateway in the onprem VPN, without having to edit the code. It also restructures IP ranges so they consume less IP space, so they can easily be used for multiple parallel installations.


fast/stages/02-networking/variables.tf
```diff
@@ -25,17 +25,15 @@ variable "custom_adv" {
   type        = map(string)
   default = {
     cloud_dns             = "35.199.192.0/19"
+    gcp_all               = "10.128.0.0/16"
+    gcp_dev               = "10.128.32.0/19"
+    gcp_landing           = "10.128.0.0/19"
```
sruffilli, line 30: Nit: this implies a contiguous IP space across different regions, which is not a given.
Not gating as it's simple enough to modify for customers.

sruffilli (approved):

PR #466, opened 2022-01-27
Update subnets

sruffilli:


sruffilli:
This aligns the subnets created by the factory with the new IP plan

ludoo (approved):

PR #476, opened 2022-01-30
Updated modules README to include details around module versioning an…

apichick:
…d how to best use the modules


ludoo (approved):
Edited the text a bit ;)

PR #481, opened 2022-02-01
feat: added missing addons gke

caiotavaresdito:
hey guys! Been awhile, noticed it's missing two addons which I've been using and thought I could add it here.
Tests are failing locally but are not related to the GKE module I've changed, any ideas what that could be?


ludoo:
Hey! Thanks for this. Tests seem to be fine here, what kind of error are you seeing when running them local? We moved things around a bit, is it possible you have stale files around?

ludoo (approved):

PR #482, opened 2022-02-01
Experimental module to derive DNS inbound policy addresses

ludoo:
Workaround for [this provider issue](https://github.com/hashicorp/terraform-provider-google/issues/3753).


juliocc (approved):
Maybe the aggregated endpoing can be used to issue a single request https://cloud.google.com/compute/docs/reference/rest/v1/addresses/aggregatedList

In any case LGTM

PR #484, opened 2022-02-01
Fix READMEs.

lcaggio:
Added small changes based on the setup I did on my org.


ludoo (changes requested):
Example makes little sense here, as customers won't have multiple orgs with the same code

lcaggio:
Setting outputs_location to `../../configs/example` on stage 00 was just for consistency with other stages READMEs.

PR #485, opened 2022-02-02
Add ip_range variable to apigee-x-instance module

srs2210:
Recently `ip_range` field was added to [google_apigee_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/apigee_instance) Terraform resource.

This is in regards to the new `ipRange` field being added to [organizations.instance](https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.instances#Instance) API. The `peeringCidrRange` field is being deprecated.

So need to update the **apigee-x-instance** module to be in line with these updates to Apigee X API. Henceforth, `ip_range` variable can be made mandatory for this module and `cidr_range` variable can be removed.

### DIFF
--- a/modules/apigee-x-instance/README.md
+++ b/modules/apigee-x-instance/README.md
@@ -11,7 +11,7 @@ module "apigee-x-instance" {
   source             = "./modules/apigee-x-instance"
   name               = "my-us-instance"
   region             = "us-central1"
-  cidr_mask          = 22
+  ip_range           = "10.0.0.0/22"
 
   apigee_org_id      = "my-project"
   apigee_environments = [
@@ -29,7 +29,7 @@ module "apigee-x-instance" {
   source              = "./modules/apigee-x-instance"
   name                = "my-us-instance"
   region              = "us-central1"
-  cidr_mask           = 16
+  ip_range            = "10.0.0.0/22"
   disk_encryption_key = "my-disk-key"
 
   apigee_org_id       = "my-project"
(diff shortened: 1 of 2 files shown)


modules/apigee-x-instance/variables.tf
```diff
@@ -34,12 +34,12 @@ variable "apigee_org_id" {
...
-variable "cidr_mask" {
-  description = "CIDR mask for the Apigee instance."
-  type        = number
+variable "ip_range" {
+  description = "Customer-provided CIDR block of length 22 for the Apigee instance."
+  type        = string
   validation {
-    condition     = contains([16, 20, 22], var.cidr_mask)
-    error_message = "Invalid CIDR mask; Allowed values for cidr_mask: [16, 20, 22]."
+    condition     = cidrnetmask(var.ip_range) == "255.255.252.0"
```
ludoo, line 41: nit: is it worth checking for a `null` value, or specifying the new `nullable` modifier?

srs2210: @ludoo We don't need to check for `null` values as all the possible scenarios are handled by `cidrnetmask()` function.
If the `ip_range` variable is `null` or does not match the pattern `X.X.X.X/22` the validation will fail.

ludoo: What I meant is guarding the condition so it does throw an error like this

```
> cidrnetmask(null)
╷
│ 
│   on <console-input> line 1:
│   (source code not available)
│ 
│ Invalid value for "prefix" parameter: argument must not be null.
```

ludoo: `condition = var.ip_range != null && cidrnetmask(var.ip_range) == "255.255.252.0"`

srs2210: Whether or not we add `var.ip_range != null` in the `condition` block, the error message remains the same while doing a `terraform plan` which is as below:

**Case# 1**: `var.ip_range != null` check added
```
$ terraform plan
╷
│ 
│   on variables.tf line 41, in variable "ip_range":
│   41:     condition     = var.ip_range != null && cidrnetmask(var.ip_range) == "255.255.252.0"
│     ├────────────────
│     │ var.ip_range is null
│ 
│ Invalid value for "prefix" parameter: argument must not be null.
```

**Case# 2**: `null` check not added
```
$ terraform plan
╷
│ 
│   on variables.tf line 41, in variable "ip_range":
│   41:     condition     = cidrnetmask(var.ip_range) == "255.255.252.0"
│     ├────────────────
│     │ var.ip_range is null
│ 
│ Invalid value for "prefix" parameter: argument must not be null.
```

srs2210: @ludoo I have modified the condition as below:
`try(cidrnetmask(var.ip_range), null) == "255.255.252.0"`

This way the error thrown by `cidrnetmask` function is being handled and we get the custom error message when validation fails.

```
$ terraform plan
╷
│ Error: Invalid value for variable
│ 
│   on variables.tf line 37:
│   37: variable "ip_range" {
│ 
│ Invalid CIDR block provided; Allowed pattern for ip_range: X.X.X.X/22.
│ 
│ This was checked by the validation rule at variables.tf:40,3-13.
```

ludoo: gah, I forgot about Terraform following both branches even if the first part of the condition is false :facepalm: 

Let me reiterate this is just nitpicking, so feel free to merge at any time :)

I would just set `nullable = false` in the variable, bump the minimum Terraform version in `versions.tf` to `>= 1.1.0`, and call it a day. :)

PR #487, opened 2022-02-03
Add delegated grants for stage 3 service accounts

juliocc:
Allow stage3 service accounts to assign roles/compute.networkUser, roles/container.hostServiceAgentUser and roles/vpcaccess.user. Required for PF, GKE and serverless stages in the pipeline

Also sneaking a small fix in the VPC factory


modules/net-vpc/main.tf
```diff
@@ -49,7 +49,7 @@ locals {
       ip_cidr_range      = v.ip_cidr_range
       name               = k
       region             = v.region
-      secondary_ip_range = try(v.secondary_ip_range, [])
+      secondary_ip_range = try(v.secondary_ip_range, {})
```
sruffilli, line 52: Isn't this an array?

sruffilli:
Can you also clean xpnServiceAdmin in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/4e86fbcd04290293b9936aff15274f393b2e74a9/fast/stages/00-bootstrap/organization.tf#L150?

I think we can remove 
```
"compute.subnetworks.getIamPolicy",
"compute.subnetworks.setIamPolicy",
"dns.networks.bindPrivateDNSZone"
```
and either create a new custom role including those perms to be applied at project level, or re-map to a suitable role.

modules/net-vpc/main.tf
juliocc: It's used as a map further down:
```hcl
secondary_ip_range = each.value.secondary_ip_range == null ? [] : [
    for name, range in each.value.secondary_ip_range :
    { range_name = name, ip_cidr_range = range }
  ]```

ludoo (approved):

PR #488, opened 2022-02-03
FAST: GKE multitenant

juliocc:


fast/stages/02-networking/outputs.tf
```diff
@@ -19,11 +19,13 @@ locals {
   tfvars = {
     "03-gke-multitenant-dev" = jsonencode({
       vpc_host_project = module.dev-spoke-project.project_id
+      shared_vpc_self_link = module.dev-spoke-vpc.self_link
```
juliocc, line 22: @danielmarzini do you need the self link in the gke stage?

PR #489, opened 2022-02-04
Fix FAST bootstrap

lcaggio:
Update custom role name


juliocc:
Hey @lcaggio I didn't notice you PR, this should be fixed once we merge #490

ludoo:
:+1: The plan is to not give this role at org level.

PR #491, opened 2022-02-04
revert net-nva specific output files

ludoo:
There's no need to have different output files for the different net stages. This contradicts our statement that we have interfaces between stages, and the utput files are identical and just bad copy/paste.


ludoo:
We decided to rename net stages and use a single set of outputs for net, closing this @sruffilli will send the rename PR.

PR #499, opened 2022-02-04
xpnAdmin roles at folder level

juliocc:
This PR adds xpnAdmin at folder level for stage3:
- split dev and prod networking projects into their 
- add xpnAdmin (to be replaced with a custom role in upcoming PR) role to stage3 SAs in their respective env folders


fast/stages/01-resman/branch-teams.tf
```diff
@@ -94,6 +94,9 @@ module "branch-teams-team-dev-folder" {
     "roles/resourcemanager.projectCreator" = [
       module.branch-teams-dev-projectfactory-sa.iam_email
     ]
+    "roles/compute.xpnAdmin" = [
```
ludoo, line 97: Should we contemplate svpcs in teams folders? Unless I mistook what this does...

ludoo (approved):

fast/stages/01-resman/branch-teams.tf
juliocc: this is just to allow attaching the project to the svpc. We'll replace it with custom role later (when everything works properly)

PR #503, opened 2022-02-05
Add utility to extract IAM bindings from state

ludoo:


tools/state-iam.py
```diff
@@ -76,6 +77,9 @@ def main(state_file, output, prefix=None):
   resources = data.get('resources', [])
   folders = dict(get_folders(resources))
   bindings = get_bindings(resources, prefix=prefix, folders=folders)
+  bindings = sorted(bindings, key=lambda b: (
+      RESOURCE_SORT.get(b.resource_type, 99), b.resource_id,
+      b.authoritative * -1, b.member_type, b.member_id))
```
juliocc, line 82: 🤯

tools/state-iam.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
```
juliocc, line 1: state_iam.py as all the other files in tools/?

juliocc (approved):
I'm not convinced of having python files inside stage directories. Perhaps we can have a fast/tools directory?

juliocc:
Beautiful 👌

PR #506, opened 2022-02-05
clusters improvement #1

danielmarzini:
- decouple initial_node_count from node_count (cannot be set at the same time) 
- using project_id as a parameters for the clusters variable to being able to place clusters into different projects


juliocc:
What's wrong with this pr? I like it, short and simple

danielmarzini:
@juliocc nothing, but there's something else related to the projects that I wanted to have inside a single PR to give the entire context.

PR #509, opened 2022-02-05
clusters improvement #1 and #2

danielmarzini:
- decouple initial_node_count from node_count (cannot be set at the same time)
- using project_id as a parameters for the clusters variable to being able to place clusters into different projects

### DIFF
--- a/fast/stages/03-gke-multitenant/prod/main.tf
+++ b/fast/stages/03-gke-multitenant/prod/main.tf
@@ -69,7 +69,7 @@ module "gke-project-0" {
 module "gke-project-1" {
   source          = "../../../../modules/project"
   billing_account = var.billing_account_id
-  name            = "${var.environment}-gke-clusters-2j"
+  name            = "${var.environment}-gke-clusters-1"
   parent          = var.folder_id
   prefix          = var.prefix
   labels          = local.labels
--- a/fast/stages/03-gke-multitenant/prod/main.tf
+++ b/fast/stages/03-gke-multitenant/prod/main.tf
@@ -25,11 +25,6 @@ locals {
     "serviceAccount:${module.gke-project-0.service_accounts.cloud_services}",
     "serviceAccount:${module.gke-project-1.service_accounts.cloud_services}"
   ]
-  # host_project_bindings = [
-  #   { role = "roles/container.hostServiceAgentUser", member = local._gke_robot_sas },
-  #   { role = "roles/compute.networkUser", member = local._gke_robot_sas },
-  #   { role = "roles/compute.networkUser", member = local._cloud_services_sas }
-  # ]
   host_project_cloud_services_bindings = [ for member in local._cloud_services_sas :
     {role = "roles/compute.networkUser", member = member }
   ]
--- a/fast/stages/03-gke-multitenant/prod/main.tf
+++ b/fast/stages/03-gke-multitenant/prod/main.tf
@@ -16,19 +16,13 @@
 
 locals {
   labels = merge(var.labels, { environment = var.environment })
-  
-  _gke_robot_sas      = [
-    "serviceAccount:${module.gke-project-0.service_accounts.robots.container-engine}"
-  ]
-  _cloud_services_sas = [
-    "serviceAccount:${module.gke-project-0.service_accounts.cloud_services}"
-  ]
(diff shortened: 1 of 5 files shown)


fast/stages/03-gke-multitenant/prod/main.tf
```diff
@@ -16,13 +16,26 @@
...
-  _gke_robot_sa      = "serviceAccount:${module.gke-project-0.service_accounts.robots.container-engine}"
-  _cloud_services_sa = "serviceAccount:${module.gke-project-0.service_accounts.cloud_services}"
-  host_project_bindings = [
-    { role = "roles/container.hostServiceAgentUser", member = local._gke_robot_sa },
-    { role = "roles/compute.networkUser", member = local._gke_robot_sa },
-    { role = "roles/compute.networkUser", member = local._cloud_services_sa }
+  
+  _gke_robot_sas      = [
+    "serviceAccount:${module.gke-project-0.service_accounts.robots.container-engine}",
+    "serviceAccount:${module.gke-project-1.service_accounts.robots.container-engine}"
```
juliocc, line 22: Managing bindings for more than one project unnecessarily complex as this file is supposed to have a single project. The `gke-project-1` was a test I left behind accidentally. Please rebase your branch

juliocc (changes requested):
Hey @danielmarzini. Thanks for the PR, se comments below:

Yu have to rebase your branch. Once you do that you'll notice you don't have to do the thing you're doing with bindings in the host project.

The part regarding the project_id seems a little redundant since we're deploying a single project (at least initially) but we can keep it.

The initial_node_count fix is definitely needed since it's currently broken.

fast/stages/03-gke-multitenant/prod/main.tf
danielmarzini: @juliocc I re-added the project-1, I think having a two projects scenario brings value for a multitenant setup, but, obviously, let's discuss about this.

Not sure about the things that I did with the binding, I did them cause weren't managed somewhere else, where the point where they are already managed?

fast/stages/03-gke-multitenant/prod/main.tf
```diff
@@ -66,15 +74,62 @@ module "gke-project-0" {
   # }
 }
 
+module "gke-project-1" {
```
juliocc, line 77: I prefer having a single project, at least initially. As you can see things get more complicated if we allow more than one.

juliocc (changes requested):

danielmarzini:
@juliocc remove gke-project-1 and its binding, left there the structure to host future projects addition. wdyt?

fast/stages/03-gke-multitenant/prod/main.tf
```diff
@@ -73,8 +79,14 @@ module "gke-dataset-resource-usage" {
   friendly_name = "GKE resource usage."
 }
 
-resource "google_project_iam_member" "host_project_bindings" {
-  for_each = { for i, v in local.host_project_bindings : i => v }
+resource "google_project_iam_member" "host_project_gke_robot_bindings" {
```
juliocc, line 82: why split these in two? It easier to manage the depends_on that you need on the cluster if you have a single resource.

fast/stages/03-gke-multitenant/prod/main.tf
```diff
@@ -16,13 +16,19 @@
...
-    { role = "roles/compute.networkUser", member = local._gke_robot_sa },
-    { role = "roles/compute.networkUser", member = local._cloud_services_sa }
+  
+  _gke_robot_sas      = [
+    "serviceAccount:${module.gke-project-0.service_accounts.robots.container-engine}"
+  ]
+  _cloud_services_sas = [
+    "serviceAccount:${module.gke-project-0.service_accounts.cloud_services}"
+  ]
+  host_project_cloud_services_bindings = [ for member in local._cloud_services_sas :
```
juliocc, line 26: why do you split this in two variables? Can't you just create variable as before?

juliocc (changes requested):
left a few comments on the code. Also, can you remove the intermediary commits (or at least squash when mergin back to fast-dev-gke)?

danielmarzini:
@juliocc now adjusted

juliocc (approved):

PR #510, opened 2022-02-05
Add Data Platform to FAST

lcaggio:
Add Data Platform example to FAST


ludoo:
Lore, gave it a quick glance. There are minor things to check (mainly extra variables added to stage 2s), apart from that it looks good. Will spend some time on it tomorrow.

fast/stages/02-security/variables.tf
```diff
@@ -174,6 +174,26 @@ variable "vpc_sc_perimeter_ingress_policies" {
   default = null
 }
 
+#TODO Ask Team
+variable "vpc_sc_dataplatform_projects" {
```
ludoo, line 178: These should not go here, as explained in the README the perimeters need to be controlled via the existing variables. Using variables for each subset of projects would make this unmanageable.

fast/stages/02-security/main.tf
```diff
@@ -44,4 +44,41 @@ locals {
     "secretmanager.googleapis.com",
     "stackdriver.googleapis.com"
   ]
+  _data_platform_projects = {
```
ludoo, line 47: This is not supposed to be here. Stage 2s don't need to know about any of stage3 resources.

fast/stages/02-networking-vpn/variables.tf
```diff
@@ -50,6 +50,13 @@ variable "data_dir" {
   default     = "data"
 }
 
+variable "data_platform_sa" {
```
ludoo, line 53: This will need to use a generic variable, shared by any service account created in stage 1.

ludoo (changes requested):
Sent a couple initial comments on integrations with the rest of fast that need to be implemented differently.

lcaggio:
Thanks Ludo for the initial comments, I have some TODO I would like to check with you on Monday. For example: I added 2 ways to handle VPC-SC, I'd like to check with you which is the more suitable one (or identify a third one :-) ).

examples/data-solutions/data-platform-foundations/02-load.tf
```diff
@@ -83,13 +83,13 @@ module "lod-prj" {
 }
 
 module "lod-vpc" {
-  count      = var.network_config.network != null ? 0 : 1
+  count      = var.network_config.network_self_link != null ? 0 : 1
   source     = "../../../modules/net-vpc"
   project_id = module.lod-prj.project_id
   name       = "${local.prefix_lod}-vpc"
```
ludoo, line 89: If `project_create` is null or false, project names should hardcoded and use prefix, as all other examples in Fabric do. It should also use the `project_ids` (with trailing `s`) variable as a flag instead of `project_create`, so project creation depends on a single variable, not two interdependent ones.

lcaggio: Fixing this in PR #467 .

PR #511, opened 2022-02-05
Avoid sa overlap when sa_auto_create  is used in multiple clusters

danielmarzini:
gke_nodepools has to option to autocreate the associated service account. In the case where two clusters in the same project with two nodepools with the same name are created, the name of sa of the first cluster is in conflict with the second and prevent the second nodepool to be created.


danielmarzini:
WIP since I didn't generated doc yet

juliocc:
As a general rule we don't add random prefix/suffixes to resources managed by modules. If you the SA created by the module is causing you trouble, you can set `node_service_account_create=false` and manage it yourself.

PR #516, opened 2022-02-06
gke multi-tenant improvements

danielmarzini:
- added network configuration for region europe-west3 in 02-networking-vpn
- introduce service account variable to attach a specific one
- gke multi-tenant / config management layer


danielmarzini:
@juliocc I know, it's long!

fast/stages/02-networking-vpn/data/subnets/prod/prod-gke-nodes-ew3.yaml
```diff
@@ -0,0 +1,8 @@
+# skip boilerplate check
```
juliocc, line 1: why do you need another subnet? We already have a gke-specific one in ew1

fast/stages/03-gke-multitenant/prod/gke-membership.tf
```diff
@@ -0,0 +1,99 @@
...
+      ? module.gke-config-management-platform-repo[0].url
+      : null
+    )
+    : var.config_management.repository_url
+  )
+}
+
+
+#add the cluster to the gke hub
+resource "google_gke_hub_membership" "membership" {
```
juliocc, line 25: everything related to GKE hub is probably better suited in a separate module.

fast/stages/03-gke-multitenant/prod/gke-membership.tf
```diff
@@ -0,0 +1,99 @@
+
```
juliocc, line 1: missing boilerplate

fast/stages/03-gke-multitenant/prod/gke-membership.tf
```diff
@@ -0,0 +1,99 @@
...
+#enable configmanagement feature on the the clusters registered into the hub
+resource "google_gke_hub_feature" "feature" {
+  provider = google-beta
+  project  = module.gke-project-0.project_id
+  name     = "configmanagement"
+  location = "global"
+}
+
+# # sa used by configsync, via WorkloadID to pull code from the repo
+module "gke-wid-platform-sa" {
```
juliocc, line 45: what is this sa for?

fast/stages/03-gke-multitenant/prod/gke-membership.tf
```diff
@@ -0,0 +1,99 @@
...
+  for_each      = local.clusters
+  membership_id = each.key
+  project       = module.gke-project-0.project_id
+  endpoint {
+    gke_cluster {
+      resource_link = "//container.googleapis.com/projects/${module.gke-project-0.project_id}/locations/${each.value.location}/clusters/${each.key}"
+    }
+  }
+}
+#enable configmanagement feature on the the clusters registered into the hub
```
juliocc, line 36: - leave space between modules
- comments are not code but should use consistent style

fast/stages/03-gke-multitenant/prod/variables.tf
```diff
@@ -59,6 +59,18 @@ variable "cluster_defaults" {
   }
 }
 
+variable "config_management" {
+  description = ""
```
juliocc, line 63: provide description

juliocc (changes requested):
there's a lot going on in this PR. I can identify at least 5 things:
- setting up workload identity
- setting up config management
- fixing issues with the initial node count
- adding an optional service account to the module
- adding a new subnet

I suggest you split this PR into individual PRs with meaningful names for each one. In any case I'm leaving a few comments here

ludoo:
Proposal: why don't we go back to the original plan. Which was that the infra part was developed as the other stages, and then we started iterating to make it accept the "appdev" part. With two separate teams managing the two sides, so in the end we hopefully define a "contract" between them.

juliocc:
> Proposal: why don't we go back to the original plan. Which was that the infra part was developed as the other stages, and then we started iterating to make it accept the "appdev" part. With two separate teams managing the two sides, so in the end we hopefully define a "contract" between them.

+1. What would you say should be done

juliocc:
(sorry hit close instead of comment accidentaly)

> > Proposal: why don't we go back to the original plan. Which was that the infra part was developed as the other stages, and then we started iterating to make it accept the "appdev" part. With two separate teams managing the two sides, so in the end we hopefully define a "contract" between them.
> 

I agree with this. I think we're close to a bare-bones gke stage that deploys a project with n clusters and the nodepools. How much more than that should we do here? rbac? or workload identity? MCI? more?

danielmarzini:
@ludoo +1, my fault on considering all the TF part as infra, I'll split them out. @juliocc of course, I'll split PRs for the infra part first, then I'll create a new branch for the "appdev" part.

ludoo:
Ok, let's put pause for a moment to the flurry of PRs, and align directly so that Julio and I can finish the infra part with what's needed. Then we merge that, then we loop you in Daniel so we can integrate all the GKE internal bits. Deal?

danielmarzini:
> 

got it

PR #518, opened 2022-02-07
Ignore initial_node_count setting for Node Pool

danielmarzini:
as per ref https://github.com/hashicorp/terraform-provider-google/issues/6901

### DIFF
--- a/modules/gke-nodepool/main.tf
+++ b/modules/gke-nodepool/main.tf
@@ -85,10 +85,17 @@ resource "google_container_node_pool" "nodepool" {
   name     = var.name
 
   initial_node_count = var.initial_node_count
-  max_pods_per_node  = var.max_pods_per_node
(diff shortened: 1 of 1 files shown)


modules/gke-nodepool/main.tf
```diff
@@ -85,10 +85,17 @@ resource "google_container_node_pool" "nodepool" {
...
 
   initial_node_count = var.initial_node_count
-  max_pods_per_node  = var.max_pods_per_node
-  node_count         = var.autoscaling_config == null ? var.node_count : null
-  node_locations     = var.node_locations
-  version            = var.gke_version
+
+  lifecycle {
+    ignore_changes = [
+      initial_node_count
```
ludoo, line 91: I'm not too keen on this. If people use this attribute, they should expect it to behave as documented in the provider resource. If they need to set lifecycle, they can easily fork the module and apply the change to their version.

PR #521, opened 2022-02-07
Jccb/fast gke use pf

juliocc:
- Remove project_id field from `clusters` variable -- the stage works with 1 project, n clusters
- Move some of the project creation complexity to the PF module


fast/stages/03-gke-multitenant/prod/main.tf
```diff
@@ -17,55 +17,37 @@
...
 }
 
 module "gke-project-0" {
-  source          = "../../../../modules/project"
-  billing_account = var.billing_account_id
-  name            = "${var.environment}-gke-clusters-0"
-  parent          = var.folder_id
-  prefix          = var.prefix
-  labels          = local.labels
+  source             = "../../../../examples/factories/project-factory"
```
ludoo, line 29: it's a bit weird to use an example as a module... if we want to do this, we should have a project-factory module probably

juliocc: Reverted back to project module as discussed

PR #522, opened 2022-02-07
Remove data example

lcaggio:
Remove old gcs2bq example.


ludoo:
This has merge conflicts, and still uses old checks. Closing and replacing with #523

PR #525, opened 2022-02-08
Add support for Shared VPC service IAM to project module

ludoo:


ludoo:
Project factory and FAST changes next.

juliocc (approved):

modules/project/shared-vpc.tf
```diff
@@ -16,6 +16,21 @@
 
 # tfdoc:file:description Shared VPC project-level configuration.
 
+locals {
+  # compute the host project IAM bindings for this project's service identities
+  _svpc_service_iam = flatten([
+    for role, services in var.shared_vpc_service_config.service_identity_iam : [
```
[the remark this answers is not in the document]
ludoo: Good point. I'm also removing the nullables while we're dealing with this.

examples/factories/project-factory/main.tf
```diff
@@ -44,12 +40,22 @@ locals {
...
+  )
+  billing_alert = (
+    var.billing_alert == null
+    ? try(var.defaults.billing_alert, null)
+    : var.billing_alert
+  )
+  essential_contacts = concat(
+    try(var.defaults.essential_contacts, []), var.essential_contacts
+  )
   host_project_bindings = merge(
```
sruffilli, line 59: This should be removed

examples/factories/project-factory/main.tf
```diff
@@ -142,7 +165,7 @@ resource "google_compute_subnetwork_iam_binding" "binding" {
   subnetwork = "projects/${local.vpc_host_project}/regions/${split("/", each.key)[0]}/subnetworks/${split("/", each.key)[1]}"
   region     = split("/", each.key)[0]
   role       = "roles/compute.networkUser"
-  members    = concat(each.value, local.network_user_service_accounts)
+  members    = each.value
 }
 
 resource "google_project_iam_member" "host_project_bindings" {
```
sruffilli, line 171: This resource should now be obsolete too.

ludoo: don't you need to allow users to set network user at host project level from pf?

examples/factories/project-factory/main.tf
```diff
@@ -122,9 +133,21 @@ module "project" {
   policy_list                = try(var.org_policies.policy_list, {})
   service_encryption_key_ids = var.kms_service_agents
   services                   = local.services
-  shared_vpc_service_config = {
-    attach       = local.vpc_setup
-    host_project = local.vpc_host_project
+  shared_vpc_service_config = !var.vpc_setup ? null : {
```
sruffilli, line 136: Var is not defined

sruffilli: Replacing it leads to 
```
var.vpc is a object, known only after apply
```

examples/factories/project-factory/main.tf
sruffilli: Aren't you doing that on `module.project.shared_vpc_service_config`?

sruffilli: This resource applied `local.host_project_bindings` to the host project, but AFAICT that has been moved to the project module.

PR #527, opened 2022-02-09
Add Zonal DNS Only org policy

eeaton:
This is a safe and sane org policy that should be recommended for most customers to prevent them from accidentally configuring internal dns in a way that has reduced availability
https://cloud.google.com/compute/docs/internal-dns#enforce_dns_by_policy


ludoo (approved):
Thank you! You should be able to merge once all checks pass. :)

PR #530, opened 2022-02-10
Update vpc-sc.tf - manage empty perimeters.

sruffilli:


ludoo:
I would prefer to avoid `try` there, as those values then become super fragile in `for_each`. Let me keep this open and replace with a warning in the README.

ludoo:
or better yet, use `coalesce` in locals

ludoo:
PR does not fix the issue anyway ;)

```hcl
│ 
│   on vpc-sc.tf line 49, in locals:
│   48:     landing_to_prod = concat(
│   49:       try(var.vpc_sc_perimeter_projects.landing, []),
│   50:       try(var.vpc_sc_perimeter_projects.prod, [])
│   51:     )
│     ├────────────────
│     │ var.vpc_sc_perimeter_projects.landing is null
│ 
```

fast/stages/02-security/vpc-sc.tf
```diff
@@ -56,7 +57,7 @@ locals {
       access_levels = coalesce(
         try(var.vpc_sc_perimeter_access_levels.dev, null), []
       )
-      resources           = var.vpc_sc_perimeter_projects.dev
+      resources           = try(var.vpc_sc_perimeter_projects.dev, [])
```
sruffilli, line 60: can't we use local.vpc_sc_perimeter_projects.$ENV here and in L78 + L97? @ludoo

ludoo: yep we should, good catch

PR #539, opened 2022-02-12
Fix FAST names

ludoo:
This adds a tool to check names, useful to quickly assess name patterns for some resource types, and eventually in the lint action check to ensure no names go over their limits, including a configurable prefix length.


fast/stages/01-resman/branch-sandbox.tf
```diff
@@ -53,7 +53,7 @@ module "branch-sandbox-gcs" {
 module "branch-sandbox-sa" {
   source      = "../../../modules/iam-service-account"
   project_id  = var.automation_project_id
-  name        = "resman-sandbox-0"
+  name        = "resman-sdbx-0"
```
juliocc, line 56: sdbx sounds related to a database. Maybe sbox?

juliocc (approved):
nice

PR #543, opened 2022-02-14
Reformat python code using yapf

juliocc:
- Reformat code using yapf
- Add linting check
- Rename REQUIREMENTS.txt to requirements.txt

PD: yapf while I convince @ludoo to use Black 😀


ludoo (approved):
+1 once you fix linting error :stuck_out_tongue:

PR #545, opened 2022-02-15
Change psa type in net-vpc, fix psa in fast net stages

ludoo:
This is an incompatible change that changes the type of the `psn_ranges` in the `net-vpc` module.

It also updates the FAST networking stages so they use the net VPC module for PSA, and use the `google_service_networking_connection` resource which was missing before.


juliocc (approved):

modules/net-vpc/main.tf
```diff
@@ -79,10 +79,10 @@ locals {
     : element(reverse(split("/", var.peering_config.peer_vpc_self_link)), 0)
   )
   psn_ranges = {
-    for r in(var.psn_ranges == null ? [] : var.psn_ranges) : r => {
-      address       = split("/", r)[0]
-      name          = replace(split("/", r)[0], ".", "-")
-      prefix_length = split("/", r)[1]
+    for k, v in(var.psn_ranges == null ? {} : var.psn_ranges) : k => {
```
juliocc, line 82: can we change this with `coalesce(var.psn_ranges, {})`?

LucaPrete (approved):

PR #546, opened 2022-02-15
[WIP] Document stage contracts

juliocc:


fast/contracts.tf
```diff
@@ -0,0 +1,54 @@
...
+  type = object({
+    folder_ids = map(string)
+    automation_resources = map(object({
+      gcs = string
+      sa  = string
+    }))
+  })
+}
+
+variable "f_networking" {
```
sruffilli, line 47: How are we grouping stuff logically?
I prefer

```
shared_vpcs = {
  prod = "self_link_a"
  dev  = "self_link_b"
}
```
over 
```
prod = {
  shared_vpc = "self_link_a"
  host_project = [...]
} 
```

...but that means that consumers will cherry pick across multiple collections (nothing wrong with that).

fast/contracts.tf
```diff
@@ -0,0 +1,54 @@
...
+  description = "Data coming from the networking stage"
+  type = object({
+    shared_vpcs       = map(string)
+    vpc_host_projects = map(string)
```
ludoo, line 51: I would shorten whenever possible: `host_projects`

ludoo:
What about renaming to `i_xxxx` for `interface`?

fast/contracts.tf
ludoo: and even `vpcs`

fast/contracts.tf
```diff
@@ -0,0 +1,61 @@
...
+    }))
+  })
+}
+
+variable "f_networking" {
+  description = "Data coming from the networking stage."
+  type = object({
+    dns_zones     = map(string)
+    host_projects = map(string)
+    vpcs          = map(string)
```
ludoo, line 52: shall we also add subnets? it won't hurt too much and might prove useful, WDYT?

sruffilli: Makes sense. PF (and I assume other S3) don't really consume them as vars, but we might want to change that.

juliocc: added

juliocc:
Closing this in favor of #549

PR #547, opened 2022-02-15
FAST: Align billing_account variables to same format

rosmo:
Use the same object format everywhere for consistency.


ludoo (commented):
+1 if you manage to escape the dreaded rebase cycle with all the merges happening lately... 

Not sure if we use `billing_account_id` more than `billing_account`, if that's the case I would have a slight preference for keeping it to minimize disruption. Feel free to ignore this if you strongly feel for `billing_account` of course.

rosmo:
Since the billing account organization ID is required for certain use cases, I suppose I'd keep the object one.

PR #548, opened 2022-02-15
FAST: Align billing_account variables to same format

rosmo:
Align `billing_account` variables across the stages.


sruffilli:
Hey Taneli, we're reshaping the way vars are passed between stages, and we should have armonized billing in the process. 
PTAL at https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/549

juliocc:
@rosmo something similar was merged with #549

PR #550, opened 2022-02-16
Data Platform: Fix Shared VPC members

lcaggio:


ludoo:
Hey Lore, we're normalizing exchanges between stages and are almost done (see #549). I'll get on this right after we merge that PR, and do the changes to align myself if you don't mind.

ludoo:
ah sorry, this is for the example :facepalm:

ludoo (approved):

PR #552, opened 2022-02-19
Add support for resource management tags and tag bindings

### LINKED ISSUE #267: Add support for tags
- google_tags_tag_binding
- google_tags_tag_key
- google_tags_tag_key_iam
- google_tags_tag_value
- google_tags_tag_value_iam

ludoo:
This adds support for tag keys, tag values, IAM bindings and tag bindings to the organization, folder, project, kms and compute vm modules.

Support for additive IAM has not been added, as this would complicate the code quite a bit having to support multiple nested loops and guard against dynamically created resources in keys. It's much simpler to use plain old resources in code for that.

Bindings are only supported for the object being managed by the module, keeping with the spirit of our module design where we deal with one object and its dependencies/features at a time. So it's not possible to assign tags to folders and projects from the organization module for example, but to do that the folder or project module needs to be used for the bindings.

This fixes #267 for most [supported resources](https://cloud.google.com/resource-manager/docs/tags/tags-supported-services#supported_service_resources) for which we have modules. BigQuery is the exception, and it will be easy to add as soon as we have a use case where we need it

### DIFF
--- a/tests/modules/organization/fixture/variables.tf
+++ b/tests/modules/organization/fixture/variables.tf
@@ -15,93 +15,76 @@
  */
 
 variable "custom_roles" {
-  type    = map(list(string))
+  type    = any
   default = {}
 }
 
 variable "group_iam" {
-  type    = map(list(string))
+  type    = any
   default = {}
 }
 
 variable "iam" {
-  type    = map(list(string))
+  type    = any
   default = {}
 }
 
 variable "iam_additive" {
-  type    = map(list(string))
+  type    = any
   default = {}
 }
 
 variable "iam_additive_members" {
-  type    = map(list(string))
+  type    = any
   default = {}
 }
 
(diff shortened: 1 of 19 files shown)


tests/modules/organization/fixture/variables.tf
```diff
@@ -15,93 +15,76 @@
  */
 
 variable "custom_roles" {
-  type    = map(list(string))
+  type    = any
```
juliocc, line 18: is this intentional?

juliocc (approved):

tests/modules/organization/fixture/variables.tf
ludoo: yep, we don't need to specify var types in test fixtures

Angg4r4: Thank you

PR #555, opened 2022-02-24
Fix broken link

jwtracy:
dnssec config link was broken


ludoo (approved):
Well spotted, thanks!

PR #561, opened 2022-02-26
FAST: add 02-networking-peering stage.

sruffilli:


sruffilli:
This stage is very, very similar to 02-networking-vpn - only difference was replacing VPNs with peerings in the hub-to-spoke communication.

Pros: 
* "completes" the set of basic designs
* is much more cheap to run
* another networking stage
Cons:
* unsure whether different enough to grant a new stage
* another stage to maintain

ludoo:
I was going to say "let's use simlinks for all the common parts" but I'm not sure t's a good idea. Probably better than clooning everything and having to maintain it. WDYT?

Maybe (and it's a big maybe) we could have

```
02-net-hub-and-spoke
    ├── variables-peering.tf.src
    ├── variables-vpn.tf.src
    ├── variables.tf -> variables-vpn.tf.src
```

Which is probably cleaner than having two separate stages.

sruffilli:
Not just variables
- vpn-spoke*.tf
- peering-spoke*.tf (which would be newly created)
- vpn-variables.tf (no need for `peering-variables`)

But then we should also branch README. To minimize drifting problems we could have a common README.md, and have one for each of the pluggables, covering hub-to-spoke interconnection.

If it wasn't for drifting risk I'd rather have separate stages tbqh - "pluggables" complicate docs and setup steps.

ludoo:
> Not just variables
> 
> * vpn-spoke*.tf
> * peering-spoke*.tf (which would be newly created)
> * vpn-variables.tf (no need for `peering-variables`)
> 
> But then we should also branch README. To minimize drifting problems we could have a common README.md, and have one for each of the pluggables, covering hub-to-spoke interconnection.
> 
> If it wasn't for drifting risk I'd rather have separate stages tbqh - "pluggables" complicate docs and setup steps.

Well, if it's a single stage it should come with a single README, with a "cross-VPC connectivity" or somesuch section, with one subesection for peering and one for VPN.

Files different between the two:

- outputs (with common outputs in a separate file)
- variables (with common variables in a separate file)
- peering-* and vpn-* files

Having two stages means *a lot* of duplicate code, and having to maintain both. On the other hand, aligning the common parts would be simple if the files were identical. Which would mean splitting out outputs and variables also in case of two stages. This is probably the simplest thing for both users and maintenance, WDYT? (I would need to update tfdoc but it's simple).

ludoo:
To make my last comment clearer: what if we had two stages, but the common parts between the two were in identical files? We'd have

- variables.tf (only common vars) in both
- variables-vpn.tf (only vpn vars) only in VPN
- variables-peering.tf (only peering vars) only in peering

Same for outputs. If you like it, it's a +1 from me.

ludoo:
> I like the end state, I'm unsure how to get there without either doing some magic ("template" stage implemented by vpn and peering stages) or drifting risk, neither of which I'm a fan of. Will let this simmer for a little bit.

Just copy manually when updating. :) If we adopt a scheme as in my last comment, it's super easy to diff between the two, because you only diff the identical files.

sruffilli:
Ofc!

ludoo:
Gave it a shot. tfdoc changes to support multile vars/outputs files still missing.

ludoo (approved):

PR #562, opened 2022-02-27
FAST: use preemptible instances for demo NVA. Comment out test VMs.

sruffilli:


fast/stages/02-networking-nva/nva.tf
```diff
@@ -54,6 +54,13 @@ module "nva-template-ew1" {
     size  = 10
   }
   create_template = true
+  instance_type   = "f1-micro"
+  options = {
+    allow_stopping_for_update = true
+    deletion_protection       = false
+    # Creates preemptible instances, cheaper than regular one. Only suitable for testing.
+    preemptible = true
```
ludoo, line 62: I would not use preemptible, f1-micros are already super cheap and the time spent troubleshooting when they go down will probably cost more (and cause a lot of frustration). :)

sruffilli: You're right - I had this intrusive thought of doing it ESPECIALLY because we're using MIGs, so unless there's a stockout everything should generally work. 
I ofc don't expect this to happen in prod.

We can probably drop the whole PR.

sruffilli: Ah well spotted. ILB has its own HC but per docs our best practice is to also have autohealing on MIGs. Will add that.

sruffilli: Done

fast/stages/02-networking-nva/test-resources.tf
```diff
@@ -18,32 +18,32 @@
...
-    size  = 10
-  }
-  metadata = {
-    startup-script = <<EOF
-      apt update
-      apt install iputils-ping bind9-dnsutils
-    EOF
-  }
-}
+# module "test-vm-landing-untrusted-ew1-0" {
```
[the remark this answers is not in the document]
sruffilli: Put some thought on it, and it's quite messy. 

- project
- region
- zone
- subnetwork
- tags

all differ

Something can be hardcoded (e.g. subnetwork or zone) but gets ugly as soon as default config is changed (e.g. default subnets removed/renamed).
I'd leave everything explicit and commented.

ludoo (approved):

PR #564, opened 2022-02-27
FAST: env DNS zones moved to spoke projects

sruffilli:
NVA (landing) pinging dev:

```
ext_sruffilli_google_com@nva-ew1-0rpz:~$ ping localhost.dev.gcp.example.com
PING localhost.dev.gcp.example.com (127.0.0.1) 56(84) bytes of data.
64 bytes from localhost (127.0.0.1): icmp_seq=1 ttl=64 time=0.028 ms
```


sruffilli:
Ignore it - will merge with the peering PR

PR #569, opened 2022-03-04
Update README for bigquery module

jwtracy:
Some variables that no longer exist. Table format is broken for view/table access.

By the way, after creating a view or table, how can I supply a specific group that can access it in this module? There seems to be mention of an `access_views` variable which sounds like what I want but that does not exist anymore


ludoo:
Thanks for spotting this and you're right, the README file for this module needs some love after the last round of updates. The same variable now also controls access for views, the key part is this in locals:

```hcl
access_view    = { for k, v in var.access : k => v if v.type == "view" }
  identities_view = {
    for k, v in local.access_view : k => try(
      zipmap(
        ["project", "dataset", "table"],
        split("|", var.access_identities[k])
      ),
      { project = null, dataset = null, table = null }
    )
  }
  ```

We should add an example to the README. Quick thing: don't edit the README variables/outputs table by hand or checks will most probably fail. You need to use `tools/tfdoc.py` to generate it. From the root folder:

```bash
./tools/tfdoc.py modules/bigquery-dataset
```

Python requirements are in the `tools` folder too.

lcaggio:
@jwtracy which use case are you trying to achieve? 

Do you want to create an authorized view (supported by `access` and `access_view`) or give access to a single table/view (not supported at the moment by the module, I do not see a pattern to add support of this, customers usually rely on datasets access level, not table/view access level)?

ludoo:
> 

John, we use `tools/tfdoc`from this same repo to format README vars/outputs. If you have PRs that change READMEs and don't want to bother using that, just let us know in the comments, we'll run it for you!

lcaggio:
@ludoo I made a change on `tools/tfdoc` to handle character `|` in the description of variables and I run it for this module.

ludoo:
> @ludoo I made a change on `tools/tfdoc` to handle character `|` in the description of variables and I run it for this module.

Thanks! We should probably also use it for outputs, and use a small characters map to escape all dangerous characters. It's good for now, we'll do that later. :)

PR #570, opened 2022-03-05
Initial sample gitlab-ce implementation

ludoo:
This is a draft PR to discuss how to implement Gitlab CE.

IMHO we should develop a few different pieces

- a simple cloud init only module leveraged by both setups below (in `modules/cloud-config-container`)
- a simple setup that correctly configures one instance without relying on external services
- a more complex one that leverages external services like Cloud SQL and Memorystore an potentially load balancers

Each of those should:

- not manage any resource apart from what is strictly necessary for Gitlab (e.g. no VPC or subnet creation, firewalls, etc.) to make it easier to leverage them in larger Terraform setups
- offer simple customization of basic Gitlab configurations (hostname, etc.) via Terraform variables
- provide a measure of reliability out of the box (i.e. we don't provide a "single instance with owned disk" module)

The current code is mainly used for getting familiar with Gitlab, but it could eventually become the simple setup described above (i.e. with no external services). It currently implements

- a cloud init file that manages a data disk and starts up Gitlab CE via Docker
- a regional PD disk used for data
- an instance template that allows simple configurations of type, tags, etc.
- two stateful autohealed MIGs, one per zone used to manage a single instance each
- commands to instruct MIGs to create or remove one instance with a stateful config that force attaches the data disk

This protects from zonal failures at the cost of a minimal amount of manual operations (create the instance in the surviving zone). Once we add scheduled snapshot and implement the Gitlab configuration variables, it might be ok as an initial "simple" implementation.

@juliocc @sruffilli putting this out here to start brainstorming.

Links we'll need to use for configuration

- [Gitlab Docker images](https://docs.gitlab.com/ee/install/docker.html)
  - [Pre-configure Docker container](https://docs.gitlab.com/ee/install/docker.html#pre-configure-docker-container)
  - [Omnibus Gitlab template](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template)
- [Environment variables](https://docs.gitlab.com/ee/administration/environment_variables.html)
- [Setting custom environment variables](https://docs.gitlab.com/omnibus/settings/environment-variables.html)
- [Configuration options](https://gitlab.com/gitlab-org/omnibus-gitlab/-/blob/master/doc/settings/configuration.md)
- [How to set up GitLab Single Sign-On with Google G Suite](https://medium.com/mop-developers/how-to-set-up-gitlab-single-sign-on-with-google-g-suite-f5e88ae8ba7)
- [NGINX settings](https://gitlab.com/gitlab-org/omnibus-gitlab/-/blob/master/doc/settings/nginx.md)
- [Database settings](https://gitlab.com/gitlab-org/omnibus-gitlab/-/blob/master/doc/settings/database.md)
- [Configuring Redis](https://gitlab.com/gitlab-org/omnibus-gitlab/-/blob/master/doc/settings/redis.md)


rosmo:
Hey @ludoo, I actually have most of the code already for the more complex case, but it uses GKE. Would you be open if I adapted the code for Fabric?

ludoo:
> Hey @ludoo, I actually have most of the code already for the more complex case, but it uses GKE. Would you be open if I adapted the code for Fabric?

Absolutely, I would still want a lighter wight alternative, but having the GKE version makes a lot of sense. Can I also bug you about the GCE version, e.g. on object storage via GCS or anything you think should be configured from the start?

PR #572, opened 2022-03-06
Sruffilli netlogging patch

sruffilli:


ludoo:
Why is your patch mixed up with gitlab ce? can you decouple and base on master? gitlab wil change quite a bit it's not ready to go in now

PR #575, opened 2022-03-07
Add module for retrieving all projects/folders under a specific parent

averbuks:
- module allows retrieving recursively all projects/folders under a specific parent
- allows filtering by label or project status
- use case: maintaining list of VPCSC projects dynamically


ludoo:
Hey Aleks, nice module. :) Can you exclude the doc examples in the README using the skip annotation?

ludoo (approved):

averbuks:
> Aleks, nice module. :) Can you exclude the doc examples in the README using the skip annotation?

I'm on it, fixing some other tests for interconnect attachment, seems provider now have one more parameter.

Issue #576, opened 2022-03-08
Add peering routes config for PSA in net-vpc module


LucaPrete:
The module currently doesn't allow to optionally import/export custom routes.
This can be easily achieved with the google_compute_network_peering_routes_config resource.

PR following soon!

ludoo:
I'm pretty sure the module already does this.

ludoo:
Ah, PSA I thought peering. Good point then.

danielmarzini:
Hey Luke, after the merge I got this errors in the 02-networking-vpn stage:

```Error: Error waiting to create GlobalAddress: Error waiting for Creating GlobalAddress: Invalid IP CIDR range: 10.128.63.0/24 conflicts with IP range 10.128.63.0/24 that was allocated by resource projects/dcx0-dev-net-spoke-0/global/addresses/dev-spoke-0-psa-dev-1.
│ 
│ 
│   with module.dev-spoke-vpc.google_compute_global_address.psa_ranges["cloudsql-sqlserver"],
│   on ../../../modules/net-vpc/main.tf line 331, in resource "google_compute_global_address" "psa_ranges":
│  331: resource "google_compute_global_address" "psa_ranges" {
│ 
╵
╷
│ Error: Error waiting to create GlobalAddress: Error waiting for Creating GlobalAddress: Invalid IP CIDR range: 10.128.95.0/24 conflicts with IP range 10.128.95.0/24 that was allocated by resource projects/dcx0-prod-net-spoke-0/global/addresses/prod-spoke-0-psa-prod-1.
│ 
│ 
│   with module.prod-spoke-vpc.google_compute_global_address.psa_ranges["cloudsql-sqlserver"],
│   on ../../../modules/net-vpc/main.tf line 331, in resource "google_compute_global_address" "psa_ranges":
│  331: resource "google_compute_global_address" "psa_ranges" {```

LucaPrete:
Ciao Dani!

I think this simply has to do with how the provider works.
I guess the apply with the new module will try to create a new range/psc,
without deleting the previous one(s), thus going in conflict.

I’d suggest to comment your range definitions (I’d assume in
terraform.tfvars), apply (thus deleting the old ranges). Then, uncomment
and apply again.

Please, let me know if you run into more issues.

Hope it helps

Il giorno ven 11 mar 2022 alle 23:13 Daniel Marzini <
***@***.***> ha scritto:

PR #577, opened 2022-03-08
[#576] net-vpc - Add the ability to export/import custom routes throu…

LucaPrete:
…gh peering with PSA VPC

### DIFF
--- a/modules/net-vpc/README.md
+++ b/modules/net-vpc/README.md
@@ -138,7 +138,14 @@ module "vpc" {
       secondary_ip_range = null
     }
   ]
-  psa_ranges = {range-a = "10.10.0.0/16"}
+  psa_config = {
+    my_service = {
+      ranges = [
+        "10.0.1.0/24"
+      ],
+      routes = null
+    }
+  }
 }
 # tftest modules=1 resources=4
 ```
@@ -158,9 +165,17 @@ module "vpc" {
       secondary_ip_range = null
     }
   ]
-  psa_ranges               = {range-a = "10.10.0.0/16"}
-  psa_export_custom_routes = true
-  psa_import_custom_routes = true
+  psa_config = {
+    my_service = {
+      ranges = [
+        "10.0.1.0/24"
+      ],
+      routes = {
+        export=true,
+        import=true
+      }
+    }
+  }
 }
 # tftest modules=1 resources=5
 ```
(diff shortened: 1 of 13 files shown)


modules/net-vpc/README.md
```diff
@@ -143,6 +143,28 @@ module "vpc" {
...
+  name       = "my-network"
+  subnets = [
+    {
+      ip_cidr_range      = "10.0.0.0/24"
+      name               = "production"
+      region             = "europe-west1"
+      secondary_ip_range = null
+    }
+  ]
+  psa_ranges               = {range-a = "10.10.0.0/16"}
```
ludoo, line 161: Luca, we need to use a single variable for everything PSA, and allow individual control.

Something like

```hcl
psa_ranges = {
  foo = { range = "10.10.0.0/24", routes = null}
  bar = { range = "10.10.0.0/24", routes = {export = true, import = true}}
}
```

ludoo (changes requested):

modules/net-vpc/README.md
ludoo: Please also use ranges that match with regular use cases in the examples (/24 or /28), which I did not in my comment above. :)

LucaPrete: @ludoo thanks.
Actually, it's even worst... :)
One PSA range can have multiple IP ranges. Also, I would try to make the route import/export params more flat.

So, in my mind the structure would be:

```
psa_config = {
  foo = { ranges = ["10.10.0.0/24"]}
  bar = { ranges = ["192.168.1.0/24", "192.168.2.0/24"], export=true, import=true}
}
```

If we agree on the variables I can proceed with the implementation.

ludoo: Well yes and no, keys are arbitrary (we need them to have static values for iteration) so we might as well keep one range per key. Unless n ranges map to one peering, then I agree with you as at that point the ke represents the peering.

LucaPrete: Also in my case the answer is yes and no :)
It depends. Peerings are created per service per region.
So potentially yes, each key may represent a different peering, and yes: each Private Service Connection can use multiple ranges.

ludoo: Hah :)

I would stick with one range per key so as to be less verbose. Unless you strongly feel for multiple ranges per key support.

LucaPrete: ok, cool! That will make my life also easier when modifying the module. Thanks!

LucaPrete: [for the records] - talking to Ludo, we've decided to allow multiple CIDR per PSA connection.

Final structure will be:

```
psa_config = {
  foo = { ranges = ["10.10.0.0/24"]}
  bar = { ranges = ["192.168.1.0/24", "192.168.2.0/24"], routes = {export=true, import=true}}
}
```

LucaPrete:
This is now actually failing for checksum error downloading a terraform provider. Needs further checks. Then tests will (possibly) pass.

sruffilli:
Re-ran checks - all green.

ludoo (approved):

PR #585, opened 2022-03-17
Fix missing service networking identity in project, PSA

ludoo:


tests/modules/net_vpc/fixture/variables.tf
```diff
@@ -63,24 +59,12 @@ variable "peering_config" {
 
 variable "psa_config" {
```
LucaPrete, line 60: Should we replicate exactly the definition we put in the module?

LucaPrete (approved):

PR #591, opened 2022-03-24
Network dashboard update

aurelienlegrand:
Now using Cloud Asset Inventory to count existing resources, instead of using GCP metrics (that are only populated once a day). This allows us to have up to date data. Optimized code and reduced execution time by 50%.


examples/cloud-operations/network-dashboard/cloud-function/main.py
```diff
@@ -59,39 +63,192 @@ def main(event, context):
...
   get_pgg_data(
       metrics_dict["metrics_per_peering_group"]
-      ["subnet_ranges_per_peering_group"], SUBNET_RANGES_USAGE_METRIC,
+      ["subnet_ranges_per_peering_group"], subnet_range_dict,
       SUBNET_RANGES_LIMIT_METRIC, LIMIT_SUBNETS)
 
   return 'Function executed successfully'
 
 
+def get_l4_forwarding_rules_dict():
```
ludoo, line 101: Once this PR is in, we might want to refactor the CAI `get` methods so they all use a shared `cloud_asset_search` function as there's quite a bit of duplicate code between them.

ludoo (approved):

examples/cloud-operations/network-dashboard/cloud-function/main.py
```diff
@@ -59,39 +63,192 @@ def main(event, context):
...
+        None
+      Returns:
+        forwarding_rules_dict (dictionary of string: int): Keys are the network links and values are the number of Forwarding Rules per network.
+  '''
+  client = asset_v1.AssetServiceClient()
+
+  read_mask = field_mask_pb2.FieldMask()
+  read_mask.FromJsonString('name,versionedResources')
+
+  forwarding_rules_dict = {}
```
juliocc, line 115: `defaultdict(int)` seems like a better option here

examples/cloud-operations/network-dashboard/cloud-function/main.py
aurelienlegrand: Good point, I think this can be addressed after the PR is merged.

ludoo (approved):

PR #596, opened 2022-03-27
Use uniform access on Cloud Functions buckets.

wiktorn:
There is no way to provide whether to use uniform access level or not other than provide own bucket.

Using uniform access in anyway recommended by GCP approach. If someone has use case, where ACL's are required, then they can provide they own bucket.

This makes life easier for cases, where uniform access is mandated by organization policy.

Apart from that, I've fixed few things around versioning:
* missing activation of versioning
* adding to the lifecycle rule that it affects only non-current versions of objects (otherwise, function will lose it's code after specified time)


ludoo (approved):
Wiktor, thanks for this. If documentation tests fail, you need to use `tools/tfdoc.py` to format the vars/outputs table.

ludoo:
Feel free to merge whenever you like!

wiktorn:
> Feel free to merge whenever you like!

Don't have access (yet 😉 )

PR #598, opened 2022-03-28
[#597] Fix link_checker

### LINKED ISSUE #597: Fix tools/link_checker.py
The tool only inspects links in paragraphs and first level elements of the MD docs.

This left out (among others) important objects as
* lists
* nested lists
* nested paragraphs

LucaPrete:
Closes #597


tools/check_links.py
```diff
@@ -29,35 +29,62 @@
...
-        if not isinstance(subel, marko.inline.Link):
-          continue
-        link_valid = None
-        url = urllib.parse.urlparse(subel.dest)
-        if url.scheme:
-          link_valid = True
-        else:
-          link_valid = (readme_path.parent / url.path).exists()
-        links.append(LINK(subel.dest, link_valid))
+    els = marko.parser.Parser().parse(readme_path.read_text()).children
```
ludoo, line 79: Luca, iterating instead of consuming the whole list is usually cheaper. Can you change `check_elements` so it checks a single element, and keep the iteration in `.children` here?

LucaPrete: @ludoo this should have been solved with the latest patch from @juliocc

juliocc (approved):

tools/check_links.py
```diff
@@ -66,12 +79,14 @@ def check_docs(dir_name):
 
 @click.command()
 @click.argument('dirs', type=str, nargs=-1)
-def main(dirs):
+@click.option('-e', '--external', is_flag=True, default=False,
+              help='Whether to test external links.')
+def main(dirs, external):
```
ludoo, line 84: nit: can you also provide a default in the function arg?

ludoo (approved):

tools/check_links.py
```diff
@@ -21,47 +21,72 @@
...
+  # If the link is public, say the link is anyway valid
+  # if --external is not set; check the link otherwise
+  print(link.dest)
+  if url.scheme:
+    link_valid = True
+    if external:
+      user_agent = {'User-Agent': 'Chrome/51.0.2704.64'}
+      try:
+        status_code = requests.get(link.dest, headers=user_agent).status_code
+        if status_code in BAD_STATUS_CODES:
```
juliocc, line 49: 1. Why change the user agent?
2. Use the [response's ok attribute](https://docs.python-requests.org/en/master/api/#requests.Response.ok)
```
response = requests.get(dest)
link_valid = response.ok
```

tools/check_links.py
```diff
@@ -21,47 +21,72 @@
...
+          link_valid = False
+      except requests.exceptions.ConnectionError:
```
juliocc, line 51: I think you want [RequestException](https://2.python-requests.org/en/latest/_modules/requests/exceptions/#RequestException)

juliocc (changes requested):

tools/check_links.py
LucaPrete: Done

tools/check_links.py
LucaPrete: Done

tools/check_links.py
LucaPrete: Done

tools/check_links.py
```diff
@@ -21,47 +21,69 @@
 
 import collections
 import pathlib
+import requests
 import urllib.parse
 
 import click
 import marko
 
+BAD_STATUS_CODES = [401, 401, 403, 404, 500]
```
juliocc, line 30: delete

juliocc (changes requested):

tools/check_links.py
```diff
@@ -31,37 +32,57 @@
...
-def check_docs(dir_name):
-  'Traverse dir_name and check links in Markdown files.'
+def check_link(link, readme_path, external):
+  'Checks if a link element has a valid destination.'
+  link_valid = None
+  url = urllib.parse.urlparse(link.dest)
+  # If the link is public, say the link is anyway valid
+  # if --external is not set; check the link otherwise
+  if url.scheme:
+    link_valid = True
```
ludoo, line 42: nit I would return early for legibility
```python
if not external:
  return True
```

similar for the try/except, so that you don't need the `link_valid` variable

tools/check_links.py
```diff
@@ -31,37 +32,57 @@
...
+def check_docs(dir_name, external=False):
+  'Traverses dir_name and checks for all Markdown files.'
   dir_path = BASEDIR / dir_name
+  parser = marko.parser.Parser()
   for readme_path in sorted(dir_path.glob('**/*.md')):
     if '.terraform' in str(readme_path) or '.pytest' in str(readme_path):
       continue
+
+    root = parser.parse(readme_path.read_text())
+    elements = collections.deque([root])
```
ludoo, line 64: Let me restate: iterating beats creating lists or list-like. Can you please iterate over children instead of assembling a list?

ludoo (changes requested):

PR #604, opened 2022-03-31
Add billing_type for Apigee Organization Module

danistrebel:
Adding the `billing_type` variable from the [Apigee Organization](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/apigee_organization) resource to allow users to specify the billing explicitly.

### DIFF
--- a/modules/apigee-organization/variables.tf
+++ b/modules/apigee-organization/variables.tf
@@ -72,4 +72,8 @@ variable "runtime_type" {
   }
 }
 
-
+variable "billing_type" {
+  description = "Billing type of the Apigee organization."
+  type        = string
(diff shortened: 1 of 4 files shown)


modules/apigee-organization/variables.tf
```diff
@@ -72,4 +72,8 @@ variable "runtime_type" {
   }
 }
 
-
+variable "billing_type" {
+  description = "Billing type of the Apigee organization."
+  type        = string
+  default     = null
+}
```
ludoo, line 79: nit would it make sense to add validation on the possible values?

ludoo (approved):

PR #613, opened 2022-04-07
Allow multiple CIDR blocks in the ip_range for Apigee Instance

danistrebel:
The parameters has changed to allow both a /22 for the runtime and a /28 for the master authorized network of an Apigee instance.

### DIFF
--- a/modules/apigee-x-instance/variables.tf
+++ b/modules/apigee-x-instance/variables.tf
@@ -35,12 +35,9 @@ variable "apigee_org_id" {
 }
 
 variable "ip_range" {
-  description = "Customer-provided CIDR block of length 22 for the Apigee instance."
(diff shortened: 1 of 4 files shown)


modules/apigee-x-instance/variables.tf
```diff
@@ -35,12 +35,9 @@ variable "apigee_org_id" {
 }
 
 variable "ip_range" {
-  description = "Customer-provided CIDR block of length 22 for the Apigee instance."
+  description = "Customer-provided CIDR blocks of length 22 and 28 for the Apigee instance (e.g. `10.0.0.0/22,10.1.0.0/28`)."
```
juliocc, line 38: Wouldn't `list(string)` be better or Is this the expected format by the resource?

danistrebel: indeed but the upstream resource expects a string so I thought it's more consistent to pass the variable with the same name as is

juliocc (approved):

PR #622, opened 2022-04-12
 Add metrics for L7 Forwarding Rules and Catch errors for permission denied

gleichda:
This PR contains following improvements
- Add metrics for L7 Forwarding Rules
- Catch errors for permission denied (Most probably due to VPC Peering in a different org or Google Managed like PSA)
- Allow deployment in an existing GCP Project
- Remove some unused TF variables


examples/cloud-operations/network-dashboard/variables.tf
```diff
@@ -22,8 +22,14 @@ variable "billing_account" {
...
 
+variable "monitoring_project_id" {
+  description = "Monitoring project where the dashboard will be created and the solution deployed; a project will be created if set to empty string"
+  default     = ""
+}
+
 variable "prefix" {
-  description = "Customer name to use as prefix for resources' naming"
+  description = "Customer name to use as prefix for monitoring project"
+  default     = ""
```
aurelienlegrand, line 32: I am in favor of having no default here so that each user sets their own prefix to get unique resources. What do you think?

aurelienlegrand (changes requested):
Thanks for your help David! 
The PR looks good, I added a minor comment.

Could you please run yapf --style="{based_on_style: google, indent_width: 2, SPLIT_BEFORE_NAMED_ASSIGNS: false}" to ensure proper linting of the python code?

I will then approve the PR.

examples/cloud-operations/network-dashboard/variables.tf
gleichda: Agree but in that case even when you don't use the prefix because you don't create a new project you still have to fill the prefix. That's why I set a default of empty

aurelienlegrand: Got it.

aurelienlegrand (approved):

aurelienlegrand:
Thanks for making the changes David!
Approving this PR now.

aurelienlegrand:
Merging David's PR into master, thanks for your contribution!

PR #624, opened 2022-04-12
Move FAST tests to main tests workflow

juliocc:


ludoo:
We also need to remove the deleted action from the branch protection rules. :)

ludoo (approved):

PR #625, opened 2022-04-12
Remove broken link and ignore globals in fast stages

juliocc:


juliocc:
Force-merging to unblock @nason-schrodinger

nason-schrodinger:
Works for me, thanks again!

PR #627, opened 2022-04-13
Improve project module README

### LINKED ISSUE #471: project module `iam` variable with `roles/editor` overrides essential `<project-number>@cloudservices.gserviceaccount.com` permission
project module `iam` variable with `roles/editor` overrides essential `<project-number>@cloudservices.gserviceaccount.com` permission

If roles/editor is specified for a project module stanza, the above cloudservices service account loses its permissions. If these permissions aren't present, GCP APIs behavior is very error prone.

When an authorative binding is used, there should at least be an option to automatically append this service account as a project Editor.

ludoo:
This adds some descriptive text for the project module and its examples, and adds a few extra examples that illustrate specific uses. It also fixes #471 .


juliocc:
Nice. Fixed a few typos here and there.

juliocc (approved):

PR #632, opened 2022-04-15
Add subscription filter support to PubSub module

flaprimo:
Enable filters for pubsub subscriptions:
```hcl
module "pubsub-0" {
  source     = "../pubsub"
  project_id = var.project_id
  name       = var.topic_name
  subscriptions = {
    subscription_name = {
      options = {
        filter                     = "attributes.attribute_name = \"attribute_value\""
        ack_deadline_seconds       = null
        message_retention_duration = null
        retain_acked_messages      = true
        expiration_policy_ttl      = null
      }
    }
```


ludoo (approved):
Grazie Flavio! If you don't see the merge button drop a comment and we'll merge it for you.

PR #635, opened 2022-04-20
Fix gke-cluster permadiff when using dns_config

juliocc:


modules/gke-cluster/main.tf
```diff
@@ -278,12 +278,11 @@ resource "google_container_cluster" "cluster" {
...
 
   dynamic "dns_config" {
-    for_each = var.dns_config != null ? [var.dns_config] : []
-    iterator = config
+    for_each = var.dns_config != null ? [""] : []
     content {
-      cluster_dns        = config.value.cluster_dns
-      cluster_dns_scope  = config.value.cluster_dns_scope
-      cluster_dns_domain = config.value.cluster_dns_domain
+      cluster_dns        = var.dns_config.cluster_dns
```
ludoo, line 283: should this use `coalesce`?

ludoo (approved):

PR #636, opened 2022-04-21
Add option for 'GLOBAL_MANAGED_PROXY' proxy_only subnets

LucaPrete:
Allow arbitrary purpose for proxy only subnets.


modules/net-vpc/subnets.tf
```diff
@@ -134,7 +134,7 @@ resource "google_compute_subnetwork" "proxy_only" {
   region        = each.value.region
   name          = each.value.name
   ip_cidr_range = each.value.ip_cidr_range
-  purpose       = "REGIONAL_MANAGED_PROXY"
```
LucaPrete, line 137: @averbuks @ludoo @juliocc as an alternative, we can also define two different blocks:
1) proxy_only_regional
2) proxy_only_global

...So we remain consistent with the rest of the code.

wdyt?

juliocc: I like a single block with the user specifying the purpose.

juliocc (approved):

modules/net-vpc/subnets.tf
averbuks: +1, I would even stick to just `subnets` var in the future if that comes to the mainstream https://www.terraform.io/language/functions/defaults (still experimental)

averbuks (approved):

ludoo:
Is this still being worked on, or shall we close it?

PR #637, opened 2022-04-21
Data Platform: Update Naming Convention

lcaggio:
Update project naming convention to align the solution with https://cloud.google.com/architecture/confidential-data-warehouse-blueprint


tools/requirements.txt
```diff
@@ -1,4 +1,5 @@
 click
+deepdiff
```
juliocc, line 2: I think you're not really using this.

juliocc (approved):

tools/requirements.txt
lcaggio: I do not use it, but it was needed to run pytest in the root folder

PR #642, opened 2022-04-29
Added 'gke_backup_agent_config' option to addons in gke-cluster

next-davidduquehernandez:
Hi!

This is my first PR :).

I've added option 'gke_backup_agent_config' to gke-cluster. This addon option is still in beta but I've tested it in some clusters and works well.

Thanks for all ;).


ludoo:
Thanks for the PR!

PR #643, opened 2022-04-30
Change hub and spoke VPN example to use VPN HA

### LINKED ISSUE #619: Update the networking example hub-and-spoke-vpn
Since Classic VPN has been partially deprecated on 31st of March, the example in the title needs to be updated with the HA VPN instead of the Classic one.

ludoo:
This changes the hub and spoke VPN example to use VPN HA, and also better aligns it with the "FAST way" of doing things.

Fixes #619


examples/networking/hub-and-spoke-vpn/README.md
```diff
@@ -1,8 +1,15 @@
 # Hub and Spoke via VPN
 
-This example creates a simple **Hub and Spoke VPN** setup, where the VPC network connects satellite locations (spokes) through a single intermediary location (hub) via [IPsec VPN](https://cloud.google.com/vpn/docs/concepts/overview), optionally providing full-mesh networking via [custom route advertisements](https://cloud.google.com/router/docs/how-to/advertising-overview).
+This example creates a simple **Hub and Spoke VPN** setup, where the VPC network connects satellite locations (spokes) through a single intermediary location (hub) via [IPsec HA VPN](https://cloud.google.com/network-connectivity/docs/vpn/concepts/overview#ha-vpn).
 
-The example has been purposefully kept simple to show how to use and wire the VPC and VPN modules together, and so that it can be used as a basis for more complex scenarios. This is the high level diagram:
+A few additional features are also shown:
+
+- [custom BGP advertisements](https://cloud.google.com/router/docs/how-to/advertising-overview) to implement optional full mesh
```
sruffilli, line 7: Full mesh has a different meaning (to me?).
I would rephrase to "to implement full line of sight between spokes"

examples/networking/hub-and-spoke-vpn/README.md
```diff
@@ -1,8 +1,15 @@
...
+- [VPC Global Routing](https://cloud.google.com/network-connectivity/docs/router/how-to/configuring-routing-mode) to use a regional set of VPN gateways from subnets in different regions (used here for illustrative purposes of course)
```
sruffilli, line 8: `regional set of VPN gateways from subnets in different regions (used here for illustrative purposes of course)` ... as next hops? Also, VPN gateways don't actually consume subnets. I'd probably remove this, especially b/c that's the default VPC behavior.

sruffilli: ...and I just got what you meant. subnets in different regions consume vpn gws. Took me a while to decode, maybe rephrase? Or maybe I need another coffee.

examples/networking/hub-and-spoke-vpn/README.md
```diff
@@ -12,44 +19,87 @@ This sample creates several distinct groups of resources:
...
+ping test-r2.dev.example.com
 ```
 
-The example does not account for HA, but the VPN gateways can be easily upgraded to use HA VPN via the [net-vpn-ha module](../../../modules/net-vpn-ha).
-
-If a single router and VPN gateway are used in the hub to manage all tunnels, particular care must be taken in announcing ranges from hub to spokes, as Cloud Router does not explicitly support transitivity and overlapping routes received from both sides create unintended side effects. The simple workaround is to announce a single aggregated route from hub to spokes so that it does not overlap with any of the ranges advertised by each spoke to the hub.
+<!-- TFDOC OPTS files:1 -->
 <!-- BEGIN TFDOC -->
 
+## Files
```
sruffilli, line 72: <3

sruffilli (approved):
LGTM, a few very minor comments

examples/networking/hub-and-spoke-vpn/README.md
ludoo: yep, will rephrase :)

PR #646, opened 2022-05-04
Adds Spot VM support to compute-vm

sruffilli:


modules/compute-vm/README.md
```diff
@@ -31,6 +31,34 @@ module "simple-vm-example" {
...
+```hcl
+module "spot-vm-example" {
+  source     = "./modules/compute-vm"
+  project_id = var.project_id
+  zone     = "europe-west1-b"
+  name       = "test"
+  options = {
+    allow_stopping_for_update = true
+    deletion_protection       = false
+    preemptible               = true
```
ludoo, line 47: do we need both `preemptible` and `spot`? or should this be either/or?

ludoo: or even better, replacing both with a single option that takes `SPOT` and `PREEMPTIBLE` as args.

PR #668, opened 2022-06-07
FAST cleanup instructions

ajlopezn:
cleanup instructions file


fast/CLEANUP.md
```diff
@@ -0,0 +1,108 @@
+# FAST deployment clean up
+In case you require destroying FAST deployment in your organization, follow these steps.
```
juliocc, line 2: s/FAST/a FAST/

fast/CLEANUP.md
```diff
@@ -0,0 +1,108 @@
...
+terraform destroy
+```
+
+## Stage 2 (Networking)
+```bash
+cd $FAST_PWD/02-networking-XXX/
+terraform destroy
+```
+
+There's a minor glitch that can surface running terraform destroy, where the service project attachments to the Shared VPC will not get destroyed even with the relevant API call succeeding. We are investigating the issue, in the meantime just manually remove the attachment in the Cloud console or via the gcloud beta compute shared-vpc associated-projects remove command when terraform destroy fails, and then relaunch the command.
```
juliocc, line 40: Quote the gcloud command with `backticks`

sruffilli:
> 1. I think the file should be placed in fast/stages. wdyt @ludoo @sruffilli?
+1

ajlopezn:
> 

done

juliocc (approved):

PR #675, opened 2022-06-10
Fix FAST audit logs when using pubsub as destination

juliocc:
- Fix FAST audit logging (stage0) when pubsub is used as destination
- Split each organization-level logging sink of type `logging` to a separate logging bucket.
- Add convenience output `id` to `gcs` module

### DIFF
--- a/modules/gcs/outputs.tf
+++ b/modules/gcs/outputs.tf
@@ -19,6 +19,21 @@ output "bucket" {
   value       = google_storage_bucket.bucket
 }
 
+# We add `id` as an alias to `name` to simplify log sink handling.
+# Since all other log destinations (pubsub, logging-bucket, bigquery)
+# have an id output, it is convenient to have in this module too to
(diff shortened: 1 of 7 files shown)


modules/gcs/outputs.tf
```diff
@@ -19,6 +19,21 @@ output "bucket" {
...
 
+# We add `id` as an alias to `name` to simplify log sink handling.
+# Since all other log destinations (pubsub, logging-bucket, bigquery)
+# have an id output, it is convenient to have in this module too to
+# handle all log destination as homogeneous objects (i.e. you can
+# assume any valid log destination has an `id` output).
+
+output "id" {
+  description = "Bucket ID (same as name)."
+  value       = "${local.prefix}${lower(var.name)}"
```
ludoo, line 30: I like this!

ludoo (approved):
nice!!!

PR #678, opened 2022-06-15
Fixed readability of psa.tf

astianseb:


ludoo:
We actually prefer `for_each` to `count` and do it on purpose. :)

PR #682, opened 2022-06-16
FAST: fix CI/CD source repositories in stage 01

imp14a:
Some changes required in order to enable sourcerepo cicd type for resman stage

Permissions required in bootstrap stage for Source Repository and Cloud Build trigger
Change the SA for sourerepo module in each branch based on the cicd.tf file from bootstrap stage
Adding 'try' statement to outputs.tf, identity_provider value is null for sourcerepo type


ludoo:
Hey Agustin, can you explain the rationale behind the two changes? Why do you need to assign owner role to resource management SA? Why are you changing the trigger service account, which would break impersonation flow in the cloud build workflow?

fast/stages/00-bootstrap/automation.tf
```diff
@@ -36,7 +36,8 @@ module "automation-project" {
   # machine (service accounts) IAM bindings
   iam = {
     "roles/owner" = [
-      module.automation-tf-bootstrap-sa.iam_email
+      module.automation-tf-bootstrap-sa.iam_email,
+      module.automation-tf-resman-sa.iam_email
```
ludoo, line 40: Why is this needed?

ludoo (changes requested):

ludoo:
> Hey Agustin, can you explain the rationale behind the two changes? Why do you need to assign owner role to resource management SA? Why are you changing the trigger service account, which would break impersonation flow in the cloud build workflow?

Sorry, just woke up and needed a coffee :) Thanks for spotting the incorrect trigger service accounts! Let me check if a different role can be assigned to the resource mgmt service account instead of owner, then this is good to go!

fast/stages/00-bootstrap/automation.tf
ludoo: Can you try with ` roles/cloudbuild.builds.editor` instead of owner for stage 1 SA?

ludoo (changes requested):

fast/stages/00-bootstrap/automation.tf
imp14a: Yes, can be this 2 roles: _roles/source.admin, roles/cloudbuild.builds.editor_, but I faced a permission problem: 
_Error 403: generic::permission_denied: user does not have impersonation permission on the trigger service account specified: projects/mm-prod-iac-core-0/serviceAccounts/mm-prod-resman-net-1@mm-prod-iac-core-0.iam.gserviceaccount.com_, 
I tried to impresionate but not sure which SA should be implesionated to generate the Cloud Build trigger, I fixed using the owner role for the resman SA

ludoo: that is weird, to impersonate you need token creator role on the destination SA, which should be there; let me cross-check as soon as I have 10m

ludoo: I removed owner and used roles for cloud build and source repo, but we still have one issue which I'll details below in a separate comment

ludoo:
So, good catch on the wrong service accounts defined in the cloud build triggers for stage 1, and sorry for the confusing review I did this morning. :)

There's one problem left to address, which is a missing role required to use the service accounts in the triggers, which is the reason why you had assigned `roles/owner` initially. We can't use that as it's too broad, and I tried various permutations of roles to both the Cloud Build Service identity and Cloud Build default service account with no success. There's quite a few people having had that issue in the past, we just need to find the role role to the right identity and then we can merge this.

ludoo:
This is probably the doc we should follow, I'll give it a try

https://cloud.google.com/build/docs/securing-builds/configure-user-specified-service-accounts#permissions

ludoo:
The latest commit makes trigger creation work, but introduces two unrelated issues.

The first one is that depending service account module outputs on its iam bindings creates cycles in some of our examples and fast stages, and tests fail; @juliocc and I need to figure out if breaking the module assumptions (removing dependencies) is the only way to address this.

The second issue is that I added a data source to get to the resource management service account email from within the stage, as that is the identity that needs the service account user role on CI/CD service accounts for trigger creation. I don't see many other options, as a) giving the role at the project level in stage 0 has security implications, b) stage 0 has no notion of the CI/CD service accounts created in stage 1. The data source depends on the provider authentication, which we don't do in our tests, so it makes the test fail. Again, I need to discuss this with @juliocc.

Bottom line, this PR works and is needed, but breaks other examples and stages, and the stage1 test. Stay tuned. :)

ludoo (approved):
I addressed the two issues mentioned above by a) exclding the problematic data source when no ci/cd repos are configured, so test work and b) reverting the SA module changes and adding explicit dependencies. We will revisit this but for now I think it's good to go.

If this works for you feel free to merge (or ping me if you don't see a merge button and I'll do it), if you think there are changes needed let's keep working on it.

Thanks *a lot* for spotting the issues and opening this!

imp14a:
Tested and works fine! thanks @ludoo

PR #683, opened 2022-06-16
Binary authorization module and example

apichick:


juliocc (approved):

ludoo:
@apichick I reverted this PR as it was using a git submodule, which breaks sourcing modules from this repo via git, can you resubmit the PR without using a submodule?

PR #685, opened 2022-06-16
Update 01-dropoff.tf

ranadip:
To keep naming of the BQ drop project consistent with rest of the example.


ludoo (approved):

examples/data-solutions/data-platform-foundations/01-dropoff.tf
```diff
@@ -129,7 +129,7 @@ module "drop-sa-bq-0" {
 module "drop-bq-0" {
   source         = "../../../modules/bigquery-dataset"
   project_id     = module.drop-project.project_id
-  id             = "${replace(var.prefix, "-", "_")}drp_bq_0"
+  id             = "${replace(var.prefix, "-", "_")}-drp_bq_0"
```
ludoo, line 132: This needs to only use underscores AFAIK, you need to chage to `"${replace(var.prefix, "-", "_")}_drp_bq_0"`

PR #689, opened 2022-06-17
Binary authorization module and example

apichick:


ludoo (approved):
Miren, if you removed the git submodule this should be good to go.

PR #692, opened 2022-06-20
Fix KMS delegation role

lcaggio:
Add custom role to let 03 stages (example: 03-data-platform) assign roles on KMS keys


ludoo:
Lore, can we explore ways of not having to define a custom role? For keys, having set IAM policy means the ability to self-assign KMS admin, so we might as well use that if we can scope the role to keys via a condition, WDYT?

ludoo:
Cloud KMS Admin scoped on resource type key via a condition should probably get you what you need

https://cloud.google.com/iam/docs/conditions-attribute-reference#resource-type

ludoo (approved):

PR #696, opened 2022-06-22
Added missing image in GLB and Cloud Armor example

apichick:


ludoo (approved):
it does not really match our diagram style, but it's better than having none :)

PR #700, opened 2022-06-23
FAST: GKE multitenant infrastructure

ludoo:
This is part 1 of the GKE multitenant feature, with the infrastructure resources. Cluster-level configurations will follow in a separate PR once this is merged.


fast/stages/03-gke-multitenant/_module/main.tf
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+
+module "gke-project-0" {
+  source          = "../../../../modules/project"
+  billing_account = var.billing_account_id
+  name            = "gke-clusters-0"
+  parent          = var.folder_id
+  prefix          = var.prefix
+  group_iam       = var.group_iam
```
apichick, line 23: There should be a way to grant roles to users too. For instance, Connect Gateway for Google Groups requires having Google Workspace. It might be that the customer only has Cloud Identity.

ludoo: Why should you give roles to users instead of groups? It's an antipattern and we never support it in FAST.

apichick: Also if they want to create a multi-cluster mesh there are also other APIs that need to be enabled in this project

ludoo: CI has groups :)

ludoo: > Also if they want to create a multi-cluster mesh there are also other APIs that need to be enabled in this project

sure, can you add all the APIs that are needed to `main.tf`?

fast/stages/03-gke-multitenant/_module/gke-clusters.tf
```diff
@@ -0,0 +1,121 @@
...
+      key_name = null
+    }
+    : {
+      enabled  = true
+      state    = "ENCRYPTED"
+      key_name = each.value.overrides.database_encryption_key
+    }
+  )
+  default_max_pods_per_node   = each.value.overrides.max_pods_per_node
+  enable_binary_authorization = each.value.overrides.enable_binary_authorization
```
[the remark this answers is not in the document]
ludoo: provider is already bumped up to latest AFAIK, so let's update

tests/modules/gke_hub/test_plan.py
```diff
@@ -40,15 +41,23 @@ def test_configmanagement_setup(resources):
...
+                  'main',
+              'sync_repo':
+                  'https://github.com/danielmarzini/configsync-platform-example',
+              'sync_rev':
+                  None,
+              'sync_wait_secs':
+                  None
           }],
           'prevent_drift': False,
           'source_format': 'hierarchy'
```
apichick, line 63: We don't recommend the use of hierarchy any more but unstractured.

fast/stages/03-gke-multitenant/_module/main.tf
apichick: The user access thing is a limitation of connect gateway
Have a look here.
https://cloud.google.com/anthos/multicluster-management/gateway/setup-groups
According to that you need a Google Workspace edition to be albe to enable the LDAP thing. Customers that use CI will not be able to set it up.

fast/stages/03-gke-multitenant/dev/outputs.tf
```diff
@@ -57,6 +57,7 @@ resource "google_storage_bucket_object" "tfvars" {
 output "clusters" {
   description = "Cluster resources."
   value       = module.gke-multitenant.clusters
+  sensitive   = true
```
ludoo, line 60: let's keep as non sensitive unless we really need to

modules/gke-hub/main.tf
```diff
@@ -53,7 +53,7 @@ resource "google_gke_hub_membership" "default" {
 
 resource "google_gke_hub_feature" "default" {
   provider = google-beta
-  for_each = { for k, v in var.features : k => v if coalesce(v, false) != false }
+  for_each = { for k, v in var.features : k => v if v != false && v != "" }
```
ludoo, line 56: coalesce is needed as the value can be null

modules/gke-hub/variables.tf
```diff
@@ -77,7 +77,7 @@ variable "features" {
     appdevexperience             = false
     configmanagement             = false
     identityservice              = false
-    multiclusteringress          = null
+    multiclusteringress          = ""
```
ludoo, line 80: we never use the empty string, we always use null, this needs to be reverted

fast/stages/03-gke-multitenant/_module/main.tf
```diff
@@ -0,0 +1,78 @@
...
+  labels          = var.labels
+  services = concat(
```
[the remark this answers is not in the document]
ludoo: added a `project_services` variable that allows specifying additional services

fast/stages/03-gke-multitenant/_module/main.tf
ludoo: added a `iam` variable that allows passing users and service accounts for project-level IAM

apichick:
changes lgtm

ludoo:
> changes lgtm

@apichick plz approve ;)

apichick (approved):

fast/stages/03-gke-multitenant/README.md
```diff
@@ -0,0 +1,10 @@
+# GKE Multitenant stage
```
ludoo, line 1: Readme needs to be completed

fast/stages/03-gke-multitenant/_module/README.md
```diff
@@ -0,0 +1,70 @@
+# GKE Multitenant Module
```
ludoo, line 1: Readme needs to be completed

fast/stages/03-gke-multitenant/dev/variables.tf
```diff
@@ -241,6 +241,7 @@ variable "nodepools" {
       node_taints       = list(string)
     })
     preemptible = bool
```
ludoo, line 243: Does it make sense to support both preemptible and spot, or can we just keep spot?

fast/stages/03-gke-multitenant/dev/README.md
```diff
@@ -0,0 +1,78 @@
+# GKE Multitenant
```
ludoo, line 1: README need to be completed

PR #711, opened 2022-06-29
FAST - updated several stage READMEs about usage of *.auto.tfvars files

mgfeller:
Updated information about using the auto generated tfvars files in the "Variable configuration" section for some stages.
Mainly about using `globals.auto.tfvars.json` instead of copying `terraform.tfvars` from bootstrap stage.


ludoo (approved):
Thanks a lot for this!

PR #716, opened 2022-07-05
FAST - added missing format argument in branch-pf-dev-sa-cicd

mgfeller:
In the module branch-pf-dev-sa-cicd, the calls to the "format" function were missing the argument var.automation.federated_identity_pool.


ludoo (approved):
nice catch!

Issue #698, opened 2022-06-22
Migrate use of google_organization_policy to google_org_policy_policy


tszebeni:
While inspecting the organization policy specific module implementation for reuse, I noticed the use of the mentioned resource which is on its terraform documentation mentioned as not recommended to use (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_organization_policy). Better to use the resource google_org_policy_policy which provides more flexibility. Do you consider migrating the module/introduce an alternative?

Please consider, thank you

ludoo:
Thanks for the suggestion it's definitely something we should implement.

PR #722, opened 2022-07-06
OrgPolicy module (factory) using new org-policy API, #698

averbuks:


modules/organization-policy/variables.tf
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+variable "config_directory" {
+  description = "Paths to a folder where organization policy configs are stored in yaml format. Files suffix must be `.yaml`."
+  type        = string
+  default     = null
+}
+
+# TODO: convert to a proper data structure map(map(object({...}))) once tf1.3 is released and optional object keys are avaliable,
```
averbuks, line 23: @ludoo @juliocc WDYT, I fould this as an easiest way to allow uses skip attributes while defining policies (like avoid tons of keys set to null), and the same time there won't be a breaking change when tf1.3 released and we define a proper data structure for the variable. 
Do we want going the same way for org policies in project, folder, org modules?

juliocc: > @ludoo @juliocc WDYT, I found this as the easiest way to allow uses skip attributes while defining policies (like avoid tons of keys set to null), and the same time there won't be a breaking change when tf1.3 released and we define a proper data structure for the variable.

I agree on the null keys point but you know we avoid variables with `type=any`. We actually started thinking how to migrate to the these new org policy resource and put it aside until optionals are GA because we didn't find a reasonable way to keep type information and not forcing the user to pass an object with a ton of null values.

> Do we want going the same way for org policies in project, folder, org modules?

I'd say yes but I can be convinced with good arguments. The main drawback I see with the new approach is that you have to pass the resource id as keys (e.g. the folder id), which will certainly break once you start trying to apply policies to folders/projects created by another module.

That being said, I really like your idea to pass policies as yaml files, +1000 to that. Maybe we need both things (standalone and org/folder/project support)?

@ludoo wdyt?

modules/organization-policy/README.md
```diff
@@ -0,0 +1,166 @@
...
+        "constraints/iam.disableServiceAccountKeyUpload" = {
+          rules = [
+            {
+                enforce = true
+            }
+          ]
+        }
+    },
+    "organizations/1234567890" = {
+      "run.allowedIngress" = {
```
juliocc, line 34: do you have to include "constraints/" or not?

modules/organization-policy/README.md
```diff
@@ -0,0 +1,166 @@
...
+```hcl
+organization_policies = {
+  "parent_id" = { # parent id in format projects/project-id, folders/1234567890 or organizations/1234567890.
+    "policy_name" = { # policy constraint id, for example compute.vmExternalIpAccess.
+      inherit_from_parent = true|false # (Optional) Only for list constraints. Determines the inheritance behavior for this policy.
+      reset               = true|false # (Optional) Ignores policies set above this resource and restores the constraint_default enforcement behavior.
+      rules               = [ # Up to 10 PolicyRules are allowed.
+        {
+          allow_all = true|false # (Optional) Only for list constraints. Setting this to true means that all values are allowed.
+          deny_all  = true|false # (Optional) Only for list constraints. Setting this to true means that all values are denied.
```
juliocc, line 67: I'd probably try to include `allow_all` and `deny_all` under the `values` object. Or perhaps even explode `values` inside the rule object.

modules/organization-policy/README.md
```diff
@@ -0,0 +1,166 @@
...
+├── org-policies
+│   ├── baseline.yaml
+│   ├── image-import-projects.yaml
+│   └── exclusions.yaml
+```
+
+Organization policies example yaml configuration
+
+```bash
+cat ./policies/baseline.yaml
```
juliocc, line 123: what's in the other files?

modules/organization-policy/variables.tf
```diff
@@ -0,0 +1,30 @@
...
+# for now it will cause multiple keys to be set to null for every policy definition
+# https://github.com/hashicorp/terraform/releases/tag/v1.3.0-alpha20220622
+variable "organization_policies" {
```
juliocc, line 26: I'd name this just "policies". Same for the output

modules/organization-policy/variables.tf
```diff
@@ -0,0 +1,30 @@
...
+  description = "Organization policies keyed by parent in format `projects/project-id`, `folders/1234567890` or `organizations/1234567890`."
+  type        = any
```
juliocc, line 28: ☹️

modules/organization-policy/variables.tf
averbuks: @juliocc 
>I agree on the null keys point but you know we avoid variables with type=any. We actually started thinking how to migrate to the these new org policy resource and put it aside until optionals are GA because we didn't find a reasonable way to keep type information and not forcing the user to pass an object with a ton of null values.
That's why I wanted to know what you think, my worry is that tf1.3 is still in alpha stage, but we need to provide users with a way to use conditions and multiple rules for org-policies, so I thought having type=any + a good documentation could be acceptable while we are waiting for optional keys.
> I'd say yes but I can be convinced with good arguments. The main drawback I see with the new approach is that you have to pass the resource id as keys (e.g. the folder id), which will certainly break once you start trying to apply policies to folders/projects created by another module.
There won't be keys with resource IDs for project/org/folder since there will be only one resource which is created by the module. When I'm asking around the same approach I mean the var format and type=any thing, but keys are going to be constants IDs which is static.

modules/organization-policy/README.md
averbuks: No, you don't

modules/organization-policy/README.md
```diff
@@ -0,0 +1,166 @@
...
+  source = "./modules/organization-policy"
+
+  config_directory = "./policies"
+}
+
+# using configuration provided in the module variable
+module "org-policy" {
+  source = "./modules/organization-policy"
+  
+  organization_policies = {
```
ludoo, line 23: I don't much like having an `organization_policies` variable in a module that only deals with organization policies ;)

modules/organization-policy/README.md
```diff
@@ -0,0 +1,166 @@
...
+                enforce = true
+            }
+          ]
+        }
+    },
+    "organizations/1234567890" = {
+      "run.allowedIngress" = {
+        rules = [
+          {
+            condition = {
```
ludoo, line 37: This is a bit complex to use, at this point what's the benefit of using a module, versus using individual resources? Can we find a way of making the interface simple?

modules/organization-policy/README.md
ludoo: Yep, I would try and simplify the interface (see my other comment above).

averbuks: I'm trying to simply keep it consistent with the API itself, so may just check the API reference or the resource itself](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/org_policy_policy#argument-reference) and understand how to fill up the var. But if you think we may simplify the interface, I'm open to options.

modules/organization-policy/README.md
averbuks: The same, you may just differentiate into multiple files like
org-level.yaml
folder-level.yaml
or env based, it's simply for a better navigation and helps keep PRs more readable.

modules/organization-policy/variables.tf
ludoo: I absolutely agree we should not have `any` as a variable type. I would

- try to simplify the interface further
- come up with proper types
- possibly start implementing optionals in this module, it's new we can require preview features on this, we're not breaking compatibility

+1 on yaml files, Aleks you're the father of factories, it makes sense it's you to introduce it for org policies! :)

modules/organization-policy/variables.tf
averbuks: +1, will do

modules/organization-policy/README.md
averbuks: I will go for `policies`

modules/organization-policy/README.md
averbuks: I will look into some options, initially I thought about simply following the API spec which makes it simpler for a developer to match and fill.

averbuks: I would probably avoid `values` block and remove `allow_all`/`deny_all`
```
rules = [
  {
  allow = [] # stands for all
  },
  {
    deny = ["some_value"] # only some values
  }
]
```
Don't see a way to simplify `condition` block tho

modules/organization-policy/variables.tf
averbuks: Experimental feature used to avoid `type=any`

modules/organization-policy/variables.tf
averbuks: done

modules/organization-policy/variables.tf
averbuks: Fixed by introducing `experimentals` for the provider configuration

modules/organization-policy/README.md
```diff
@@ -0,0 +1,167 @@
...
+}
+# tftest skip
+```
+
+### Structure of configuration provided in a yaml file/s
+
+Configuration should be placed in a set of yaml files in the config directory. Policy entry structure as follows:
+
+```yaml
+parent_id: # parent id in format projects/project-id, folders/1234567890 or organizations/1234567890.
```
ludoo, line 97: I would put the parent id in the file name, so that one file maps to a single parent only. Managing multiple parents in a single file makes it harder to manage this.

Just my personal PoV so feel free to ignore. :)

ludoo (approved):
very nice module Aleks :)

modules/organization-policy/README.md
averbuks: Fixed

PR #727, opened 2022-07-08
apigee-x-instance.ip_range cannot take Support CIDR

alexlo03:
Current docs are incorrect. It is not possible to specify the Support CIDR on instance create.

See https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.instances#Instance

> ipRange	
> string
> Optional. IP range represents the customer-provided CIDR block of length 22 that will be used for the Apigee instance creation. This optional range, if provided, should be freely available as part of larger named range the customer has allocated to the Service Networking peering. If this is not provided, Apigee will automatically request for any available /22 CIDR block from Service Networking. The customer should use this CIDR block for configuring their firewall needs to allow traffic from Apigee. Input format: "a.b.c.d/22", Output format: a.b.c.d/22, e.f.g.h/28"


ludoo:
Thanks for the PR! Could you run `tools/tfdoc` on the module folder to regenerate the variables table, as explained in our [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)?

ludoo (approved):

PR #745, opened 2022-07-19
FIX: 00-cicd stage - Gitlab and Github providers

imp14a:


imp14a:
To solve error in the terraform init for 00-cicd stage

│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/gitlab: provider registry registry.terraform.io does not have a provider named
│ registry.terraform.io/hashicorp/gitlab
│ 
│ Did you intend to use gitlabhq/gitlab? If so, you must specify that source address in each module which requires that provider. To see which modules are currently
│ depending on hashicorp/gitlab, run the following command:
│     terraform providers

PR #746, opened 2022-07-20
Update multi region cloud SQL markdown file

bensadikgoogle:
Update documentation for the "Cloud SQL multi-regional" architecture example with  a detailed tutorial


ludoo:
Thanks for this! Can you provide a description of what it does?

bensadikgoogle:
updated the description !! 👍

ludoo (approved):
Once you fix linting this is good to go, many thanks! To fix linting you need to run our `tfdoc` tool against the directory, it will generate a new variable table. Those are not supposed to be edited by hand, and it's why the test fails. Instructions are in our top-level contributing doc.

PR #759, opened 2022-07-26
FIX: Missing value to format principalSet

imp14a:


imp14a:
To fix error in resman stage

Error: Error in function call

  on cicd-data-platform.tf line 112, in module "branch-dp-dev-sa-cicd":
 112:         ? format(
 113:           local.identity_providers[each.value.identity_provider].principalset_tpl,
 114:           each.value.name
 115:         )
    ├────────────────
    │ each.value.identity_provider is "github-fip"
    │ each.value.name is "imp14a/fast-data-platform-dev"
    │ local.identity_providers is map of object with 1 element

Call to function "format" failed: not enough arguments for "%s" at 58: need
index 2 but have 1 total.

ludoo (approved):

ludoo:
thanks for spotting this!

PR #760, opened 2022-07-28
GKE hub module refactor

ludoo:


apichick (approved):

apichick:
It fails because we require google-beta provider v4.25.0 @ludoo not sure exactly where to change the version, tried to do it in the versions.tf of the module, but didn't work and I was scared of touching the defaults.tf :)

PR #770, opened 2022-08-07
fixing to move without output_location

daisuky-jp:
01-resman failed when outpus_location was not specified. In case of using Cloud Build with SoureRepo for runninc terraform commands, outputs_location is unnecessary but the line 36 fails due to the null value. Need to have null check at the beginning.


ludoo (approved):

ludoo:
thanks for spotting this!

PR #774, opened 2022-08-15
FAST: fix data-platform-dev folder in stage 03-data-platform

sttomm:
I did run all Fabric Fast steps and noticed an issue when applying the 03-data-platform stage. It references to a non-existing folder. The 01-resman stage creates specific data-platform-dev and data-platform-prod folders and writes them to the resulting tfvars file. As this file is used as input to stage 03-data-platform, the specific data-platform-dev folder must be referenced, but actually only a generic data-platform folder, which does not exist, was referenced.

With this PR i fixed it to use the existing data-platform-dev folder.


ludoo (approved):

ludoo:
Hey Stefan, thanks for this! Could you merge master in, or give us permissions to do so? Linting and tests also need attention, you can check our contributing guide, or we can do it once we can push to your fork.

sttomm:
Hi Ludovico, thanks for the quick review! I think i fixed the linting and the tests. I also rebased to master. Just wasn't able to execute pytest locally due to some errors.

You also have access to our repo now and could also apply changes if needed.

ludoo:
Thanks Stefan, all good! I think you can mark forks as accessible from the forked repo admins when you create the fork, no need to share individually next time. :)

PR #775, opened 2022-08-17
net-glb: Added support for regional external HTTP(s) load balancing

rosmo:
Added tests too.


modules/net-glb/backend-services.tf
```diff
@@ -21,10 +21,14 @@ locals {
     for k, v in coalesce(var.backend_services_config, {}) :
     k => v if v.bucket_config != null
   }
-  backend_services_group = {
+  backend_services_group = var.region == null ? {
     for k, v in coalesce(var.backend_services_config, {}) :
     k => v if v.group_config != null
-  }
+  } : {}
+  regional_backend_services_group = var.region != null ? {
```
ludoo, line 28: nit: can we use a single local, and apply the condition where it's set in the resource?

modules/net-glb/backend-services.tf
```diff
@@ -208,3 +212,150 @@ resource "google_compute_backend_service" "group" {
     }
   }
 }
+
```
ludoo, line 215: I would move the regional resources to a `regional-backend-services.tf` file

ludoo (changes requested):
Can you add a section with usage example to the README?

modules/net-glb/backend-services.tf
rosmo: Should I move all regional resources to `regional-` files? My original idea was to keep the resources in same files where it made sense (eg. the file is short enough for comprehension) given the resource config is almost 1:1.

ludoo: Hmmm, maybe only for "fat files" like backends then?

ludoo (approved):

modules/net-glb/backend-services.tf
rosmo: I moved the backends to regional file.

modules/net-glb/backend-services.tf
rosmo: Fixed!

rosmo:
> Can you add a section with usage example to the README?

Added a small section with an example.

Issue #777, opened 2022-08-20
Add the ability to set instance_termination_action to DELETE for Spot VMs


sruffilli:
The [Spot VM lifecycle](https://cloud.google.com/compute/docs/instances/create-use-spot?hl=en) allows for either stop or delete the VM when the resource is pre-empted. `compute-vm` doesn't currently support configuring this option.

PR #778, opened 2022-08-20
instance_termination_action must be set for compute-vm spot instances

sruffilli:
`google_compute_instance` and `google_compute_instance_template` having `provisioning_model = SPOT` must have `scheduling.instance_termination_action` set to either `STOP` or `DELETE`.

Even if the field is optional, not having it set triggers the following behaviors:
- `google_compute_instance` -> state flapping
- `google_compute_instance_template` -> state flapping triggering an error, as update is not supported.

This PR sets `scheduling.instance_termination_action=STOP` (which is the default).
#777 tracks the ability to optionally customize instance_termination_action


ludoo:
Can you also address #777 here?

modules/compute-vm/main.tf
```diff
@@ -212,10 +212,11 @@ resource "google_compute_instance" "default" {
...
   scheduling {
-    automatic_restart   = !var.options.spot
-    on_host_maintenance = local.on_host_maintenance
-    preemptible         = var.options.spot
-    provisioning_model  = var.options.spot ? "SPOT" : "STANDARD"
+    automatic_restart           = !var.options.spot
+    on_host_maintenance         = local.on_host_maintenance
+    preemptible                 = var.options.spot
+    provisioning_model          = var.options.spot ? "SPOT" : "STANDARD"
+    instance_termination_action = var.options.spot ? "STOP" : null
```
ludoo, line 219: can you order alphabetically?

modules/compute-vm/variables.tf
```diff
@@ -187,14 +187,22 @@ variable "network_interfaces" {
 variable "options" {
   description = "Instance options."
   type = object({
-    allow_stopping_for_update = bool
-    deletion_protection       = bool
-    spot                      = bool
+    allow_stopping_for_update        = bool
+    deletion_protection              = bool
+    spot                             = bool
+    spot_instance_termination_action = string
```
ludoo, line 193: can we rename to `instance_termination_action` like in the resource, or just `termination_action`?

ludoo (approved):

PR #783, opened 2022-08-23
fix service unit indent on cloud-config-container module

drebes:


drebes:
don't approve/merge it yet, i'm still sorting out some issues with newlines vs indents.

drebes:
This should be good to go now.

ludoo (approved):

PR #784, opened 2022-08-24
fix envoy-traffic-director config for xDS v3

drebes:


ludoo:
It would be nice having a few lines in the README explaining the intended use, I could never wrap my head around this. :)

ludoo (approved):

drebes:
Ludo, I'll have an end-to-end example using this, and that will have a good description :). But on another PR. :D

PR #785, opened 2022-08-24
nginx-tls module

drebes:


juliocc (approved):

lcaggio:
I recently had the need to configure a TCP reverse_proxy with NGINX. 

```
# nginx.conf
events {}
stream {
    server {
        listen     443;
        proxy_pass $URL:443;
    }
}
```

Can we make the NGINX configuration a parameter that the user can specify?

lcaggio:
This can be achieved using the standard nginx module, please ignore.

PR #787, opened 2022-08-31
Support manager role in cloud identity group module

lcaggio:
Add support for MANAGERS role.
Associated bug is fixed: https://github.com/hashicorp/terraform-provider-google/issues/7616


sruffilli:
Hey Lorenzo - just to be on the safe side, have you tried re-triggering the aforementioned issue?

lcaggio:
I did the following test:
 - Created a group with 1 MEMBER and 1 MANAGER
 - Moved the MANAGER user to MEMEBR
 - Moved back the MEMBER to MANAGER

The test worked with no error. Does it trigger the use case mentioned in the issue?

Additional note: since the MANAGER is also a MEMBER, I added the depends_on attribute you find [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/a2534464060cdf58d1cad765ea29d66053832b9c/modules/cloud-identity-group/main.tf#L52).

sruffilli (approved):

PR #803, opened 2022-09-09
Fix automation gcs location

miklosn:
Currently the creation of gcs buckets fails:
- regional vs. multi_regional condition is inverted
- it's MULTI_REGIONAL (with underscore)


ludoo (approved):
Thanks for this! Can you also copy/paste to stage 01?

PR #805, opened 2022-09-09
Change `modules/project` service_config default

juliocc:
this commit:
- Disables API deactivation by default on project destruction
- Adds "jit" service accounts as explicit dependency to the `project_id` output
- Makes `project_id` and `number` outputs depend_on the same resources


ludoo (approved):
Oh God, yes!

PR #820, opened 2022-09-15
New architecture: org audit log routing to BQ 

bensadikgoogle:
@erivarola for visibility


ludoo:
Can you please resolve conflicts, and make it so you're not renaming hundreds of unrelated files?

PR #823, opened 2022-09-16
New architecture: org audit log routing to BQ

bensadikgoogle:
@erivarola for visibility


blueprints/cloud-operations/org_audit_logs_bq/main.tf
```diff
@@ -0,0 +1,76 @@
...
+      name   = exclusion.key
+      filter = exclusion.value
+    }
+  }
+  bigquery_options {
+    use_partitioned_tables = true
+  }
+}
+
+resource "google_project_iam_member" "org_sa_bq_role" {
```
juliocc, line 60: you cane use `iam_additive` variable in the project module for this

juliocc (changes requested):

ludoo (commented):
I'm not too sure this makes sense as an example: we already have similar things in the log module, and in FAST. It's also super trivial to assemble. What is the added value you think this bring to our users?

PR #826, opened 2022-09-19
Added new binary_authorization argument in gke-cluster module

sirohia:
Added updated binary_authorization argument as per below url

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#binary_authorization

### DIFF
--- a/modules/gke-cluster/main.tf
+++ b/modules/gke-cluster/main.tf
@@ -292,6 +292,13 @@ resource "google_container_cluster" "cluster" {
     }
   }
 
+  dynamic "binary_authorization" {
+    for_each = var.enable_binary_authorization == true ? [""] : []
+    content {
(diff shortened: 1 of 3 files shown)


modules/gke-cluster/main.tf
```diff
@@ -292,6 +292,13 @@ resource "google_container_cluster" "cluster" {
     }
   }
 
+  dynamic "binary_authorization" {
+    for_each = var.enable_binary_authorization == true ? [""] : []
```
juliocc, line 296: the `== true` is redundant. Can you make the condition just `var.enable_binary_authorization`?

juliocc (approved):

modules/gke-cluster/main.tf
sirohia: Sure. I have removed ==true

juliocc:
Thanks @sirohia!

PR #830, opened 2022-09-20
Add ability to add KMS encryption to google_compute_instance_template

bjbloemker-google:
Adding `disk_encryption_key` block to `google_compute_instance_template`

This is already possible when creating a single instance, but adding to compute instance template as well.

ref: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance_template#disk_encryption_key


ludoo:
closing this and reopening so we can fix/merge

PR #832, opened 2022-09-21
Addedd Ingress for ILB (IFI) support

astianseb:
Feature: https://cloud.google.com/kubernetes-engine/docs/concepts/ingress-ilb
Current gke-cluster module does not support it and in order to create Ingress for ILB there's a need to create two additional resources: subnet & firewall. Having it supported in a gke-cluster module create cleaner config.

### DIFF
--- a/modules/gke-cluster/README.md
+++ b/modules/gke-cluster/README.md
@@ -106,6 +106,9 @@ module "cluster-1" {
 | [resource_usage_export_config](variables.tf#L288) | Configure the ResourceUsageExportConfig feature. | <code title="object&#40;&#123;&#10;  enabled &#61; bool&#10;  dataset &#61; string&#10;&#125;&#41;">object&#40;&#123;&#8230;&#125;&#41;</code> |  | <code title="&#123;&#10;  enabled &#61; null&#10;  dataset &#61; null&#10;&#125;">&#123;&#8230;&#125;</code> |
 | [vertical_pod_autoscaling](variables.tf#L315) | Enable the Vertical Pod Autoscaling feature. | <code>bool</code> |  | <code>null</code> |
 | [workload_identity](variables.tf#L321) | Enable the Workload Identity feature. | <code>bool</code> |  | <code>true</code> |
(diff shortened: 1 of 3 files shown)


ludoo:
Hey, thanks for this! I'm not sure I like having the GKE module manage VPC resources: this will require extra IAM roles, and conflict with Terraform setups managed by network teams. In general we always try to focus a module on a single resource and its immediate dependencies (IAM, policies) and nothing else.

@juliocc WDYT?

juliocc:
> Hey, thanks for this! I'm not sure I like having the GKE module manage VPC resources: this will require extra IAM roles, and conflict with Terraform setups managed by network teams. In general we always try to focus a module on a single resource and its immediate dependencies (IAM, policies) and nothing else.
> 
> @juliocc WDYT?

Agreed. The proposed change would require permissions tu create subnets and firewall rules on the VPC. The gke module shouldn't be responsible for those tasks.

astianseb:
Fair comment with separation of duties, therefore default is "false". If someone wants to leverage it anyhow, for test/dev - there's ready made instrumentation in the module.

ludoo:
Would it make sense to transform this into a blueprint example? Or even a doc example in the module's README.

astianseb:
I can do a special description in a module README on how to use and when to use it. Default is FALSE therefore its not breaking anything in  a separation of duties - but it's handy and helps in a prototyping.

ludoo:
> I can do a special description in a module README on how to use and when to use it. Default is FALSE therefore its not breaking anything in a separation of duties - but it's handy and helps in a prototyping.

I mean, could we add an example to the module's README, showing how to do this in combination with other modules, given we don't want to include extraneous resources in the module itself?

juliocc:
> I can do a special description in a module README on how to use and when to use it. Default is FALSE therefore its not breaking anything in a separation of duties - but it's handy and helps in a prototyping.

You can create an example in the README where you use the gke module to bring up the cluster and use the native modules to create the 2 other resources (e.g. the subnet using `net-vpc` and the the firewall rule using `net-vpc-firewall`)

astianseb:
This is how I did it initially when my Ingress for ILB provisioning failed. I reverted back to documentation and it turned out that I need a "proxy-only" subnet and a firewall rule. Decided that it will be cool to have this built into the gke-cluster module just for convenience.

PR #839, opened 2022-09-28
Update to terraform 1.3

juliocc:
This PR fixes terrafom 1.2 -> 1.3 incompatibilities and updates all versions.tf


ludoo (approved):
we probably need to update the CI pipelines to use the new version

PR #840, opened 2022-09-28
Refactor net-address module for 1.3

ludoo:


sruffilli (approved):
LGTM - we might consider in the future unifying all types of supported addresses into a single attribute. Or not :)

PR #844, opened 2022-09-29
Management of GCP project default service accounts

ddaluka:
Project module creates default service account for (depending on the services enabled) for compute and app services and provided the Editor role by default . [Doc here](https://cloud.google.com/iam/docs/service-accounts#default)

To control the high privileged role, we can use terraform resource 'google_project_default_service_accounts'. I have added this resource in project module with an input variable to provide the setting for Keep, Deprivilege, disable or delete.

### DIFF
--- a/modules/project/variables.tf
+++ b/modules/project/variables.tf
@@ -46,6 +46,12 @@ variable "descriptive_name" {
   default     = null
 }
 
+variable "default_service_account" {
+  description = "Project default service account setting: can be one of `delete`, `deprivilege`, `disable`, or `keep`."
+  default     = "keep"
+  type        = string
+}
+
 variable "group_iam" {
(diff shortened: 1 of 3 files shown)


modules/project/variables.tf
```diff
@@ -46,6 +46,12 @@ variable "descriptive_name" {
   default     = null
 }
 
+variable "default_service_account_setting" {
```
ludoo, line 49: nit: would `default_service_accounts` work the same while being shorter?

ludoo (approved):
Really nice, thanks a lot for this.

modules/project/variables.tf
ddaluka: yes . you mean, just rename the variable name to be shorter, right?

ludoo: yep! not critical, but the shorter the better in general.

ddaluka: Done

Issue #817, opened 2022-09-15
Network Dashboard function is getting Timeout


som-nitjsr:
I am using the network dashboard function for generating the custom dashboard related Peering and ILB rules.
Since I have large no of Peering VPC it is getting timeout. I have increased the timeout to 9 minutes. Please suggest how to proceed on this.

aurelienlegrand:
Hi, 

Are you still facing timeouts after increasing it to 9 minutes?
9 minutes is the maximum timeout value for Cloud Functions v1. 

I am planning to move to Cloud Functions v2, which will provide up to 60 minutes timeout and would be better for large GCP deployments: https://cloud.google.com/functions/docs/configuring/timeout.

In the meantime that I deploy this, you can try by yourself using the same python code and using this terraform resource to deploy the cloud function v2: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions2_function.

Hope that helps,
Aurélien

som-nitjsr:
Thanks i have already moved to cloud function version 2

PR #849, opened 2022-09-30
Moved cloud function to 2nd genration

som-nitjsr:
Converting the cloud function to 2nd generation and using http trigger as in case of enterprise there are lot s of data and it takes lots of time . in MY case function was getting timed out. so moved it to second generation where http trigger has 60 minutes of time out. 
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/817


modules/cloud-function/variables.tf
```diff
@@ -104,7 +104,7 @@ variable "project_id" {
 variable "region" {
   description = "Region used for all resources."
   type        = string
-  default     = "europe-west1"
+  default     = "asia-south1"
```
ludoo, line 107: we always default to a European region across Fabric :)

modules/cloud-function/main.tf
```diff
@@ -54,79 +54,49 @@ resource "google_vpc_access_connector" "connector" {
   network       = var.vpc_connector_config.network
 }
 
-resource "google_cloudfunctions_function" "function" {
```
ludoo, line 57: There should a module-level variable to control if v1 or v2 gets deployed.

ludoo (changes requested):
Thanks for this. We should still support v1 in the module.

PR #850, opened 2022-09-30
Made sample alert creation optional

maunope:
- Made sample alert creation optional
- Updated doc to explain alert can be created only after CF has run at least once


maunope:
linting failing with error code 3, not sure about that the catch is

ludoo (approved):
Once you fix linting errors (hint: read up on `tfdoc` in our contributing guide), this is good to go, thanks Mauri!

PR #860, opened 2022-10-06
Refactor compute-vm for Terraform 1.3

ludoo:
This PR takes a minimal approach, so as to preserve the original interface as much as possible while still allowing to omit most non-required attributes. Tests pass, some additional tests might be needed if we want to have full coverage of the interface.


.github/workflows/tests.yml
```diff
@@ -30,7 +30,7 @@ env:
   PYTEST_ADDOPTS: "--color=yes"
   PYTHON_VERSION: "3.10"
   TF_PLUGIN_CACHE_DIR: "/home/runner/.terraform.d/plugin-cache"
-  TF_VERSION: 1.3.0
```
juliocc, line 33: Can you update the version in linting.yml too?

juliocc (approved):

.github/workflows/tests.yml
ludoo: I will send a separate PR updating versions and CI

PR #865, opened 2022-10-07
Enable FAST 00-cicd provider test

ludoo:


.github/workflows/tests.yml
```diff
@@ -155,8 +155,9 @@ jobs:
 
       - name: Pin provider versions
         run: |
-          sed -i 's/>=\(.*# tftest\)/=\1/g' default-versions.tf
-          find -name versions.tf -exec cp default-versions.tf {} \;
+          for f in $(find . -name versions.tf); do
+            sed -i 's/>=\(.*# tftest\)/=\1/g' $f;
+          done
```
juliocc, line 167: Horrible horrible.
For the record: `gfind . -name versions.tf -print0 | xargs -0 gsed -i 's/>=\(.*# tftest\)/=\1/g'`

juliocc (approved):

.github/workflows/tests.yml
ludoo: So basically Perl FTW 😂

PR #857, opened 2022-10-05
added support 2nd generation cloud function

som-nitjsr:
adding support for 2nd generation  cloud function  and using http trigger as in case of enterprise there are lot  of data and it takes lots of time . In my case function was getting timed out. so moved it to second generation where http trigger has 60 minutes of time out.
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/817

PR #872, opened 2022-10-11
added support 2nd generation cloud function 

som-nitjsr:
adding support for 2nd generation cloud function and using http trigger as in case of enterprise there are lot of data and it takes lots of time . In my case function was getting timed out. so moved it to second generation where http trigger has 60 minutes of time out.
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/817

This is Related to PR
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/857


juliocc:
Couple more comments:

- The function resource for version 1 creates event_triggers, secret_environment_variables, and secret_volumes blocks. Can you ensure the version 2 resource considers those elements too?
- In the dashboard you use an http trigger when you enable the version 2 function. Why don't you use the scheduler+pubsub topic already in place for the v1 function?

som-nitjsr:
> Couple more comments:
> 
> * The function resource for version 1 creates event_triggers, secret_environment_variables, and secret_volumes blocks. Can you ensure the version 2 resource considers those elements too?
> * In the dashboard you use an http trigger when you enable the version 2 function. Why don't you use the scheduler+pubsub topic already in place for the v1 function?

First one i will add.
for second i thought since people will mostly go for timeout issue and only http has timeout of 60 minutes other triggers still has the same timeout of 10 minute. and also we need to create the EventArc for pub sub trigger. if you still want i can fix that.

juliocc:
> First one i will add. for second i thought since people will mostly go for timeout issue and only http has timeout of 60 minutes other triggers still has the same timeout of 10 minute. and also we need to create the EventArc for pub sub trigger. if you still want i can fix that.

Cool. Let's start with the first one for now

som-nitjsr:
> > First one i will add. for second i thought since people will mostly go for timeout issue and only http has timeout of 60 minutes other triggers still has the same timeout of 10 minute. and also we need to create the EventArc for pub sub trigger. if you still want i can fix that.
> 
> Cool. Let's start with the first one for now

done

juliocc (approved):

juliocc:
@som-nitjsr thanks for the PR and following through with all the requests.

PR #881, opened 2022-10-14
Run tests in parallel using `pytest-xdist`

ludoo:
Testing what you suggested Julio.


ludoo:
not difference in run times, closing this

ludoo:
let's give one more try to pytest-xdist

juliocc (approved):

PR #897, opened 2022-10-19
Project-factory: allow folder_id to be defined in defaults_file

Malet:
Currently, setting the `folder_id` in `defaults.yaml` but not in the `project.yaml`, results in:
```
╷
│ Error: Unsupported attribute
│ 
│   on main.tf line 44, in module "projects":
│   44:   folder_id              = each.value.folder_id
│     ├────────────────
│     │ each.value is object with 1 attribute "labels"
│ 
│ This object does not have an attribute named "folder_id".
╵
```
This change allows you to set the folder_id in the `defaults.yaml` and have it be used for all the projects selected by the glob pattern.


ludoo (approved):

juliocc:
Thanks @Malet!

PR #904, opened 2022-10-21
Add missing description field

dsbutler101:
Important so we can flag these groups are TF managed and must not be manually updated.


ludoo (approved):

juliocc:
Good to see you over here @dsbutler101 ;)

PR #905, opened 2022-10-24
Support separate SA for the network dashboard scheduler

marcos-uy:
Date: 2022-10-24

add terraform changes to blueprints/cloud-operations/network-dashboard to support custom service accounts. The requirements for the cloud functions is to have Org level scope, if a custom SA is provided on tfvars, the create SA module will not execute.


blueprints/cloud-operations/network-dashboard/variables.tf
```diff
@@ -81,3 +81,13 @@ variable "schedule_cron" {
...
+
+variable "custom_cloud_function_sa" {
+  description = "Preprovisioned custom SA. If one is provided, no new SA will be created."
+  default = ""
+}
+
+variable "custom_scheduler_sa" {
+  description = "Preprovisioned custom SA. If one is provided, no new SA will be created."
+  default = ""
+}
```
juliocc, line 93: Can you make this a single variable with `default=null`? If not provided, you create the SA, otherwise you use the one provided.

juliocc (changes requested):
Hi Oscar. Thanks for this PR. Can you fix the linting errors and address the comment?

blueprints/cloud-operations/network-dashboard/variables.tf
marcos-uy: yes sure, however, there was a recent change to the main.tf to add the creation of a second SA for the scheduler, before, both the Cloud Function and the Scheduler for the Network Dashboard used the same SA.

juliocc: I think that change is unrelated. The ask here is two consolidate the two variables into a single one. As far as I can tell you don't need two variables to trigger the creation of the SA

juliocc: Ignore me, apparently I can't read

blueprints/cloud-operations/network-dashboard/main.tf
```diff
@@ -42,6 +42,7 @@ module "project-monitoring" {
 ################################################
 
 module "service-account-function" {
+  count           = var.monitoring_project_id == "" ? 1 : 0
```
aurelienlegrand, line 45: Should this be 
var.custom_cloud_function_sa == "" ? 1 : 0
instead?

blueprints/cloud-operations/network-dashboard/main.tf
```diff
@@ -126,7 +128,7 @@ resource "google_cloud_scheduler_job" "job_httptrigger" {
     uri         = module.cloud-function.uri
 
     oidc_token {
-      service_account_email = module.service-account-scheduler.email
+      service_account_email = var.custom_scheduler_sa == "" ? module.service-account-scheduler.email : var.custom_scheduler_sa
```
aurelienlegrand, line 131: if var.custom_scheduler_sa is provided, this line will never be executed anyway (as the count in this block will be 0). So I think this condition is not required here?

marcos-uy:
I need to close this PR since I'm having issues with the CLA. After that is clear I will open a new PR and try to submit some of the changes I have been working on.

PR #913, opened 2022-10-25
Adding support for PSA ranges, starting with Redis instances.

aurelienlegrand:
Refactoring subnets.py and adding support for PSA ranges.
Implementing support for Redis instances.
Next step will be to count Cloud SQL instances in a similar way.

### DIFF
--- a/blueprints/cloud-operations/network-dashboard/cloud-function/metrics/subnets.py
+++ b/blueprints/cloud-operations/network-dashboard/cloud-function/metrics/subnets.py
@@ -25,7 +25,6 @@ import ipaddress
 def get_all_subnets(config):
   '''
     Returns a dictionary with subnet level informations (such as IP utilization)
-
       Parameters:
         config (dict): The dict containing config like clients and limits
       Returns:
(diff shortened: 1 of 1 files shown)


blueprints/cloud-operations/network-dashboard/cloud-function/metrics/subnets.py
```diff
@@ -214,9 +238,90 @@ def compute_subnet_utilization(config, all_subnets_dict):
...
+
+    # Only handling PSA for Redis for now
+    if connect_mode == "PRIVATE_SERVICE_ACCESS":
+      # Reddis instance asset doesn't contain the subnet information in Asset Inventory
+      # We need to find the correct subnet with IP address matching
+      redis_ip_range = ipaddress.ip_network(ip_range)
+      for subnet_key, subnet_dict in all_subnets_dict[project_id].items():
+        if subnet_dict["network_name"] == network_name:
+          if redis_ip_range.overlaps(
+              ipaddress.ip_network(subnet_dict['ip_cidr_range'])):
```
gleichda, line 300: I'm a bit confused.
Pls help me to understand why this will be only counted if it overlaps.
Maybe a short comment would also be good

blueprints/cloud-operations/network-dashboard/cloud-function/metrics/subnets.py
```diff
@@ -214,9 +238,90 @@ def compute_subnet_utilization(config, all_subnets_dict):
...
+            all_subnets_dict[project_id][subnet_key][
+                'used_ip_addresses'] += redis_ip_range.num_addresses
```
gleichda, line 302: Do we really know if Redis will consume all of these IP Addresses or just needs them in case of?

blueprints/cloud-operations/network-dashboard/cloud-function/metrics/subnets.py
aurelienlegrand: In the Redis asset in Cloud Asset Inventory, I only have the network, not the PSA range info.
To find the PSA range info, I loop through all subnets (includes PSA ranges) in the project and compare the IP range to the redis instance IP range. If it overlaps, then it is the PSA range used for this redis instance...

I will add an additional comment to explain this.

gleichda: Okay wow that's an 'interesting' way of getting there, but got your point! Definitely a comment is required ;)

blueprints/cloud-operations/network-dashboard/cloud-function/metrics/subnets.py
aurelienlegrand: I don't think Redis will consume all these IPs but they are reserved and cannot be used for other services. So for now, I will assume they are not available (even if that is a bit conservative).

gleichda (approved):

juliocc (approved):

PR #916, opened 2022-10-25
Add support for DNS routing policies

juliocc:


modules/dns/variables.tf
```diff
@@ -45,7 +45,9 @@ variable "dnssec_config" {
       { algorithm = "rsasha256", key_length = 1024 }
     )
   })
-  default = null
+  default = {
```
ludoo, line 48: we should add `nullable = false` if we don't want null; I kind of like null as the default, it matches what we do in many other modules which logically means "I don't have it since I don't care about it" :)

ludoo (approved):
nit on dnssec var default

PR #919, opened 2022-10-25
Rename workflow names

juliocc:
Renaming some workflows to align CI to #808 
doc-examples -> examples
examples -> blueprints


ludoo (approved):
I always postponed as you need to update branch protection rules, and sometimes the new names don't show up for a while. Let's go for it. :)

PR #921, opened 2022-10-26
Align documentation, move glb blueprint

ludoo:
This aligns our README files so that modules and blueprints are contained in top-level documentation. It also moves the GLB+Cloud Armor example from the Cloud Operations to the Networking section.


juliocc (approved):
what are you procrastinating today? ;)

ludoo:
> what are you procrastinating today? ;)

looool, actually lunch :)

PR #923, opened 2022-10-27
Fix service account creation error in gke nodepool module

ludoo:
This changes variables so that count does not depend on the SA email, avoiding this error when an external SA is created in the same root module

```
Error: Invalid count argument

  on ../../../../modules/gke-nodepool/main.tf line 61, in resource "google_service_account" "service_account":
  61:   count        = local.service_account_create ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```


juliocc (approved):

modules/gke-nodepool/variables.tf
```diff
@@ -158,10 +158,12 @@ variable "reservation_affinity" {
...
   description = "Nodepool service account. If this variable is set to null, the default GCE service account will be used. If set and email is null, a service account will be created. If scopes are null a default will be used."
   type = object({
-    email        = optional(string)
-    oauth_scopes = optional(list(string))
+    create       = optional(bool, false)
+    email        = optional(string, null)
+    oauth_scopes = optional(list(string), null)
   })
-  default = null
+  default  = {}
```
juliocc, line 165: What do you think `create=optional(bool, true)` and the `default={create=false}`

This way setting `service_account = {}` creates the service account. Currently you have to use `service_account = { create = true }` (see examples, and blueprints changes)

ludoo: yep, thanks for doing this got sidetracked by a meeting ;)

ludoo: Depends on what we want as a default: the implicit default till now was to use the default GCE SA, with the new format you would need `{create=false, email=null}` to achieve the same. Which kind of makes sense, as we would make it harder to do the "wrong" thing. Except on autopilot, where a pending provider bug ignores whatever you pass in and always uses the GCE default.

ludoo: it would also have the side effect of always requiring create=false for existing SAs, which is kind of counterintuitive; for me creation should be explicit, non-creation the default or you can very easily get into weird situations

PR #925, opened 2022-10-27
Updating main.tf and README following #922

brianhmj:
Adding a Terraform provider block and edited the comment in Main.tf 
Added quick note and next step action item in README


blueprints/cloud-operations/network-dashboard/main.tf
```diff
@@ -13,7 +13,16 @@
...
-
+     
+terraform {
+  required_version = "~> 1.3.3"
+  required_providers {
+    google = {
+      version = "~> 4.41.0"
+    }
+  }
+}
```
juliocc, line 24: We typically copy `default-versions.tf` at the root of the repo and put it in `versions.tf` inside the module/blueprint. Can you do that here too?

juliocc (changes requested):

blueprints/cloud-operations/network-dashboard/main.tf
brianhmj: Yes, just made the fixes.

juliocc (approved):

PR #931, opened 2022-10-30
Refactor compute-mig module for Terraform 1.3

ludoo:
This heavily refactors the compute-mig module for tf 1.3, adds support for regional per-instance stateful policies, and implements support for all features in the latest provider.


juliocc (commented):
coming late but lgtm ;)

PR #945, opened 2022-11-03
Org policy factory

juliocc:


ludoo (approved):
Worth adding the yaml schema to the FAST assets?

juliocc:
> Worth adding the yaml schema to the FAST assets?

ah great idea. Let me open a new PR

PR #948, opened 2022-11-03
Use display_name instead of description for FAST service accounts

### LINKED ISSUE #944: Service accounts have all the same name, Asset Inventory search results hard to read
Hi!
FAST SAs are all named "Terraform managed", as a result, CAI search results looks like this
https://imgur.com/a/Eq4rMtW

Note that CAI search displays these fields only:

Resource | Role | Principal | Resource type | Project Id

therefore not sure adding descriptions would cut it

bye!

juliocc:
Fixes #944

### DIFF
--- a/fast/stages/00-bootstrap/automation.tf
+++ b/fast/stages/00-bootstrap/automation.tf
@@ -111,11 +111,11 @@ module "automation-tf-bootstrap-gcs" {
 }
 
 module "automation-tf-bootstrap-sa" {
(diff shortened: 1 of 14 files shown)


ludoo (approved):
is this a breaking change? we should mark it as such in case

juliocc:
> is this a breaking change? we should mark it as such in case

Nothing breaks. It just updates the SAs in place.

PR #951, opened 2022-11-05
cloud-functions v2 - fix reference to bucket_name

wiktorn:
Use local variable for storage source as it is done for v1 functions.

When bucket_name is provided on input of the module then creation of v2 function fails with an error:
```
╷
│ Error: Invalid index
│ 
│   on .terraform/modules/[...]/modules/cloud-function/main.tf line 144, in resource "google_cloudfunctions2_function" "function":
│  144:         bucket = google_storage_bucket.bucket[0].name
│     ├────────────────
│     │ google_storage_bucket.bucket is empty tuple
│ 
```

These changes fixes this error.

Copied tests from v1 functions to ensure that v2 function is working properly.

### DIFF
--- a/modules/cloud-function/main.tf
+++ b/modules/cloud-function/main.tf
@@ -141,7 +141,7 @@ resource "google_cloudfunctions2_function" "function" {
     environment_variables = var.environment_variables
     source {
       storage_source {
-        bucket = google_storage_bucket.bucket[0].name
+        bucket = local.bucket
         object = google_storage_bucket_object.bundle.name
       }
     }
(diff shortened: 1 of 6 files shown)


juliocc (approved):

wiktorn:
One more thing I noticed with Cloud Functions v2 is that they do not support eventarc triggers.

Because of the difference in structure I think that the best approach is to create separate variable that contains necessary variables. If you are OK with this approach, I can work on that in separate PR.

ludoo:
> One more thing I noticed with Cloud Functions v2 is that they do not support eventarc triggers.
> 
> Because of the difference in structure I think that the best approach is to create separate variable that contains necessary variables. If you are OK with this approach, I can work on that in separate PR.

Sure. It would be great to also refactor the cloud function module to leverage Terraform 1.3 `optional`, like we've been doing for several other modules. Would you be up for it?

wiktorn:
@ludoo sure, I'll come back with another PR soon.

PR #960, opened 2022-11-08
Fix README typo in firewall module

valeriobponza:


modules/net-vpc-firewall/README.md
```diff
@@ -138,10 +138,10 @@ module "firewall" {
   project_id       = "my-project"
   network          = "my-network"
   factories_config = {
-
+    rules_folder    = "config/firewall"
+    cidr_tpl_fil    = "config/cidr_template.yaml"
```
averbuks, line 142: Thank you for spotting that and contribution. 
A small correction, suppose `e` is missing, should be `cidr_tpl_file`.

ludoo (approved):

PR #962, opened 2022-11-10
Add filtering-proxy-psc blueprint

kunzese:
This PR adds a new variation of the `filtering-proxy` blueprint in which the shared VPC is replaced by isolated VPCs and the proxy service is published and consumed via the Private Service Connect.

FYI @gaspar-chilingarov


ludoo:
Thanks for this submission!

Most of our blueprints try to minimize requirements by allowing reuse of a preexisting project, VPC, etc. as many users only have a project available for testing.

Do you think a folder is necessary? Can we make project creation optional like in most other blueprints, and allow reuse of an existing VPC for clients? There are several examples here on how to do that.

gaspar-chilingarov:
I would suggest that we move example of how the environment should be set up (folder + org.policy for external IPs) into separate example-environment.tf (or something similarly named), so that user can remove such file if they want to or move into some other part of the TF codebase.

Same with creating folter 'netops' - most probably customer already has it, so we can just leave it in the example and user wants they can replace local object/variable that points to the folder. 

About using  ` connection_preference = "ACCEPT_AUTOMATIC"` - we may want to warn users in the documentation that this is potential security problem (any project in GCP can connect to such provider) and they need to use `MANUAL` mode and provide list of the projects (most probably integrating this with the project factory).

ludoo:
> I would suggest that we move example of how the environment should be set up (folder + org.policy for external IPs) into separate example-environment.tf (or something similarly named), so that user can remove such file if they want to or move into some other part of the TF codebase.

I disagree: blueprints are limited examples and don't prescribe "how the environment should look like", which is a foundational design topic. In this case, the example shows a proxy via PSC, there's no good giudance that would involve folders or specific project layouts.

This repository has a very specific approach, and examples that are reflecting it will not be accepted.

ludoo:
One more thing: we also strongly discourage submodules and nested patterns. This is all detailed clearly in our contributing guidelines. What's the rationale for having a submodule here?

kunzese:
@ludoo done - everything is now in the same project, which can already exist or will be created if not
@gaspar-chilingarov - connection_preference = "ACCEPT_AUTOMATIC" has been changed to "ACCEPT_MANUAL"

ludoo:
Let me rephrase the comments above in a more productive way: what is the bare minimum design and resources that would allow showing this feature?

kunzese:
> One more thing: we also strongly discourage submodules and nested patterns. This is all detailed clearly in our contributing guidelines. What's the rationale for having a submodule here?

This has been taken from the `psc-hybrid` blueprint, but i can change that as well.

kunzese:
@ludoo submodule removed

gaspar-chilingarov:
We may need 2 projects to showcase the feature 
- first where the squid + PSC producer is installed and 
- second proejct - with the VPC with PSC endpoint is created and VM with proxy configuration is installed.

How should be go about it? Have 2 different directories with examples of proxy and example of consumer?

blueprints/networking/filtering-proxy-psc/main.tf
```diff
@@ -0,0 +1,300 @@
...
+resource "google_compute_service_attachment" "service_attachment" {
+  name                  = "psc"
+  project               = module.project.project_id
+  region                = var.region
+  enable_proxy_protocol = false
+  connection_preference = "ACCEPT_MANUAL"
+  nat_subnets           = [module.vpc.subnets_psc["${var.region}/psc"].self_link]
+  target_service        = module.squid-ilb.forwarding_rule_self_link
+  consumer_accept_lists {
+    project_id_or_num = module.project.project_id
```
gaspar-chilingarov, line 106: this should be project IDs of consumer projects, not the producer, isn't it? most probably coming from the variable?

blueprints/networking/filtering-proxy-psc/main.tf
```diff
@@ -0,0 +1,300 @@
...
+  }]
+  health_check_config = {
+    enable_logging = true
+    tcp = {
+      port = 3128
+    }
+  }
+}
+
+###############################################################################
```
gaspar-chilingarov, line 212: ohhh, consumer is in here. can we split it into separate .tf file?

kunzese: Done - moved all consumer resources to consumer.tf.

blueprints/networking/filtering-proxy-psc/main.tf
kunzese: After @ludoo remark, i moved everything to own project with multiple VPCs to showcase this.

ludoo (approved):
This is really nice, thanks! :)

A couple minor additions if you still have the energy

- variables and outputs should be sorted alphabetically, tests will fail if they aren't
- you might want to add links to the top-level/blueprint/section READMEs to make this discoverable

I am also adding you to the repo users, so you will be able to run tests, merge once all checks pass, and in the future use local branches for PRs.

blueprints/networking/filtering-proxy-psc/README.md
```diff
@@ -0,0 +1,28 @@
+# Network filtering with Squid with isolated VPCs using Private Service Connect
+
+This blueprint shows how to deploy a filtering HTTP proxy to restrict Internet access. Here we show one way to do this using isolated VPCs and Private Service Connect:
+
+- The `app` subnet hosts the consumer VMs that will have their Internet access tightly controlled by a non-caching filtering forward proxy.
```
gaspar-chilingarov, line 5: `Usually such consumer VMs will be deployed in the separate projects` ?

PR #963, opened 2022-11-10
Refactor vps-sc module for Terraform 1.3

ludoo:
This also refactors and streamlines the FAST security stage, as the two are tightly coupled.


modules/vpc-sc/variables.tf
```diff
@@ -66,94 +67,103 @@ variable "access_policy_create" {
...
+  }))
+  default  = {}
+  nullable = false
+  validation {
+    condition = alltrue([
+      for k, v in var.ingress_policies :
+      v.from.identity_type == null || contains([
+        "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
+        "ANY_USER", "ANY_SERVICE_ACCOUNT"
+      ], coalesce(v.from.identity_type, "-"))
```
juliocc, line 124: I think the coalesce here is not needed (the branch of the OR guarantees it always non-null)

tests/modules/vpc_sc/test_plan.py
```diff
@@ -12,9 +12,16 @@
...
   "Test with auto-created policy."
-  _, resources = plan_runner()
+  access_policy_create = '''{
+    parent = "organizations/123456"
+    title  = "vpcsc-policy"
+  }'''
+  _, resources = plan_runner(tf_var_file='test.regular.tfvars',
+                             access_policy='null',
+                             access_policy_create=access_policy_create)
   counts = {}
```
juliocc, line 25: `collections.Counter` (see [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/tests/modules/cloudsql_instance/test_plan.py#L113)). I know you'll say no but this is very python 2.2.

juliocc (approved):

modules/vpc-sc/variables.tf
ludoo: hah, I got an error on null type and had to ad the coalesce :)

tests/modules/vpc_sc/test_plan.py
ludoo: oh wow, I should go and check up on collections from time to time

ludoo: done, thanks

modules/vpc-sc/README.md
```diff
@@ -107,65 +99,61 @@ module "test" {
...
+    }
+  }
+  ingress_policies = {
+    # allow management from external automation SA
+    sa-tf-test = {
+      from = {
+        identities = [
+          "serviceAccount:test-tf@myproject.iam.gserviceaccount.com",
+        ],
+        source_access_levels = ["*"]
```
lcaggio, line 136: What about adding in the example an Access level for company proxy IP ranges and use it here to limit the ingress policy to company proxy ips?

The example would suggest a more secure solution (automation SA has high roles that is better secure).

lcaggio (approved):

modules/vpc-sc/README.md
ludoo: Absolutely. Do you mind adding it after we merge?

Issue #954, opened 2022-11-07
revisit prefix handling


juliocc:
Handling of the `prefix` variable is inconsistent across modules. For example `project` doesn't allow `""` but gcs does.

Agree on a single pattern and use it in all modules.

skalolazka:
To summarize the proper usage:
1. Conditional in locals to detect null and empty string
2. Default null in variables.tf
So the ideal way is, for example, in the iam-service-account module:
prefix = var.prefix == null || var.prefix == "" ? "" : "${var.prefix}-"
Is my understanding correct?

juliocc:
I would prefer validating that `prefix != ""` and just handling the case when `prefix == null`

The local for iam-service-account would end up like this:
```
prefix = var.prefix == null ? "" : "${var.prefix}-"
```

@ludoo wdyt?

skalolazka:
Opened https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/967 for the blueprints.

PR #964, opened 2022-11-10
prefix variable consistency across modules

skalolazka:
Modules/ folder ready - consistent usage of the "prefix" variable

Fixes https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/954


modules/cloud-function/variables.tf
```diff
@@ -94,6 +94,10 @@ variable "prefix" {
   description = "Optional prefix used for resource names."
   type        = string
   default     = null
+  validation {
+    condition     = var.prefix != ""
+    error_message = "Prefix can not be empty, please use null instead."
```
juliocc, line 99: nit: s/can not/cannot/

juliocc (commented):
There's also a few cases inside the blueprints folder. What do we want to do with those?

skalolazka:
> 

I'll try to see how long it might take me to go through the blueprints=) Being a perfectionist, I would ideally fix everywhere%)

The other question is, should we add the same in fast/ (adding to existing validation and things)? Example: 00-bootstrap. No default null, can be an empty string though. Right? Wouldn't that break some things? Or am I misreading it?

juliocc:
> > 
> 
> I'll try to see how long it might take me to go through the blueprints=) Being a perfectionist, I would ideally fix everywhere%)
> 
> The other question is, should we add the same in fast/ (adding to existing validation and things)? Example: 00-bootstrap. No default null, can be an empty string though. Right? Wouldn't that break some things? Or am I misreading it?

FAST does its own thing with prefixes. I'd focus only on blueprints for now.

ludoo (approved):

ludoo:
Thanks :)

PR #965, opened 2022-11-10
Add triggers to Cloud Functions v2

### LINKED ISSUE #973: Cloud Functions bindings for v2
The `cloud-function` module should use `google_cloudfunction2_function_iam_binding` when v2 functions is requested. Currently it uses `google_cloudfunctions_function_iam_binding` regardless the `v2` variable.

wiktorn:
* (incompatibile change) change structure for `trigger_config` and create substructures for v1 Cloud Functions and v2
* (incompatibile change) rename `bucket_config.lifecycle_delete_age` to `bucket_config,lifecycle_delete_age_days`
* (incompatibile change) rename `function_config.instances` to `function_config.instance_count`
* (incompatibile change) rename `function_config.memory` to `function_config.memory_mb`
* (incompatibile change) rename `function_config.timeout` to `function_config.timeout_seconds`
* add optional for objects in variables.tf
* make examples in README runable as tests
* add example for Cloud Function v2
* add example for trigger for Cloud Function v2
* remove optional variables from examples with `null` value
* assign IAM bindings to function based on version 
* bump default Python version to 3.10

Closes #973


modules/cloud-function/variables.tf
```diff
@@ -136,7 +136,24 @@ variable "trigger_config" {
   type = object({
     event    = string
     resource = string
-    retry    = bool
+    retry    = optional(bool)
+  })
+  default = null
+}
+
+variable "trigger_config_v2" {
```
ludoo, line 144: Would it make sense to have `trigger_config` be an object with `v1` and `v2` members? I'm not sure, just asking how you feel about it.

ludoo (approved):
A couple minor comments and a doubt which can be ignored, but good to go many thanks!

modules/cloud-function/variables.tf
```diff
@@ -50,13 +50,13 @@ variable "environment_variables" {
...
   type = object({
-    entry_point = string
-    instances   = number
-    memory      = number # Memory in MB
-    runtime     = string
-    timeout     = number
+    entry_point = optional(string, "main")
+    instances   = optional(number, 1)
+    memory      = optional(number, 256) # Memory in MB
+    runtime     = optional(string, "python37")
```
juliocc, line 58: Can we default to python310 which is the current [recommended runtime for python](https://cloud.google.com/functions/docs/concepts/python-runtime)?

modules/cloud-function/variables.tf
```diff
@@ -17,8 +17,8 @@
 variable "bucket_config" {
   description = "Enable and configure auto-created bucket. Set fields to null to use defaults."
   type = object({
-    location             = string
-    lifecycle_delete_age = number
+    location             = optional(string)
```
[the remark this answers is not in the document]
wiktorn: Currently we have following structure when using location:
```hcl
  location = (
    var.bucket_config.location == null
    ? var.region
    : var.bucket_config.location
  )
```
And its not allowed to provide other variable as default value. I prefer the current approach and make it default to the region of function.

modules/cloud-function/variables.tf
wiktorn: I did not that originally do that to preserve backward compatibility. But having it within one variable allows setting condition that checks that only one was provided, which I consider an advantage.

Pushed changes containing this approach. The only thing I don't like about it is the need to use `try(...)` to check existence of fields and this prevents detecting typos in these constructs.

modules/cloud-function/variables.tf
wiktorn: Happy to update that in 2f71b68db327ac26dc3248ab705b537f2b2f4bb1

wiktorn:
@juliocc @ludoo the only feature I also consider for this, is to also support creation of trigger service accounts. As you can notice in [example](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/965/files#diff-7189a2f5f0eee4cd18df46c761f98c1b28b94ddcc90c2e1977336aa787be5d4cR74-R83) - caller needs to specify service account for trigger by themselves.

I can add `create_service_account = optional(bool, true)` in v2 trigger_config and grant `roles/run.invoker` by default. This will help `network-dashboard` blueprint (that's my guess, why it's not working with V2 CF and requires `ALLOW_ALL`).

The other thing that I guess we could improve here is to provide option to specify roles for Cloud Function service account. Though this kind of overlaps with `iam` attribute. Any thoughts on that?

ludoo:
> 

That looks like a good idea, up to you if you want to add it here or in a separate PR.

Julio was also thinking about splitting the module into a v1 and a v2 module, which kind of makes sense. Again, feel free to merge this and we can discuss it later, or split it now if it makes more sense to you.

BTW, you should see a merge button already, if that's not the case let us know.

juliocc:
> Julio was also thinking about splitting the module into a v1 and a v2 module, which kind of makes sense. Again, feel free to merge this and we can discuss it later, or split it now if it makes more sense to you.

Lets's discuss it. I've been thinking about it for a few days but tbh I don't know if it's a good idea.

modules/cloud-function/variables.tf
```diff
@@ -17,8 +17,8 @@
...
+    lifecycle_delete_age = optional(number)
   })
   default = null
```
juliocc, line 23: I think this variable should be non-nullable. Your location check depends on the object existing and having the two fields defined.
```hcl
  location = (
    var.bucket_config.location == null
    ? var.region
    : var.bucket_config.location
  )
```

modules/cloud-function/variables.tf
```diff
@@ -138,11 +138,29 @@ variable "service_account_create" {
...
+        value     = string
+        operator  = string
+      })))
+      service_account_email = optional(string)
+      retry_policy          = optional(string)
+    }))
   })
   default = null
+  validation {
+    condition     = try(((var.trigger_config.v1 == null) != (var.trigger_config.v2 == null)), var.trigger_config == null)
```
juliocc, line 161: I think this is more readable
```var.trigger_config == null || (var.trigger_config.v1 == null) != (var.trigger_config.v2 == null)```

File: modules/cloud-function/main.tf [hunk 6]
```diff
@@ -84,14 +84,14 @@ resource "google_cloudfunctions_function" "function" {
   )
 
   dynamic "event_trigger" {
-    for_each = var.trigger_config == null ? [] : [""]
+    for_each = try(var.trigger_config.v1 == null, true) ? [] : [""]
```
juliocc, line 87: We usually do this as
`try(var.trigger_config.v1, null) != null ? [""] : []`. It's the same, but this one is more consistent with the rest of the repo.

Or even better, would it make sense to make `trigger_config` non-nullable?

modules/cloud-function/variables.tf
```diff
@@ -50,13 +50,13 @@ variable "environment_variables" {
...
+  description = "Cloud function configuration. Defaults to using main as entrypoint, 1 instance with 256MiB of memory, and 180 second timeout"
   type = object({
-    entry_point = string
-    instances   = number
-    memory      = number # Memory in MB
-    runtime     = string
-    timeout     = number
+    entry_point = optional(string, "main")
+    instances   = optional(number, 1)
+    memory      = optional(number, 256) # Memory in MB
```
juliocc, line 57: should we add the units to so `memory` and `timeout` become `memory_mb` `timeout_seconds`? (idea from [here](https://cloud.google.com/apis/design/naming_convention#quantities))

@ludoo wdyt?

modules/cloud-function/variables.tf
wiktorn: Regrettably this doesn't work:
```
Error: Attempt to get attribute from null value

  on /home/wiktorn/git/cloud-foundation-fabric/modules/cloud-function/variables.tf line 162, in variable "trigger_config":
 162:     condition     = var.trigger_config == null || (var.trigger_config.v1 == null) != (var.trigger_config.v2 == null)
    ├────────────────
    │ var.trigger_config is null
```

modules/cloud-function/variables.tf
wiktorn: Existence of `bucket_config` is guaranteed few lines above with:
```
count                       = var.bucket_config == null ? 0 : 1
```

I like the current interface of the module, that if you don't provide `location` it will default `var.region`

modules/cloud-function/main.tf
```diff
@@ -255,3 +289,21 @@ resource "google_service_account" "service_account" {
   account_id   = "tf-cf-${var.name}"
   display_name = "Terraform Cloud Function ${var.name}."
 }
+
+resource "google_service_account" "trigger_service_account" {
+  count = try(var.trigger_config.v2.service_account_create, null) == null ? 0 : (
```
juliocc, line 294: `try(var.trigger_config.v2.service_account_create, false) == true ? 1 : 0`

modules/cloud-function/main.tf
```diff
@@ -192,14 +221,23 @@ resource "google_cloudfunctions2_function" "function" {
 }
 
 resource "google_cloudfunctions_function_iam_binding" "default" {
-  for_each       = var.iam
+  for_each       = var.v2 == false ? var.iam : {}
```
juliocc, line 224: `!var.v2 ? var.iam : {}`

modules/cloud-function/main.tf
```diff
@@ -192,14 +221,23 @@ resource "google_cloudfunctions2_function" "function" {
...
   project        = var.project_id
   region         = var.region
   cloud_function = local.function.name
   role           = each.key
   members        = each.value
 }
 
+resource "google_cloudfunctions2_function_iam_binding" "default" {
+  for_each       = var.v2 == true ? var.iam : {}
```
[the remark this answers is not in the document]
wiktorn: Fixed, thanks

modules/cloud-function/main.tf
wiktorn: Fixed, thanks

modules/cloud-function/main.tf
wiktorn: I've made `trigger_config` default to `{v1 = null, v2 = null}` and simplified some these conditions. Thanks.

modules/cloud-function/variables.tf
wiktorn: Added this in 2ce55e20ef87c1ffeed766a39824898b10f25e8f.

modules/cloud-function/main.tf
```diff
@@ -39,6 +39,16 @@ locals {
     )
     : var.service_account
   )
+  trigger_service_account_email = (
```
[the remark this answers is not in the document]
wiktorn: The idea is that is either null or an SA email. Let's dissect the expression:
```hcl
    try(var.trigger_config.v2.service_account_create, null) == null
    ? false
    : var.trigger_config.v2.service_account_create ? (
      length(google_service_account.trigger_service_account) > 0
      ? google_service_account.trigger_service_account[0].email
      : null
    )
    : try(var.trigger_config.v2.service_account_email, null)
```

So first part:
```hcl
    try(var.trigger_config.v2.service_account_create, null) == null
    ? false
    : var.trigger_config.v2.service_account_create 
```
gives either false or value of `var.trigger_config.v2.service_account_create` when it is not null.

Then the result of above expression is used to choose whether:
````hcl
above_expression ? (
      length(google_service_account.trigger_service_account) > 0
      ? google_service_account.trigger_service_account[0].email
      : null
    )
    : try(var.trigger_config.v2.service_account_email, null)
```
Return email of service account that was created by resource or email that was provided in `var.trigger_config.v2.service_account_email`.

Hopefully I wrote what I meant. When do you think it returns `false`?

Regrettably adding extra brackets doesn't help with tf formating:
```hcl
  trigger_service_account_email = (
    (
      try(var.trigger_config.v2.service_account_create, null) == null
      ? false
      : var.trigger_config.v2.service_account_create
      ) ? (length(google_service_account.trigger_service_account) > 0
      ? google_service_account.trigger_service_account[0].email
      : null
    )
    : try(var.trigger_config.v2.service_account_email, null)
  )
``` 
I'd love the line `) ? (length(google_service_account.trigger_service_account) > 0` to be one-step less indented than  the surrounding ones.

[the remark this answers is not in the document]
juliocc: > When do you think it returns false?

The very first branch returns a false:
```hcl
  trigger_service_account_email = (
    (
      try(var.trigger_config.v2.service_account_create, null) == null
      ? false # <--- here
```


I would format it like this:
```hcl
trigger_service_account_email = (
  try(var.trigger_config.v2.service_account_create, null) == null
  ? false
  : (
    var.trigger_config.v2.service_account_create
    ? (
      length(google_service_account.trigger_service_account) > 0
      ? google_service_account.trigger_service_account[0].email
      : null
    )
    : try(var.trigger_config.v2.service_account_email, null)
  )
)
```
But that's besides the point. You have 3 nested conditions that are really hard to follow.

I'm going to throw a few (untested) ideas for you to explore:

```hcl
locals{
  # use try for the inner conditional
  trigger_service_account_email1 = (
    try(var.trigger_config.v2.service_account_create, null) == null
    ? false
    : (
      var.trigger_config.v2.service_account_create
      ? try(google_service_account.trigger_service_account[0].email, null)
      : try(var.trigger_config.v2.service_account_email, null)
    )
  )

  # use try to check create flag
  trigger_service_account_email2 = (
    try(var.trigger_config.v2.service_account_create, false)
    ? try(google_service_account.trigger_service_account[0].email, null)
    : try(var.trigger_config.v2.service_account_email, null)
  )

  # just use try
  trigger_service_account_email3 = try(
    google_service_account.trigger_service_account[0].email,
    var.trigger_config.v2.service_account_email,
    null
  )
}
```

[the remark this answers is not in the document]
wiktorn: Just for the record, the whole expression:
```hcl
    try(var.trigger_config.v2.service_account_create, null) == null
    ? false
    : var.trigger_config.v2.service_account_create 
```
Returns `true` or `false`. This is later used in following expression. I agree that this is far to complex to follow. I came up with this simplification:
```hcl
  trigger_service_account_email = (
    coalesce(try(var.trigger_config.v2.service_account_create, false), false)
    ? google_service_account.trigger_service_account[0].email
    : null
  )
```

The part with checking whether the service account was created or not I took from what is already there:
```hcl
  service_account_email = (
    var.service_account_create
    ? (
      length(google_service_account.service_account) > 0
      ? google_service_account.service_account[0].email
      : null
    )
    : var.service_account
  )
```
It's kind of really defensive programming as `var.service_account_create` is used to control, whether service account is created at all. I guess this could too can be simplified to
``hcl
  service_account_email = (
    var.service_account_create
    ? google_service_account.service_account[0].email
    : var.service_account
  )
```
or please help me understand what I'm missing there .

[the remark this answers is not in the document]
juliocc: My whole point is that the original expression is unnecessarily complex. If you can simplify it without compromising the logic, you should.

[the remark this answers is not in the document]
wiktorn: Done for `trigger_service_account` in 0cb1f6ae3c822bfe7dd2d37d30861178eac1ef61 and for `service_account_email` in 78d20b1f16307cd3c5d5bbdd7c8f9bfd6bbf7f7d

PR #966, opened 2022-11-11
FAST: improve GitHub workflow, stage 01 output fixes

ludoo:
This PR

- fixes errors in stage 01 outputs when `output_files` is set to null after having applied
- improves the GitHub workflow to make it more resilient

Our GitHub workflow posts a comment to the PR with the result of the validation and plan steps, and for large plans it's fairly easy for the comment text to go over the maximum allowed limit of characters. This set of changes works around the issue by

- posting the full plan output to the action status (visible in the action details) and preserving it in the action log
- posting a smaller plan output to the comment where lines for the refresh part of the plan have been omitted

The action status is also posted for the apply phase that runs after merging, which does not post a comment to the PR.

Action status

image

Trimmed down plan in PR comment

image


fast/stages/01-resman/outputs-files.tf
```diff
@@ -16,23 +16,27 @@
...
+  outputs_location = try(pathexpand(var.outputs_location), "")
+}
+
 resource "local_file" "providers" {
   for_each        = var.outputs_location == null ? {} : local.providers
   file_permission = "0644"
-  filename        = "${pathexpand(var.outputs_location)}/providers/${each.key}-providers.tf"
-  content         = each.value
+  filename        = "${local.outputs_location}/providers/${each.key}-providers.tf"
+  content         = try(each.value, null)
```
juliocc, line 27: You have a `try()` here but not on the other `local_file`s. Is that WAI?

juliocc (approved):

fast/stages/01-resman/outputs-files.tf
ludoo: that is the only one that was failing...

PR #968, opened 2022-11-11
Enforce PROXY protocol in `filtering-proxy-psc` blueprint

kunzese:
This PR enables the PROXY protocol in the `filtering-proxy-psc` blueprint. Consumer workload IPs are now visible in the Squid access logs via `docker exec squid cat /var/log/squid/access.log`.

FYI @gaspar-chilingarov @ludoo


gaspar-chilingarov:
Can we change the readme to mention that it collects logs + how to view them?

ludoo (approved):

ludoo:
> but without any success. So either we start poking around some privilege escalation or it is what it is at the moment. What do you think @gaspar-chilingarov @ludoo - any other ideas?

One option would be to write logs to a mount, and run a second container to tail them.

BTW I seem to remember `gcplogs` is suboptimal and it's better to run the logging agent by enabling it via metadata on COS. Not important, but if you have a spare cycle maybe it's better to switch (and we'll also have to check other COS modules/blueprints too). Ah, and I did not check your source, but can you make sure the docker logs are set to a maximum size via docker daemon opts or filesystem will be filled (sorry if you did it already!).

kunzese:
Then let's do it like this: i merge this PR, because this enables the PROXY protocol @gaspar-chilingarov asked about, then i will have look at the `./modules/cloud-config-container/squid` part (and potentially others) to replace the `gcplogs` driver with the Cloud Logging in the COS image, and then i will come back here for the access log visibility. If i don't get any vetos by tomorrow morning, i will start working on this.

ludoo:
> Then let's do it like this: i merge this PR, because this enables the PROXY protocol @gaspar-chilingarov asked about, then i will have look at the `./modules/cloud-config-container/squid` part (and potentially others) to replace the `gcplogs` driver with the Cloud Logging in the COS image, and then i will come back here for the access log visibility. If i don't get any vetos by tomorrow morning, i will start working on this.

great plan! as always, you should be able to merge once checks and review are in as in this case :)

PR #970, opened 2022-11-11
Update logging sinks to tf1.3 in resman modules

juliocc:


fast/stages/00-bootstrap/organization.tf
```diff
@@ -192,13 +192,9 @@ module "organization" {
...
-      bq_partitioned_table = attrs.type == "bigquery"
-      destination          = local.log_sink_destinations[name].id
-      exclusions           = {}
-      filter               = attrs.filter
-      iam                  = true
-      include_children     = true
-      type                 = attrs.type
+      destination                    = local.log_sink_destinations[name].as_logging_destination
+      filter                         = attrs.filter
+      bigquery_use_partitioned_table = attrs.type == "bigquery"
```
ludoo, line 197: can we rename this to `bq_partitioned_table`?

modules/bigquery-dataset/outputs.tf
```diff
@@ -14,6 +14,21 @@
  * limitations under the License.
  */
 
+output "as_logging_destination" {
```
ludoo, line 17: I still am not 100% convinced about this output, I would personally prefer only having `id` and typing `destination.type` by hand, without creating a dependency between two modules. I also don't like that inputs for logginq become opaque, you read `destination = module.foo.as_destination` and to have a clue on what that means, you have to open the other module and go look at its outputs. Feels like a slippery slope if we start doing this.

ludoo (approved):

modules/bigquery-dataset/outputs.tf
ludoo: Can we make `destination.type` optional, and infer from `destination.id` if `null`? For example, a simple regexp can be used on the id to detect a dataset, etc. It would negate the need to have a specialized output.

fast/stages/00-bootstrap/organization.tf
juliocc: done

modules/bigquery-dataset/outputs.tf
juliocc: > Can we make destination.type optional, and infer from destination.id if null?

The reason we always had two fields is because the underlying sink resource wants the destination in this format:
- storage.googleapis.com/[GCS_BUCKET]
- bigquery.googleapis.com/projects/[PROJECT_ID]/datasets/[DATASET]
- pubsub.googleapis.com/projects/[PROJECT_ID]/topics/[TOPIC_ID]
- logging.googleapis.com/projects/[PROJEC_ID]/locations/[BUCKET_LOCATION]/buckets/[BUCKET_NAME]

The `type` field is basically to help us build these 4 strings based on the destination because the `id` alone doesn't give us enough information.

> I still am not 100% convinced about this output, I would personally prefer only having id and typing destination.type by hand, without creating a dependency between two modules. I also don't like that inputs for logginq become opaque, you read destination = module.foo.as_destination and to have a clue on what that means, you have to open the other module and go look at its outputs. Feels like a slippery slope if we start doing this.

We still have the two fields (`type` and `target`), they're just inside the `destination` object. If you want, you could provide them manually which is no different to what we had before. IMO they belong as an object since the type is an attribute of the destination, not the sink itself. 

As for the `as_logging_destination` outputs. I agree that it now makes things more opaque, but on the other hand the `type` field was always in implementation detail forced by the underlying resource. My whole idea with this specialized output is that in general  the user doesn't care about the `type` field, they just want to send the logs to a GCS bucket, or a to BiqQuery dataset, etc. Forcing the user to specify the field was leaking our own abstraction.

With this change, the 90% use case looks something like this:
```hcl
module "gcs" {
  source        = "./fabric/modules/gcs"
  //...
}
module "folder-sink" {
  source = "./fabric/modules/folder"
  //...
  logging_sinks = {
    warnings = {
      destination = module.gcs.as_logging_destination
      filter      = "severity=WARNING"
    }
  }
}
```

Without that output you would have to do something like this:
```hcl
module "folder-sink" {
  source = "./fabric/modules/folder"
  //...
  logging_sinks = {
    warnings = {
      target = module.gcs.as_logging_destination
      type   = "storage"
      filter = "severity=WARNING"
    }
  }
}
```

It's a small convenience for the case when you have the destination created close by. I agree it's a slippery slope and I was kind of hesitant about it too because of the dependency between the two modules. 

We're just saving one line, so if you feel strongly about it, we can got back to having the user specify the type manually. But I kind of [like this](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/8fe19ad7c2232957d84d22ba00d0338b79428d3d/fast/stages/00-bootstrap/organization.tf#L193) 🙂

ludoo: > > Can we make destination.type optional, and infer from destination.id if null?
> 
> The reason we always had two fields is because the underlying sink resource wants the destination in this format:

I know, but having the source id already gives us the type, so I'm not saying to remove type (the var type is great), just make type optional and infer from id if type is null.

> The `type` field is basically to help us build these 4 strings based on the destination because the `id` alone doesn't give us enough information.

- id for bigquery datasets is of the form `projects/{{project}}/datasets/{{dataset_id}}`
- id for log buckets is of the form `projects/{{project}}/locations/{{location}}/buckets/{{bucket_id}}`
- id for pubsub topics os of the form `projects/{{project}}/topics/{{name}}`
- storage buckets are the only destination id without a path, so that's fine too

> We still have the two fields (`type` and `target`), they're just inside the `destination` object. If you want, you could provide them manually which is no different to what we had before. IMO they belong as an object since the type is an attribute of the destination, not the sink itself.

I'm just saying to not use the specialized logging output, and just pass in the id which already contains the type (see above):

```hcl
module "gcs" {
  source        = "./fabric/modules/gcs"
  //...
}
module "folder-sink" {
  source = "./fabric/modules/folder"
  //...
  logging_sinks = {
    warnings = {
      destination = { id = module.gcs.id }
      filter      = "severity=WARNING"
    }
  }
}
```

juliocc: Nope, tried using a regex and it breaks with dynamic values. Left it as it was originally with the type field

ludoo: ah, because then you do the count/foreach based on a dynamic value via the regexp. Makes sense. I still don't like the purpose-built outputs and feel it's a slippery slope for worse things later. But go ahead as you like them a lot apparently. :)

juliocc: removed that too. Has the added benefit that this PR doesn't break existing users.

PR #977, opened 2022-11-15
Replace Docker's `gcplogs` driver with the GCP COS logging agent

kunzese:
This PR removes Docker's `gcplogs` driver from all modules and blueprints and replaces it with the [logging agent included with the COS image](https://cloud.google.com/container-optimized-os/docs/how-to/logging).


ludoo (changes requested):
Very very nice. I just would not want to start having specialized stuff int he compute VM module, can you please revert those changes and implement the metadata injection at the COS module level?

kunzese:
> Very very nice. I just would not want to start having specialized stuff int he compute VM module, can you please revert those changes and implement the metadata injection at the COS module level?

Sure, but as far is i see it i can not put in the `cos-generic-metadata` module, because this only assembles a cloud config which is later injected into the `google_compute_instance.metadata.user-data` and the `google-logging-enabled = true` property must be set on VM instance level (or project level). But we could remove the new variable and let the users add the metadata property themselves, like:

From:
```
module "vm-squid" {
  source                = "./fabric/modules/compute-vm"
  enable_google_logging = true
  metadata = {
    user-data = module.cos-squid.cloud_config
  }
}
```

To:
```
module "vm-squid" {
  source = "./fabric/modules/compute-vm"
  metadata = {
    user-data              = module.cos-squid.cloud_config
    google-logging-enabled = true
  }
}
```

If you think of anything else, just let me know.

ludoo:
> If you think of anything else, just let me know.

Heh, I should learn to connect my brain before replying. :) Yes, the cos modules only manage `user-data` value. I would do what you suggest, adding one additional metadata to the examples, and the blueprints who use the COS modules. Thanks for your patience. :)

ludoo (approved):

PR #979, opened 2022-11-15
Add network tags support to the organization module

LucaPrete:


juliocc:
`networks_tags`, pretty please? `network_tags` might cause confusion, but `tags_network` is worse IMO.

And now that we're on it, the existing tags could be renamed to `resource_tags`?

LucaPrete:
Don't have an opinion on it @juliocc. I've just used tags_network to keep them closed in alphabetical order, but I'm fine either way.
Let me rename tags_network to network_tags in this patch, and then create another one to rename resource_tags, so we keep the scope of the PR small.

ludoo (approved):
Luca, can you just make sure variables and outputs are in alphabetical order?

LucaPrete:
Done.

PR #980, opened 2022-11-15
Have Squid log to /dev/stdout to stream logs to Cloud Logging

kunzese:
This PR uses dark Docker voodoo to make the Squid container able to log to /dev/stdout, which then get streamed to Cloud Logging via the COS logging agent.

FYI: @gaspar-chilingarov


ludoo (approved):
almost calls for docker compose :)

ludoo:
> > almost calls for docker compose :)
> 
> Good idea - i will adjust this PR.

you probably need to download docker-compose, which will then fail if there's no public access. :) interested in what you come up with, but this PR is totally fine as is if you want to merge!

gaspar-chilingarov:
i have a strange feeling about starting container with **USER=${squid_user}** and then using `sudo` to get more privileges. Why not to run as a root then?

gaspar-chilingarov:
LGTM :)

PR #985, opened 2022-11-16
Network dashboard - VPC-SC compliance features

maunope:
Private Worker pool, private VPC access, split metrics/CF deployment.


blueprints/cloud-operations/network-dashboard/variables.tf
```diff
@@ -82,3 +93,8 @@ variable "schedule_cron" {
   description = "Cron format schedule to run the Cloud Function. Default is every 10 minutes."
   default     = "*/10 * * * *"
 }
+
+variable "vpc_connector_name" {
+  description = "Serverless VPC connection name for the Cloud Function"
+  default     = ""
```
ludoo, line 99: I seem to remember we already had a variable for this; if that's not the case, could you check cloud run module and see if we can offer the same interface?

BTW default for empty strings is `null` not `""`

maunope: done

ludoo:
Closing this as we refactored the networking dashboard. Mauri, let's redo this together so you don't waste too much time getting familiar with the new code.

PR #994, opened 2022-11-20
Add schemas to Pubsub topic module.

iht:
Pubsub topics can now have schemas (https://cloud.google.com/pubsub/docs/admin#schemas). 

This PR adds an option to set the schema settings and create a new optional resource of type `google_pubsub_schema` attached to the `google_pubsub_topic`.

### DIFF
--- a/modules/pubsub/main.tf
+++ b/modules/pubsub/main.tf
@@ -37,7 +37,7 @@ locals {
 
 resource "google_pubsub_schema" "default" {
   count      = var.schema == null ? 0 : 1
-  name       = format("%s-%s", var.name, "schema")
+  name       = "{$var.name}-schema"
   type       = var.schema.schema_type
   definition = var.schema.definition
   project    = var.project_id
@@ -57,7 +57,6 @@ resource "google_pubsub_topic" "default" {
     }
   }
 
-  depends_on = [google_pubsub_schema.default]
   dynamic "schema_settings" {
     for_each = var.schema == null ? [] : [""]
     content {
--- a/modules/pubsub/main.tf
+++ b/modules/pubsub/main.tf
@@ -35,6 +35,14 @@ locals {
   }
 }
 
+resource "google_pubsub_schema" "default" {
+  count      = var.schema == null ? 0 : 1
+  name       = format("%s-%s", var.name, "schema")
+  type       = var.schema.schema_type
+  definition = var.schema.definition
+  project    = var.project_id
+}
+
 resource "google_pubsub_topic" "default" {
   project                    = var.project_id
   name                       = var.name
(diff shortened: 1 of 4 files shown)


modules/pubsub/outputs.tf
```diff
@@ -47,3 +47,19 @@ output "topic" {
     google_pubsub_topic_iam_binding.default
   ]
 }
+
+output "schema_id" {
+  description = "Schema resource id."
+  value       = google_pubsub_schema.default[0].id
+  depends_on = [
+    google_pubsub_schema.default
+  ]
```
juliocc, line 56: - you need to wrap the value in a `try()` otherwise this will fail when `var.schema` is null. Same for the other output
- the `depends_on` is not needed
- can you sort the outputs and variables alphabetically?

modules/pubsub/main.tf
```diff
@@ -48,6 +56,15 @@ resource "google_pubsub_topic" "default" {
       allowed_persistence_regions = var.regions
     }
   }
+
+  depends_on = [google_pubsub_schema.default]
```
juliocc, line 60: this `depends_on` is not needed

modules/pubsub/variables.tf
```diff
@@ -114,3 +114,13 @@ variable "subscriptions" {
   }))
   default = {}
 }
+
+variable "schema" {
+  description = "Topic schema. If set, all messages in this topic should follow this schema."
+  type = object({
+    schema_type  = string
+    definition   = string
+    msg_encoding = optional(string, "ENCODING_UNSPECIFIED")
```
juliocc, line 123: can you sort the fields alphabetically?

modules/pubsub/main.tf
```diff
@@ -35,6 +35,14 @@ locals {
   }
 }
 
+resource "google_pubsub_schema" "default" {
+  count      = var.schema == null ? 0 : 1
+  name       = format("%s-%s", var.name, "schema")
```
juliocc, line 40: we tend to prefer interpolation over `format()`. Can you change to `"${var.name}-schema"`?

juliocc (changes requested):
welcome @iht :)

modules/pubsub/variables.tf
iht: Done.

modules/pubsub/main.tf
iht: Done

modules/pubsub/main.tf
iht: Removed

modules/pubsub/outputs.tf
iht: Done, thanks for the tips! I was actually wondering how to make the output "optional".

juliocc (approved):

ludoo:
nice commit message :) 
image

Israel, I'm adding you to the repo users, you should be able to merge yourself after you accept the invite (I don't think we added you before)

PR #995, opened 2022-11-21
Push container images to GitHub instead of Google Container Registry

kunzese:
This PR adds workflows that push the `Squid`, `strongSwan` and `Toolbox` container images to GitHub. This solution consists of a generic Build & Push reusable workflow which is triggered by a separate workflow for each image. In my opinion this is the best, GitHub-nativ approach which does not depend on some `git diff`/`bash` magic - but other opinions are highly welcomed.


ludoo (approved):
Thanks for this! I'm ignorant so I'll ask: do we need to update container image links in the examples using these?

PR #999, opened 2022-11-22
Default nodepool creation fix

astianseb:
In case an organization has a Shielded VM policy enforced (constraints/compute.requireShieldedVm) module is failing to create default nodepool. Current fix is leveraging "shielded_nodes" variable value to set shielded_instance_config values in node_config for default node pool

### DIFF
--- a/modules/gke-cluster/main.tf
+++ b/modules/gke-cluster/main.tf
@@ -48,7 +48,12 @@ resource "google_container_cluster" "cluster" {
   enable_autopilot = var.enable_features.autopilot ? true : null
 
   # the default nodepool is deleted here, use the gke-nodepool module instead
-  # node_config {}
+  node_config {
+    shielded_instance_config {
+      enable_secure_boot = var.enable_features.shielded_nodes ? true : null
+      enable_integrity_monitoring = var.enable_features.shielded_nodes ? true : null
+    }
+  }
 
   addons_config {
     dynamic "dns_cache_config" {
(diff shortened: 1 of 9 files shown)


modules/gke-cluster/variables.tf
```diff
@@ -96,6 +96,13 @@ variable "enable_features" {
     }))
     vertical_pod_autoscaling = optional(bool, false)
     workload_identity        = optional(bool, false)
+    # node_config  = optional(object({
```
ludoo, line 99: do we need this?

modules/gke-cluster/main.tf
```diff
@@ -48,7 +48,12 @@ resource "google_container_cluster" "cluster" {
   enable_autopilot = var.enable_features.autopilot ? true : null
 
   # the default nodepool is deleted here, use the gke-nodepool module instead
-  # node_config {}
+  node_config {
+    shielded_instance_config {
```
ludoo, line 52: sorry for doing this in pieces, a bit of a hectic day, we might want to use a dynamic block definition here and avoid the block if both flags are false, something like

```hcl
dynamic node_config {
  for_each = var.enable_features.shielded_nodes == true  [""] : []
  content {
      enable_secure_boot = true
      enable_integrity_monitoring = true
  }
}
```

ludoo (changes requested):

ludoo (approved):

PR #1003, opened 2022-11-22
Normalize prefix handling in blueprints

### LINKED ISSUE #967: Prefix handling in blueprints
Similar to https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/954 , but for the blueprints folder. Mostly a perfectionistic ask, involves quite some boring work.

kunzese:
This PR normalizes the prefix handling in blueprints, similar to #964.

Fixes #967

### DIFF
--- a/blueprints/factories/project-factory/README.md
+++ b/blueprints/factories/project-factory/README.md
@@ -69,6 +69,7 @@ module "projects" {
   kms_service_agents     = try(each.value.kms, {})
   labels                 = try(each.value.labels, {})
   org_policies           = try(each.value.org_policies, {})
+  prefix                 = each.value.prefix
   service_accounts       = try(each.value.service_accounts, {})
   services               = try(each.value.services, [])
   service_identities_iam = try(each.value.service_identities_iam, {})
@@ -109,9 +110,9 @@ vpc_host_project: project-example-host-project
 
 # [opt] Billing account id - overrides default if set
 billing_account_id: 012345-67890A-BCDEF0
-                    
+
 # [opt] Billing alerts config - overrides default if set
-billing_alert:      
+billing_alert:
   amount: 10
   thresholds:
     current:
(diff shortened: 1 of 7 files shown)


kunzese:
@skalolazka FYI

juliocc:
Thanks @kunzese!

I'm not sure if we want the `default=null` for blueprints. @ludoo wdyt?

ludoo:
> Thanks @kunzese!
> 
> I'm not sure if we want the `default=null` for blueprints. @ludoo wdyt?

true, blueprints could require the prefix to be passed in, so as not to default to existing project/bucket names

kunzese:
> OMG, Sebastian, this is awesome! Thanks for taking your time! Nit: we did change "can not" to "cannot" in modules/ %)

But not in the `./CONTRIBUTING.md` where i copied it from :p but i adjusted it everywhere.

kunzese:
@ludoo @juliocc what if we set a reasonable default value for prefix?

```
variable "prefix" {
  description = "Optional prefix used for resource names."
  type        = string
  default     = "blueprint" // demo or fast or example or whatever
  validation {
    condition     = var.prefix != ""
    error_message = "Prefix cannot be empty, please use null instead."
  }
}
```

This way everything has a prefix per default, the user does not have to supply one when they don't need, but we keep the possible to just not use any prefix if someone decides to do so.

ludoo:
> @ludoo @juliocc what if we set a reasonable default value for prefix?
> 
> ```
> variable "prefix" {
>   description = "Optional prefix used for resource names."
>   type        = string
>   default     = "blueprint" // demo or fast or example or whatever
>   validation {
>     condition     = var.prefix != ""
>     error_message = "Prefix cannot be empty, please use null instead."
>   }
> }
> ```
> 
> This way everything has a prefix per default, the user does not have to supply one when they don't need, but we keep the possible to just not use any prefix if someone decides to do so.

The problem for blueprints (where there is no org-wide naming convention with a unique token, usually at the beginning of the name) is that setting a default prefix will originate clashes for each user after the first one that runs apply on project ids and GCS bucker names, which are unique across GCP.

For blueprints we should really not set a default for prefix, and ask users to set it in a tfvars file or pass it in via environment or arg.

kunzese:
Reworked the prefix handling, so that it is now mandatory.

ludoo (approved):

PR #1006, opened 2022-11-23
Add settings for autoscaling to Bigtable module.

iht:
Bigtable clusters can now autoscale, adding the corresponding settings to the `bigtable-instance`.

The default `num_workers` must be set to `null`, because setting `num_workers` cannot be done at the same time as specifying `autoscaling_config`.


lcaggio:
At the moment `autoscaling_config` and `num_nodes` have a null default. 

If not wrong: if the user relies on module defaults, an instance with unset num_nodes will be created (suggested for DEV). I would improve the num_nodes description suggesting to set num_nodes > 0 or set the auto_scaling variable for a prod env.

iht:
> At the moment `autoscaling_config` and `num_nodes` have a null default.
> 
> If not wrong: if the user relies on module defaults, an instance with unset num_nodes will be created (suggested for DEV). I would improve the num_nodes description suggesting to set num_nodes > 0 or set the auto_scaling variable for a prod env.

If not `num_nodes` nor `autoscaling_config` are set, the creation of the instance fails because the number of nodes is understood to be zero (default value is null). So the user is safe, no instance would be created in that situation. 

I have added some more remarks about this situation in the documentation.

See commits  e05dec98e07440601abdc8587e3e4c32be63dbb7 and 59e1d1313a58cd37b1afc2ff5efddeab89fd6cae.

lcaggio:
I see. In this case, the user has to set an optional variable otherwise the module won't work.

What about having a logic to not set `num_nodes` on the resource if the value is null? This should create a resource suitable for DEV (see [resource documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigtable_instance#num_nodes))

lcaggio:
num_nodes = null requires a `instance_type` configuration. `instance_type` is deprecated.

I suggest do one of the following:
 - set a default to 1 for `num_nodes` and ignore if `autoscaling_config` is configured (mentioning the behaviour in the variable description).
 - create a single variable with 2 attribute: num_nodes and autoscaling_config and use condition to verify that only one of the 2 is configured. 

In both ways the user can have a simple deployment just settings mandatory attributes.

iht:
> num_nodes = null requires a `instance_type` configuration. `instance_type` is deprecated.
> 
> I suggest do one of the following:
> 
> * set a default to 1 for `num_nodes` and ignore if `autoscaling_config` is configured (mentioning the behaviour in the variable description).
> * create a single variable with 2 attribute: num_nodes and autoscaling_config and use condition to verify that only one of the 2 is configured.
> 
> In both ways the user can have a simple deployment just settings mandatory attributes.

I have implemented the first option in commit 5d38e312c8f8c3ce62c70e2f0c03f83c6b0ed099.

Thanks for the suggestions!

lcaggio (approved):
Thanks for the PR!

juliocc:
nice branch name

PR #1010, opened 2022-11-24
Enforce nonempty descriptions ending in a dot

juliocc:


tools/check_documentation.py
```diff
@@ -37,6 +37,10 @@ class State(enum.IntEnum):
   FAIL_STALE_README = enum.auto()
   FAIL_UNSORTED_VARS = enum.auto()
   FAIL_UNSORTED_OUTPUTS = enum.auto()
+  FAIL_VARIABLE_COLON = enum.auto()
```
ludoo, line 40: s/colon/dot or period/

ludoo (approved):

tools/check_documentation.py
juliocc: 🤦🏻‍♂️

PR #1034, opened 2022-12-03
feat(blueprints): get audience from tfc environment variable

Thomgrus:
# Subject

This PR is about using Workload Identity Federation with Terraform Cloud.

# Description

I propose this change because the **audience** in Terraform Cloud / Enterprise is already defined by the user in the `TFC_WORKLOAD_IDENTITY_AUDIENCE` environment variable for the workspace. So you can avoid to define it in the Terraform code as variable.

I take this opportunity to also correct a typo error in the documentation.

Thank you for the sample code, it helped me a lot to understand how to use Workload Identity Federation with Terraform Cloud.

Have a nice day ! 👋


averbuks (approved):

averbuks:
Thank you @Thomgrus for the contribution, you are right, the var is already defined, that way provider configuration is even simpler.

Issue #1042, opened 2022-12-08
Reverse lookup dns zone


chemapolo:
I'm missing reverse lookup private zone feature at the dns module. 

it is already supported by Google-beta provider : 

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dns_managed_zone#reverse_lookup

https://cloud.google.com/dns/docs/zones/managed-reverse-lookup-zones?hl=es-419

chemapolo:
Allways a pleasure to collaborate, im going to change as per your request to reverse-managed and commit the changes. Thanks  a lot.

PR #1043, opened 2022-12-08
added reverse lookup feature to module dns #1042

chemapolo:
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/1042

I've missed reverse lookup feature at dns modules while using cloud foundation fabric. 

Basically i've added a new type in variables.tf file called reverse and I've changed google_dns_managed_zone. non-public adding reverse_lookup attribute set to var.type=="reverse" in that way if type is differente to reverse it will be false and only when setting it to reverse the value will be true and will create the required reverse zone.

I've modified as well README.md file to add an example and new type.

I've run already tfdoc and tests as per your contributing guide.

Regards


ludoo (approved):
Super-clean PR, thanks a lot!

One trivial doubt: should we use `reverse-managed` instead of `reverse`, to indicate that custom reverse zones are a different thing and use the `private` type?

chemapolo:
> Super-clean PR, thanks a lot!
> 
> One trivial doubt: should we use `reverse-managed` instead of `reverse`, to indicate that custom reverse zones are a different thing and use the `private` type?

No prob ill change it, i've seen that by mistake I left   ${reverse-ip}.in-addr.arpa. in the readme esample instead of an ip. im going to change it and commit with your suggestion 'reverse-managed' Thanks a lot!

PR #1047, opened 2022-12-11
Network firewall policy module

fawzihmouda:
Since the Firewall Policies is now Google's recommended/preferred firewall construct over VPC Firewall rules. This module creates regional and/or global network firewall policy using HCL and YAML file.

This PR creates:

Global/regional Network Firewall rules:
    - firewall policy
    - Firewall Rules attached to the policy
    - usage of Tags (https://cloud.google.com/resource-manager/docs/tags/tags-overview), Service Account and 
        IP range as target type
    - attach the policy to a VPC (if required). Only 1 global policy can be attached to a VPC and as many regional policies can be attached to a VPC as long as they are in a different region
    - Firewall rules can be created with YAML or HCL
    
<s>The use of ```data.google_tags_tag_value.src_secure_tags``` and ```data.google_tags_tag_target_secure_tags``` is to use the Tag name instead of Tag ID like ```tagValues/123456``` in the firewall rules to makes the configuration easier on the user also easy to troubleshoot the rules by looking at the Tag name instead of Tag ID</s>


ludoo (commented):
Thanks for this! A few comments for discussion:

- does it make sense to infer regional/global scope from the rules themselves, or should we set scope via a variable, meaning a module invocation would be either regional or global?
- is there really a need of using data sources for tags?

Nit: can you fix the uneven spacing? There should be one empty line between variables/outputs/resources, and inside them only one optional empty line before blocks.

fawzihmouda:
> Thanks for this! A few comments for discussion:
> 
> * does it make sense to infer regional/global scope from the rules themselves, or should we set scope via a variable, meaning a module invocation would be either regional or global?
> * is there really a need of using data sources for tags?
> 
> Nit: can you fix the uneven spacing? There should be one empty line between variables/outputs/resources, and inside them only one optional empty line before blocks.

Thanks @ludoo , i have refactored my code based on your comments:

- removed the data sources for tag and uses tagValue number instead of name and kept it simple
- created a ```deployment_scope``` variable "global" or "regional" to invoke module
- add a dynamic block for ```layer4_configs``` (forgot about it last time)

modules/net-vpc-network-firewall/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+module "regional_policy" { # regional firewall rules using YAML by defining rules file location
+  source                 = "./fw"
+  regional_policy_name   = "regional-policy-1"
+  project_id             = "my-project"
+  firewall_policy_region = "australia-southeast1"
+  deployment_scope       = "regional"
+  data_folders           = ["./regional-rules"]
+}
+# tftest skip
```
juliocc, line 70: Can you enable tests for this example?

modules/net-vpc-network-firewall/README.md
```diff
@@ -0,0 +1,233 @@
...
+        - 666
+  dest_ip_ranges:
+    - 192.168.0.0/24
+    - 172.16.10.0/24
+  target_secure_tags:
+    - 446253268572
+  
+```
+
+<!-- BEGIN_TF_DOCS -->
```
juliocc, line 165: Please make sure to generate this section use the `tfdoc.py` script in the `tools/` folder.

modules/net-vpc-network-firewall/variable.tf
```diff
@@ -0,0 +1,89 @@
+/**
```
juliocc, line 1: this file should be named `variables.tf`

juliocc (changes requested):

modules/net-vpc-network-firewall/README.md
fawzihmouda: yup iv added the tftest section for my example but im not sure how to test it locally

fawzihmouda: hi @juliocc iv created a test module in tests/modules/net_vpc_network_firewall

fawzihmouda: Hi @juliocc iv created a test module for my code

juliocc:
Hi @fawzihmouda. Thanks for taking time to tackling this.

I'd prefer if the API for firewall policies was closer to what we currently have in the standard `net-vpc-firewall` module. The API we have there has converged to something that streamlines the 90%-case. 

Here's a non-exhaustive list of things we currently do in `net-vpc-firewall` that I'd like to see somehow replicated here:
- Separate variables for ingress/egress rules. 
- For rules, a `deny` boolean field instead of an `action` field with two values.
- Some way to specify default rules (see [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/net-vpc-firewall/variables.tf#L17))
- Support for named ranges

I'm also not sure how to handle the region/global aspect. Perhaps we can start with a simpler module that only manages global policies. Once we have that we can add support for the regional case.

On a more cosmetic front:
- I'm not a big fan of the `net-vpc-network-firewall` name. I don't have a better suggestion though.
- We tend to prefer dotted notation over indexing to access attributes (i.e. we prefer `each.value.name` over `each.value["name"]`)

juliocc:
Closing for now. Feel free to reopen if you want to continue working on this.

PR #1051, opened 2022-12-15
FAST: add instructions for billing export to stage 0 README

KPRepos:
Because of the billing API limitations, Though BigQuery dataset will be created within billing project, it will be empty.  BigQuery export have to be manually enabled in order for the data to be populated within the dataset. Since the Readme doesn't capture this, a note and reference link was added.


ludoo (approved):
Thanks for this!

PR #1056, opened 2022-12-19
Feature - CloudSQL pre-allocation private IP range and GKE Cluster ignore_change lifecycle hook.

itsavvy-ankur:
This PR add supports for  : 
1. CloudSQL  - provide an allocated IP range for private access. In some situations organizations have a pre-allocated /24 to be used as the private IP CIDR for CloudSQL instances. The changes in this PR help supply a pre-allocated IP range for the primary and replica instance

2. GKE-Cluster - The default pool is deleted on initial cluster creation, however the next iteration of plan/ apply using the module looks at changes on boot_disk_kms_key and always attempts to re-create the cluster there by doing a destroy / create of the cluster and nodepool. Addition of the ignore_changes lifecycle hook for the boot_disk_kms_key on the nodepool prevents the cluster re-creation on a subsequent plan / apply.


modules/cloudsql-instance/variables.tf
```diff
@@ -109,6 +109,18 @@ variable "network" {
   type        = string
 }
 
+variable "allocated_ip_range_primary" {
```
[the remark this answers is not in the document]
itsavvy-ankur: @ludoo  - Thanks for your review, I have changed the definition to 

```
variable "allocated_ip_ranges" {
  description = "(Optional)The name of the allocated ip range for the private ip CloudSQL instance. For example: \"google-managed-services-default\". If set, the instance ip will be created in the allocated range. The range name must comply with RFC 1035. Specifically, the name must be 1-63 characters long and match the regular expression a-z?."
  type = object({
    primary = optional(string)
    replica = optional(string)
  })
  default = null

}
```

modules/cloudsql-instance/variables.tf
```diff
@@ -14,6 +14,15 @@
  * limitations under the License.
  */
 
+variable "allocated_ip_ranges" {
+  description = "(Optional)The name of the allocated ip range for the private ip CloudSQL instance. For example: \"google-managed-services-default\". If set, the instance ip will be created in the allocated range. The range name must comply with RFC 1035. Specifically, the name must be 1-63 characters long and match the regular expression a-z?."
+  type = object({
+    primary = optional(string)
+    replica = optional(string)
+  })
+  default = null
```
ludoo, line 23: The default should probably be `{}` and the variable should have `nullable = false`, or you run the risk of `primary` and `replica` members not being set. Can you also remove the extra newline at line 24?

ludoo (approved):
This is good to go once the last comment has been addressed, thanks!!!

modules/cloudsql-instance/variables.tf
itsavvy-ankur: Thanks @ludoo  - I have updated the variable definition. This will also fixed the broken tests, which might need someone to kick off the workflow.

PR #1062, opened 2022-12-21
Fixes for GKE

wiktorn:
* Fix non-empty plan when spot instances are used
* Add cluster_id and recommend its use, as this prevents inconsitencies when only cluster is recreated (with no changes on node pool)


ludoo (approved):
really clean PR, thanks!

UpperM:
Thanks you !

PR #1063, opened 2022-12-21
Network dashboard: PSA ranges support, starting with Cloud SQL

aurelienlegrand:
Network dashboard: PSA ranges support, starting with Cloud SQL


blueprints/cloud-operations/network-dashboard/src/plugins/discover-cloudsql.py
```diff
@@ -0,0 +1,119 @@
+# Copyright 2022 Google LLC
```
ludoo, line 1: this could be done in cai-compute discovery, but as you said it also makes sense to have a dedicated discovery module for PSA resources; I would not add more than one though :)

ludoo:
Aurelien, I'm going to send a patch to reduce the number of loops in the series plugin, and to merge discovery with the existing CAI plugin as there's a ton of code duplication otherwise.

ludoo:
Pushed a refactor, please check if it works for you.

ludoo:
One syntax comment: please please do not ever use camel case for variable names...

ludoo (approved):

PR #1069, opened 2022-12-22
Allow tables with several column families in Bigtable

### LINKED ISSUE #1064: Bigtable module only allows a single column family
A Bigtable table may contain more than one column family ([tens and hundreds even](https://cloud.google.com/bigtable/docs/schema-design#column-families)). The current module implementation only creates the latest column family specified in the `tables` option.

For instance:
```hcl
module "my-bigtable-instance" {
 [...]
  tables = {
    test1 = {
      column_family = "cf1"
      column_family = "cf2"
      column_family = "cf3"
    }
  }
}
```

The output of `tf plan` is:

```
  + resource "google_bigtable_table" "default" {
      [...]
      + column_family {
          + family = "cf3"
        }
    }
```

The output should be something like this:

```
  + resource "google_bigtable_table" "default" {
      [...]
      + column_family {
          + family = "cf1"
        }
      + column_family {
          + family = "cf2"
        }
      + column_family {
          + family = "cf3"
        }
    }
```

The option `split_keys` should also be optional for each column family.

The [documentation of the resource `google_bigtable_table`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigtable_table#argument-reference) specifies that the option `column_family` can be specified several times.

iht:
A table in Bigtable can have more than one column family (or none), and `split_keys` is optional for `google_bigtable_table` resources. 

This PR fixes #1064. It lets the user specify 0 or more column families for each table, and makes `split_keys` an optional parameter for each declared table. 

Because column families are schema properties of a Bigtable table and `split_keys` are related to the actual values stored in a table, I don't think it makes sense to speak of default settings for those properties, and I have removed that option from this module. Now users should specify (or not) `column_families` and `split_keys` for each one of the declared tables.

The default value for a table with no additional options was `null`. I have changed that to an empty map `{}` so the default values of the corresponding `optional` parameters are used.


modules/bigtable-instance/main.tf
```diff
@@ -54,17 +51,17 @@ resource "google_bigtable_instance_iam_binding" "default" {
 }
 
 resource "google_bigtable_table" "default" {
-  for_each      = local.tables
+  for_each      = var.tables
```
ludoo, line 54: this will fail if the user passes a `null` value to the variable, I would add `nullable = false` to the variable definition, and set a default value of `{}`

ludoo (approved):
Nice change, once the single comment is addressed this can be merged, thanks Israel!

modules/bigtable-instance/main.tf
iht: New commit with those changes added. Thanks for the review, I did not test that case...

ludoo (approved):

PR #1072, opened 2022-12-22
Add gc_policy to Bigtable module, bump provider versions to 4.47

iht:
Column families have now a new property, to specify the garbage collection policy. A new option also allows to set a default policy if none is specified.

This changes the previous syntax for column families, that was necessary since the policy is column-family specific. The new syntax makes it easier to specify a policy per column family.

I had to update the default version of the Google TF provider to 4.47, in order to match the latest options available for `google_bigtable_gc_policy` resources.

### DIFF
--- a/blueprints/cloud-operations/adfs/versions.tf
+++ b/blueprints/cloud-operations/adfs/versions.tf
@@ -17,11 +17,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 4.40.0" # tftest
+      version = ">= 4.47.0" # tftest
     }
     google-beta = {
       source  = "hashicorp/google-beta"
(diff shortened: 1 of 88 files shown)


ludoo:
Israel, we use the same version for all modules and blueprints. Can you bump the provider versions in `default-versions.tf` in the root of the repo, and then replace each of the version files? I use something like this

```bash
find . -name versions.tf -exec cp default-versions.tf "{}" \;
```

iht:
> ```shell
> find . -name versions.tf -exec cp default-versions.tf "{}" \;
> ```

Done. Thanks for the tip, I was not sure about how to bump the version, whether just touching the top level file would "magically" change versions elsewhere or not.

ludoo (approved):

PR #1075, opened 2022-12-28
Add cluster replicas to Bigtable module.

iht:
This adds the possiblity to define several clusters in a Bigtable instance, which enables replication in Bigtable.

Some of the configurations options are moved inside a `map` that defines the properties of each cluster.

These changes alter the interface of the module, so any previous code using this module will have to adapt to the new options interface.

### DIFF
--- a/modules/bigtable-instance/main.tf
+++ b/modules/bigtable-instance/main.tf
@@ -15,7 +15,6 @@
  */
 
 locals {
-  num_nodes = var.autoscaling_config == null ? var.num_nodes : null
   gc_pairs = flatten([
     for table, table_obj in var.tables : [
       for cf, cf_obj in table_obj.column_families : {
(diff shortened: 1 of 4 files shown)


modules/bigtable-instance/main.tf
```diff
@@ -25,30 +24,45 @@ locals {
       }
     ]
   ])
+
+  clusters_autoscaling = flatten([
```
ludoo, line 28: it's better if you preserve the map type from the variable, as `for_each` on lists/sets uses positional indexes as keys, and it's very very easy to inadvertently drop multiple resources if an item is added or removed

ludoo (approved):
one comment on the local type, other than that it's the usual great PR, thanks Israel

modules/bigtable-instance/main.tf
iht: I am still learning the ropes with Terraform, I have changed this type back to map, thanks for the tip, I was not sure about how to do it or whether it could be done. I will be submitting a new commit soon.

iht: Commit added, resolving conversation to enable auto merge.

PR #1077, opened 2023-01-03
Added support for subnetwork data sources in net-vpc

rosmo:
Subnetwork creation is now controlled using vpc_subnet_create variable.


ludoo:
I don't think this is worth integrating in the module, IMHO existing subnet ids should be treated like constants and passed in statically, no need to add extra complexity.

ludoo:
> You could argue the same about `vpc_create` flag though - network self links are also trivially constructed. This allows one to it for: a) creation of a new VPC + subnets, b) adding subnets to existing VPC, c) quickly reusing an existing VPC + subnets.

Not really, as a frequent use case is adding subnets or configuring shared vpc, etc to an existing VPC. Subnet data source is pretty pointless IMHO.

PR #1097, opened 2023-01-18
Use terraform resource to activate Anthos Service Mesh

wiktorn:


apichick:
As we are creating the sample namespace in the blueprint maybe you also want to add the managed data plane using this command in the ansible:

```
kubectl annotate --overwrite namespace sample \
  mesh.cloud.google.com/proxy='{"managed":"false"}
```

apichick:
Need to update the tests

tests/modules/gke_hub/fixture/variables.tf
```diff
@@ -31,7 +31,7 @@ variable "features" {
     configmanagement             = true
     identityservice              = false
     multiclusteringress          = null
-    servicemesh                  = false
+    servicemesh                  = true
```
juliocc, line 34: Is there any reason to change the default for servicemesh to true?

modules/gke-hub/main.tf
```diff
@@ -38,7 +38,7 @@ resource "google_gke_hub_membership" "default" {
   provider      = google-beta
   for_each      = var.clusters
   project       = var.project_id
-  membership_id = each.key
+  membership_id = reverse(split("/", each.value))[0] # forces re-enrollment of the cluster in the fleet in case when cluster is recreated
```
juliocc, line 41: Is there a possibility for this to break if you have clusters with the same name in different projects?

Perhaps you can use something like `replace(each.value, "/", "-")` or keep `each.key` as part of the membership id.

modules/gke-hub/main.tf
```diff
@@ -70,6 +70,20 @@ resource "google_gke_hub_feature" "default" {
   }
 }
 
+resource "google_gke_hub_feature_membership" "servicemesh" {
+  provider   = google-beta
+  for_each   = var.features.servicemesh ? var.clusters : {}
+  project    = var.project_id
+  location   = "global"
```
juliocc, line 77: `location` should probably be specified by the use instead of having it hard-coded

tests/modules/gke_hub/fixture/variables.tf
wiktorn: This is only for test fixture, so the test will cover also servicemesh.

juliocc: Ah sorry. I missed that this was the fixture.

modules/gke-hub/main.tf
wiktorn: I've tested this again and it looks like my workaround was needed only when I did not configure workload identity endpoints. Once they are configured, then hub membership attributes are anyway forcing replacement, so there is no need for change here.

modules/gke-hub/main.tf
wiktorn: "global" is the only location that works here.

juliocc:
LGTM. I defer to @apichick for final approval.

juliocc (approved):

PR #1101, opened 2023-01-19
First batch of testing updates to core modules

juliocc:
Updating the following modules to use the new example-based testing framework:
- dns
- gke-cluster
- gke-nodepool
- net-vpc

This PR also 
- adds the `--example` flag to the `plan_summary.py` tool to simulate execution as it happens within the examples contained in READMEs
- Fixes long-standing bug in `net-vpc` dns policies not being applied correctly
- Adds several new tests/examples to the modules mentioned above
- Removes several python-based tests which are now handled as tested examples


ludoo (approved):
💯

PR #1106, opened 2023-01-20
Network Dashboard: PSA support for Filestore and Memorystore

aurelienlegrand:
Network Dashboard: IP utilization support for 2 more PSA based services: Filestore and Memorystore
Next step: GKE and secondary ranges.


blueprints/cloud-operations/network-dashboard/src/plugins/discover-cai.py
```diff
@@ -82,10 +84,16 @@ def _handle_resource(resources, asset_type, data):
...
-      'name': attrs['name'],
-      'self_link': _self_link(attrs['selfLink']),
-      'assetType': asset_type
+      'id':
+          attrs.get('id'),
+      'name':
+          attrs['name'],
+      # Some resources (ex: Filestore) don't have a self_link, using parent + name in that case
+      'self_link':
+          data['parent'] + '/' + attrs['name']
```
ludoo, line 93: can you use an f-string instead of adding strings? something like `f'{data["parent"]}/{attrs["name"]}'`

ludoo: do we really need the filestore self link anywhere? otherwise I would just ignore it

ludoo (approved):
Looks good! Dropped a trivial comment, and I'm not entirely sure about the change in formatting but I added the net dashboard folder to our Python formatting check, so we will catch any difference there.

PR #1112, opened 2023-01-26
Add HTTPS frontend with SNEG example

juliodiez:
Serverless NEGs with HTTPS frontends require setting values differently than other backends. Adding example to explain this case.


ludoo (approved):
Thanks Julio!!!

PR #1113, opened 2023-01-26
compute-vm module align attached device_name with disk name, default value of console attach

rafstef:
Align attached device_name with disk name, default value of console attach


ludoo:
Hi, we don't prefix `device_name` on purpose, as that is the name that will created in the device tree `/dev/disk/by-id/google-{{device_name}}`. Thanks for this, but I have to close it.

PR #1119, opened 2023-01-29
Multi-Cluster Ingress gateway api config

wiktorn:
* Introduce new feature `gateway_api` for GKE clusters
* Bump version of terraform provider to 4.48  as this is the first version that supports `gateway_api_config` block
* Make `enable_features` in **modules/gke-cluster** consistent between default value and what is set in optionals
* Introduce `optional` into **modules/gke-hub** `features` variable
* Add  `multicluster-ingress` and `multicluster-discovery` robot Service Accounts to **module/project**, JIT Service Account for `multiclusteringress.googleapis.com`


juliocc (approved):
You're changing the [default value for workload_identity from false to true](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1119/files#diff-8856338b65ccbc2f5e1332071dd386ae281f70b3a0b0217f60a713125e3e7d46R99). Is this intended?

wiktorn:
Yes, this is intended, as we had inconsistency between what was the default value and what was provided as optional default value breaking POLA when someone was setting something different than `workload_identity`.

PR #1128, opened 2023-02-03
Remove info about non-existing vpc-peering-*.tf files

skalolazka:


juliocc (approved):

ludoo:
can we wait before merging, or mt PR will have a conflict :)

PR #1131, opened 2023-02-06
Add Autopilot Support for cluster_autoscaling Configuration in GKE Module

tacchino:
Add support for the cluster_autoscaling configuration block in the gke-cluster module when autopilot = true.

Previously this would fail because the enabled parameter could not be set with autoscaling enabled. Changed to set enabled = null and added support for the auto_provisioning_defaults configuration, primarily to allow for specification of a non-default service account

### DIFF
--- a/modules/gke-cluster/main.tf
+++ b/modules/gke-cluster/main.tf
@@ -146,9 +146,9 @@ resource "google_container_cluster" "cluster" {
         for_each = var.cluster_autoscaling.auto_provisioning_defaults != null ? [""] : []
         content {
           boot_disk_kms_key = var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key != null ? var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key : null
-          image_type = var.cluster_autoscaling.auto_provisioning_defaults.image_type != null ? var.cluster_autoscaling.auto_provisioning_defaults.image_type : null
-          oauth_scopes = var.cluster_autoscaling.auto_provisioning_defaults.oauth_scopes != null ? var.cluster_autoscaling.auto_provisioning_defaults.oauth_scopes : null
-          service_account = var.cluster_autoscaling.auto_provisioning_defaults.service_account != null ? var.cluster_autoscaling.auto_provisioning_defaults.service_account : null
+          image_type        = var.cluster_autoscaling.auto_provisioning_defaults.image_type != null ? var.cluster_autoscaling.auto_provisioning_defaults.image_type : null
+          oauth_scopes      = var.cluster_autoscaling.auto_provisioning_defaults.oauth_scopes != null ? var.cluster_autoscaling.auto_provisioning_defaults.oauth_scopes : null
+          service_account   = var.cluster_autoscaling.auto_provisioning_defaults.service_account != null ? var.cluster_autoscaling.auto_provisioning_defaults.service_account : null
         }
       }
       dynamic "resource_limits" {
--- a/modules/gke-cluster/main.tf
+++ b/modules/gke-cluster/main.tf
@@ -140,7 +140,17 @@ resource "google_container_cluster" "cluster" {
   dynamic "cluster_autoscaling" {
     for_each = var.cluster_autoscaling == null ? [] : [""]
     content {
-      enabled = true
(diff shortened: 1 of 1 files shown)


modules/gke-cluster/main.tf
```diff
@@ -140,7 +140,17 @@ resource "google_container_cluster" "cluster" {
   dynamic "cluster_autoscaling" {
     for_each = var.cluster_autoscaling == null ? [] : [""]
     content {
-      enabled = true
+      enabled = var.enable_features.autopilot ? null : true
+
+      dynamic "auto_provisioning_defaults" {
+        for_each = var.cluster_autoscaling.auto_provisioning_defaults != null ? [""] : []
+        content {
+          boot_disk_kms_key = var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key != null ? var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key : null
```
juliocc, line 148: Would it make sense to remove the ternary in all these cases?
```
          boot_disk_kms_key = var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key
          image_type        = var.cluster_autoscaling.auto_provisioning_defaults.image_type
          oauth_scopes      = var.cluster_autoscaling.auto_provisioning_defaults.oauth_scopes
          service_account   = var.cluster_autoscaling.auto_provisioning_defaults.service_account
```

juliocc:
@danielmarzini can you review this?

modules/gke-cluster/main.tf
tacchino: Ah, yeah, you are right, optional() call in the variable takes care of that

modules/gke-cluster/main.tf
```diff
@@ -140,7 +140,17 @@ resource "google_container_cluster" "cluster" {
   dynamic "cluster_autoscaling" {
     for_each = var.cluster_autoscaling == null ? [] : [""]
     content {
-      enabled = true
+      enabled = var.enable_features.autopilot ? null : true
```
juliocc, line 143: I think instead of setting `enabled` when autopilot is request, just disable the whole block (adding the extra check to the outer cluster_autoscaling block). Something like this:
```hcl
dynamic "cluster_autoscaling" {
    for_each = (var.cluster_autoscaling == null || var.enable_features.autopilot) ? [] : [""]
```

tacchino: @juliocc  that would remove the feature this PR adds which is the ability to specify auto_provisioning_defaults, a block that is contained within cluster_autoscaling

ludoo: it will remove it only when autopilot is enabled, so as not to break autopilot AFAIK

tacchino: Not exactly. The problem I'm trying to address here is that we should be able to specify a non-default node service account when autopilot is enabled. When the cluster_autoscaling block is provided enabled is defaulted to true and it fails at the plan stage

`Error: Conflicting configuration arguments
│
│   with module.test_gke_cluster.google_container_cluster.cluster,
│   on .terraform/modules/test_gke_cluster/modules/gke-cluster/main.tf line 17, in resource "google_container_cluster" "cluster":
│   17: resource "google_container_cluster" "cluster" {
│
│ "cluster_autoscaling.0.enabled": conflicts with enable_autopilot`

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#nested_cluster_autoscaling
> (Optional) Whether node auto-provisioning is enabled. Must be supplied for GKE Standard clusters, true is implied for autopilot clusters

ludoo: I remember a provider bug that prevented using a different SA with autopilot, has that been fixed? I can try finding it...

tacchino: I've got a cluster up and running with a custom SA using the change implemented here

ludoo: ahhhh nice!

juliocc: let's approve then. We can always come back if anything breaks

juliocc:
@tacchino can you rebase this pr too?

juliocc (approved):

tacchino:
@juliocc done. thanks!

PR #1136, opened 2023-02-06
Dev foundation

dominicrjames:


ludoo:
Can you please provide a meaningful PR title and description? We also tend to minimize cost in module defaults as they are heavily used for testing, so a better choice would be `e2-micro` or `e2-small`.

PR #1140, opened 2023-02-07
CloudSQL Backup Configuration: Support Point In Time Recovery

tacchino:
Add support for the point_in_time_recovery_enabled property in CloudSQL backup_configuration
Use optional() for configuration map in order to maintain backwards compatibility with previous configurations

### DIFF
--- a/modules/cloudsql-instance/variables.tf
+++ b/modules/cloudsql-instance/variables.tf
@@ -37,14 +37,15 @@ variable "availability_type" {
 
 variable "backup_configuration" {
   description = "Backup settings for primary instance. Will be automatically enabled if using MySQL with one or more replicas."
+  nullable    = false
   type = object({
-    enabled                        = optional(bool)
-    binary_log_enabled             = optional(bool)
-    start_time                     = optional(string)
+    enabled                        = optional(bool, false)
+    binary_log_enabled             = optional(bool, false)
+    start_time                     = optional(string, "23:00")
     location                       = optional(string)
-    log_retention_days             = optional(number)
+    log_retention_days             = optional(number, 7)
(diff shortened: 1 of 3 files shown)


modules/cloudsql-instance/variables.tf
```diff
@@ -38,20 +38,22 @@ variable "availability_type" {
...
-    location           = null
-    log_retention_days = 7
-    retention_count    = 7
+    enabled                        = false
+    binary_log_enabled             = false
+    start_time                     = "23:00"
+    location                       = null
+    log_retention_days             = 7
+    point_in_time_recovery_enabled = null
+    retention_count                = 7
```
juliocc, line 56: Can you put the defaults as part of the fields definition? This way users can specify only the ones they care about:

```hcl
  type = object({
    enabled                        = optional(bool, false)
    binary_log_enabled             = optional(bool, false)
    start_time                     = optional(string, "23:00")
    location                       = optional(string)
    log_retention_days             = optional(number, 7)
    point_in_time_recovery_enabled = optional(bool)
    retention_count                = optional(number, 7)
  })
  default = {}
```

It might also be a good time to set the variable as non-nullable.

juliocc (changes requested):

modules/cloudsql-instance/variables.tf
tacchino: Sure, done, thanks for the catch

juliocc (approved):

juliocc:
@tacchino thanks for this PR. Can you rebase and merge?

tacchino:
@juliocc up to date with master now. let me know if I didn't get that right

juliocc:
All good, thanks @tacchino!

PR #1144, opened 2023-02-09
Fixes for service-mesh example in gke-hub

wiktorn:
* Fix missing subnets and duplicate service account names for nodepools
* add fix for missing IAM role for fleet service account
* enable dataplane_v2 and workload_identity as they are required for Service Mesh


ludoo (approved):
I have no clue what you are doing but I trust you :)

PR #1153, opened 2023-02-13
net-vpc - add missing iam properties to factory_subnets

jamesdalf:
Module: net-vpc 
When creating subnets via a data folder input, the properties iam_groups, iam_users, and iam_service_accounts are ignored. These are required in order to bind "roles/compute.networkUser" 

The default value in the try function is an empty list as formatting in _factory_subnets_iam cannot be performed on a null value.


jamesdalf:
Hi, regarding the example tests. I believe the only failing test occurs because the extra resource 'google_compute_subnetwork_iam_binding' was ignored in the previous module and therefore not accounted for in the tests. Now that the iam properties are there, an extra resource would be created.

juliocc (approved):
Good catch. Thanks for the PR

I fixed the tests for you.

Issue #1157, opened 2023-02-14
gke-nodepool module retrieve an Error: Unsupported attribute while running terraform plan.


chemapolo:
While using Module [gke-nodepool](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/gke-nodepool) with a pod_range config block like:

```hcl
  pod_range = {
    secondary_pod_range = {
      cidr   = "X.X.X.X/14"
      create = true
      name   = "pods_range"
    }
  }
```

It returns the next error:

```
_Error: Unsupported attribute
on .terraform/modules/nodepool-1/main.tf line 119, in resource "google_container_node_pool" "nodepool":
      pod_ipv4_cidr_block = var.pod_range.cidr
This object does not have an attribute named "cidr".

Error: Unsupported attribute
on .terraform/modules/nodepool-1/main.tf line 120, in resource "google_container_node_pool" "nodepool":
      pod_range           = var.pod_range.name
This object does not have an attribute named "name".

Error: Unsupported attribute
on .terraform/modules/nodepool-1/main.tf line 118, in resource "google_container_node_pool" "nodepool":
      create_pod_range    = var.pod_range.create
This object does not have an attribute named "create"._
```

Having a look at the [variables.tf](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/gke-nodepool/variables.tf) definition looks correct.

```hcl
variable "pod_range" {
  description = "Pod secondary range configuration."
  type = object({
    secondary_pod_range = object({
      cidr   = optional(string)
      create = optional(bool)
      name   = string
    })
  })
  default = null
}

```
Nevertheless at [main.tf](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/gke-nodepool/main.tf) at the affected lines you can see:

```hcl
  dynamic "network_config" {
    for_each = var.pod_range != null ? [""] : []
    content {
      create_pod_range    = var.pod_range.create
      pod_ipv4_cidr_block = var.pod_range.cidr
      pod_range           = var.pod_range.name
    }
  }
```
It s calling _**var.pod_range.**_  object but pod_range contains an object itself called _**secondary_pod_range**_ which contains the proper attributes.

chemapolo:
[...]

PR #1158, opened 2023-02-14
changed pod_range reference to include secondary_pod_range issue #1157

chemapolo:
replaced .secondary_pod_range by var.pod_range.secondary_pod_range that is the object which contins create, cidr an name attributes.


ludoo (approved):
Thanks for fixing this!

PR #1162, opened 2023-02-18
Fix Terraform formatting and add module_prefix attribute to modules_config

antonkovach:
Fix the formatting of Terraform files in 0-cicd-github under repository population. Without proper formatting  generated repositories may show formatting mismatches.
The module_prefix attribute has been added to the modules_config variable to handle non modules only repository use cases. The module_prefix attribute specifies the location of Terraform modules within the repository, which can help organize the modules and avoid naming conflicts.


fast/extras/0-cicd-github/main.tf
```diff
@@ -143,8 +144,8 @@ resource "github_repository_file" "default" {
     endswith(each.value.name, ".tf") && local.modules_repo != null
     ? replace(
       file(each.value.file),
-      "/source\\s*=\\s*\"../../../modules/([^/\"]+)\"/",
-      "source = \"git@github.com:${local.modules_repo}.git//$1${local.modules_ref}\"" # "
+      "/source(\\s*)=\\s*\"../../../(modules/[^/\"]+)\"/",
+      "source$1= \"git@github.com:${local.modules_repo}.git//$2${local.modules_ref}\"" # "
```
ludoo, line 148: Anton, what we target is a modules-only repository, where the `modules/` prefix is not used. Can you revert this change?

antonkovach: Of course.

ludoo: Thanks. :) Thinking about it, we might want to introduce an optional "module prefix" in the variables, to account for different usecases. Feel free to do it, or ignore and we can do it later in a subsequent PR. Thanks a lot for this and just let me know when you feel like it can be merged.

ludoo (approved):

fast/extras/0-cicd-github/main.tf
antonkovach: @ludoo , could you please review the 'module_prefix' attribute I created in 'modules_config'? It's intended to handle use cases for non-root modules.

ludoo (approved):

fast/extras/0-cicd-github/main.tf
ludoo: Thanks! Looks great. :)

PR #1163, opened 2023-02-18
Projects-data-source module new version

averbuks:
The new version of the module is based on the Cloud Asset Inventory API. It now only requires one API call to handle the logic, while in the previous version there were many to unpack the directory tree. The new version also supports exclusions for projects or folders.


modules/projects-data-source/README.md
```diff
@@ -15,11 +20,7 @@ module "my-org" {
 }
 
 output "projects" {
-  value = module.my-org.projects
-}
-
-output "folders" {
-  value = module.my-org.folders
+  value = module.my-org.projects_numbers
```
ludoo, line 23: this looks like a typo (`project_numbers`)

ludoo (approved):

modules/projects-data-source/README.md
averbuks: Ohh, good catch. Fixed now

PR #1169, opened 2023-02-21
Added support for INTERNAL_SELF_MANAGED proxies in net-ilb-l7

rosmo:
As most of the configuration is fairly similar, add support for `INTERNAL_SELF_MANAGED` proxies in `net-ilb-l7` (this is the type used by Traffic Director).


modules/net-ilb-l7/variables.tf
```diff
@@ -26,6 +26,12 @@ variable "description" {
   default     = "Terraform managed."
 }
 
+variable "global" {
```
ludoo, line 29: Can we avoid using `load_balancing_scheme` and infer it from `global`?

ludoo (commented):
Taneli, can you add a usage example in the README, which will also serve to test this change?

ludoo:
@rosmo can you add an example to the README, or shall we close this and pick it up later?

ludoo:
I am closing this, feel free to reopen once we a clearer use case that can be translated into tests and examples.

PR #1175, opened 2023-02-24
Serverless networking program

juliodiez:
Blueprint to privately expose a Cloud Run service in a variety of use cases.

The content of this blueprint corresponds to the chapter 'Developing an enterprise application - The corporate environment' of the Serverless Networking Guide (to be released soon).


ludoo:
Wow! Can you add tests (examples if they make sense, or regular tests) otherwise it will be broken as soon as we touch one of the modules it uses.

juliodiez:
I added tests at the end of README that covers most of them, possibly missing modules only of the use case with VPN. I'll add one for that.

ludoo (approved):
thanks Julio :)

ludoo:
@juliodiez can you merge this? I can do it otherwise.

Issue #1185, opened 2023-02-27
https health check config is incorrectly mapped to http 


tanmaybangale:
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/net-ilb/health-check.tf#L76
It should be below

`  
dynamic "https_health_check" {
    for_each = local.hc_https ? [""] : []
    content {
      host               = local.hc.https.host
      port               = local.hc.https.port
      port_name          = local.hc.https.port_name
      port_specification = local.hc.https.port_specification
      proxy_header       = local.hc.https.proxy_header
      request_path       = local.hc.https.request_path
      response           = local.hc.https.response
    }
  }

`

jogoldberg:
Problem also exists in compute-mig module.
Closed by #1194

ludoo:
Thanks you Joshua!

PR #1194, opened 2023-02-28
Fix HTTPS health check mismapped to HTTP in compute-mig and net-ilb modules

jogoldberg:
## Changed
* **compute-mig** and **net-ilb** module: https healthcheck mapped properly to local.hc.https (was previously mismapped to local.hc.http) #1185


ludoo:
Thanks a lot for this. :)

ludoo (approved):

PR #1196, opened 2023-02-28
Fix compute-vm:CloudKMS test for provider>=4.54.0

dan-farmer:
* TF provider >= 4.54.0 now returns `rsa_encrypted_key` for `google_compute_disk.disks["attached-disk"]` (see hashicorp/terraform-provider-google#4448)
* Add this field to expected model to fix test assertion failure
* Update required TF provider to 4.55.0 (latest) since the assertion will now fail with <4.54.0, which do not return `rsa_encrypted_key`
  * Updated the whole repo on advice from @ludoo

Please sanity-check me that bumping the TF provider version is the right thing to do, and that I have done it appropriately.

The exact test in question is `tests/examples/test_plan.py::test_example[modules/compute-vm:Disk encryption with Cloud KMS]`. Aassertion copied below, for reference - line-breaks added for clarity:
```
AssertionError: disk_encryption_key at module.kms-vm-example.google_compute_disk.disks["attached-disk"] failed.

Got
`[{'kms_key_self_link': 'kms_key_self_link', 'kms_key_service_account': None, 'raw_key': None, 'rsa_encrypted_key': None}]`,

expected
`[{'kms_key_self_link': 'kms_key_self_link', 'kms_key_service_account': None, 'raw_key': None}]`
```

### DIFF
--- a/blueprints/apigee/bigquery-analytics/versions.tf
+++ b/blueprints/apigee/bigquery-analytics/versions.tf
@@ -17,11 +17,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 4.50.0" # tftest
+      version = ">= 4.55.0" # tftest
     }
     google-beta = {
       source  = "hashicorp/google-beta"
-      version = ">= 4.50.0" # tftest
+      version = ">= 4.55.0" # tftest
     }
   }
 }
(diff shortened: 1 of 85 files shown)


dan-farmer:
> Dan, thanks for this. We always update versions for the whole repo, using the `default-versions.tf` file in the root as a source. If you feel up to it, just bump versions there then run something like `find . -name versions.tf -exec cp default-versions.tf "{}" \;`.

@ludoo Done. Happy to make more changes as necessary - LMK.

ludoo:
> @ludoo Done. Happy to make more changes as necessary - LMK.

If you don't see a merge button once checks run ping me, I'll add you to the repo.

PR #1203, opened 2023-03-02
Update subnet sample yaml files to use subnet_secondary_ranges

jmound:
This change is a replace of `secondary_ip_range:` to `secondary_ip_ranges:` in the code samples and tests.

The net-vpc module's `data_folder` gets parsed expecting the form with the "s": `secondary_ip_ranges:`

https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/1f713557c21c4b38154138e58e324cf2ccb51d82/modules/net-vpc/subnets.tf#L33

Testing:
* A terraform apply using the `secondary_ip_range:` form causes no secondary ranges to be managed because of the try(..., null). 
* Using the plural form works as intended.


tests/modules/net_vpc/examples/factory.yaml
```diff
@@ -31,7 +31,7 @@ values:
     project: my-project
     region: europe-west1
     role: null
-    secondary_ip_range:
+    secondary_ip_ranges:
```
juliocc, line 34: this part should be ok, no need to update any of the files in this folder.

juliocc (changes requested):
The test yaml files should be fine since the provider uses `secondary_ip_range` (without the s).

The factory files in the FAST networking stages indeed had to be fixed. Thanks for spotting this.

tests/modules/net_vpc/examples/factory.yaml
jmound: Fixed.

juliocc (approved):

PR #1212, opened 2023-03-04
feat: skip committing unchanged files in 0-cicd-github

antonkovach:
The cloud-foundation-fabric repository is constantly evolving. Running 0-cicd-github multiple times may be necessary, for instance, when files have changed, and you need to update the repository. However, committing unchanged files can make it challenging to track changes and understand the repository's history. To address this issue, this feature introduces the capability to skip committing unchanged files and improve the repository's readability.

This feature checks whether any changes have been made to the files before committing them. If no changes are detected, the files will not be committed. This results in a cleaner and more focused commit history that is easier to follow and understand.


ludoo (approved):
good idea, thanks for this :)

PR #1213, opened 2023-03-05
feat: Add Pull Request support to 0-cicd-github

antonkovach:
The cloud-foundation-fabricre pository is continually evolving, and to help keep up with the changes, it would be beneficial to introduce a pull request mechanism to review and approve changes. This feature is 100% backward compatible, and by default, no pull request is created, and changes are committed directly to the main branch. However, an optional variable pull_request_config can be used to configure the title, body, head_ref, and base_ref of the pull request that will be created for the initial population or update of files. To create a pull request, in pull_request_config set the create attribute to true. base_ref defaults to main, and head_ref to the name of the head branch. If the head branch doesn't exist, it will be created from the base_ref branch.

In the future, it might be worth considering enabling the pull request mechanism by default and committing changes to a branch other than the hardcoded main branch.

I have been using this feature for several days, and it works as expected.

Please review the idea, changes and suggest any possible improvements or approve the pull request.


ludoo (approved):
Really nice PR, thanks a lot as always!

PR #1216, opened 2023-03-06
Use composite action for test workflow prerequisite steps

ludoo:


.github/workflows/tests.yml
```diff
@@ -126,38 +71,11 @@ jobs:
...
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ env.PYTHON_VERSION }}
-          cache: 'pip'
-          cache-dependency-path: 'tests/requirements.txt'
-
-      - name: Set up Terraform
-        uses: hashicorp/setup-terraform@v2
+      - name: Call Composite Action say-hello
```
juliocc, line 74: replace say-hello?

ludoo: nice catch, lemme fix :)

PR #1226, opened 2023-03-08
Fix policy_based_routing.sh script on simple-nva module

simonebruzzechesse:
Fixed policy_based_routing.sh script on simple-nva module when dealing with network interfaces either not associated to an internal TCP Load Balancer (infinite loop in while cycle) or in case of multiple LB pointing to the NIC (probably ip route command failing with multiple line response received from ip r show table local).

As per documentation added in the script executing a curl request to  http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/$IF_NUMBER/forwarded-ips/ would return "" when no load balancer is associated to that network interface or a new line separated list of indexes (0 1 ..), one for each load balancer associated. A further request to http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/$IF_NUMBER/forwarded-ips/[0,1,..] will return the specific LB ip address.

We could evaluate an improvement using that curl instead of waiting for the route to be installed on the local routing table by the guest agent.


ludoo (approved):

simonebruzzechesse:
Review the script since previous version was not taking into account load balancer configured after instance creation. In the newer version policy_based_routing.sh script runs on background which allows routing table to be configure properly via the start_routing.sh script. The policy_based_routing.sh script will check every two seconds for new load balancer routes configured from the guest agent, for each LB available on that network interface it will check if the policy based route is configured, otherwise it will do that.

Overall this should solve the previous state issue with network interfaces not having LB associated (stuck in the while loop without configuring VM routes) as well as managing more than 1 LB associated to the network interface (i believe the previous version of the script would have failed in such circumstance.

modules/cloud-config-container/simple-nva/files/policy_based_routing.sh
```diff
@@ -15,20 +15,35 @@
...
+
+# Skip route configuration when no LB is associated to the network interface
+if [ ! -z "$IF_LB" ] ; then
+  # Convert whitespace separated list of IPs to array
+  IPS_LB=($IP_LB)
+  for IP in "${IPS_LB[@]}" 
+  do
+    ip rule add from $IP/32 table hc-$IF_NAME
+  done
+fi
```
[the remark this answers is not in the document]
simonebruzzechesse: Done!

modules/cloud-config-container/simple-nva/files/policy_based_routing.sh
```diff
@@ -15,20 +15,27 @@
...
+  for IP in "${IPS_LB[@]}"
+  do
+    # Configure hc routing table if not available for this network interface
+    grep -qxF "$((200 + $IF_NUMBER)) hc-$IF_NAME" /etc/iproute2/rt_tables || {
+      echo "$((200 + $IF_NUMBER)) hc-$IF_NAME" >>/etc/iproute2/rt_tables
+      ip route add $IF_GW src $IF_IP dev $IF_NAME table hc-$IF_NAME
+      ip route add default via $IF_GW dev $IF_NAME table hc-$IF_NAME
+    }
+    # configure PBR route for LB
+    ip rule list | grep -qF "$IP" || ip rule add from $IP/32 table hc-$IF_NAME
```
LucaPrete, line 38: can we please add something it also removes the route if the LB is no longer there?

simonebruzzechesse: Done!

LucaPrete (approved):

PR #1230, opened 2023-03-09
Update contributing guide with new test framework

juliocc:


CONTRIBUTING.md
```diff
@@ -642,19 +651,247 @@ Options:
...
+  google_dns_managed_zone: 1
+  google_dns_record_set: 1
+```
+
+#### Using external files
+
+In some situations you module might require additional files to properly test your module. This is a common situation with modules that implement [factories](blueprints/factories/README.md) that drive the creation of resource from YAML files. If you're in this situation, you can still example-based tests as described below:
+
+- create your regular hcl code block example and add the `tftest` directive as described above.
+- create a new code block with the contents of you additional file and use the `tftest-file` directive. You have to specify a label for the file and a relative path where the file will live.
```
ludoo, line 760: `your additional`

CONTRIBUTING.md
```diff
@@ -642,19 +651,247 @@ Options:
...
+A myhost:
+ ttl: 600
+ records: ["10.10.0.1"]
+```
+
+Note that you use the `files` parameters together with `inventory` to allow more fine-grained assertions. Please review the [subnet factory](modules/net-vpc#subnet-factory) in the `net-vpc` module for example of this.
+
+#### Running tests for specific examples
+
+As mentioned before, we use pytest as our test runner, so you can use any of the standard test selection options available in pytest.
```
ludoo, line 791: code syntax on pytest?

ludoo (approved):
awesome stuff

PR #1234, opened 2023-03-10
Fixed connection tracking configuration on LB backend in net-ilb module 

simonebruzzechesse:
As per documentation the tracking_mode argument in the connection_tracking section of the backend_service_config supports  PER_SESSION or PER_CONNECTION, instead we were passing a boolean value through the track_per_session variable and this was resulting in an error. We could also update the logic and set the tracking_mode value to PER_SESSION when track_per_session was true but updating the variable to tracking_mode seems to be a more general solution for this case.


modules/net-ilb/variables.tf
```diff
@@ -27,7 +27,7 @@ variable "backend_service_config" {
     connection_tracking = optional(object({
       idle_timeout_sec          = optional(number)
       persist_conn_on_unhealthy = optional(string)
-      track_per_session         = optional(bool)
+      tracking_mode             = optional(string)
```
ludoo, line 30: if the choice is just between two values, I would still use a boolean as a string requires figuring out the actual value needed, and is prone to typos.

simonebruzzechesse: You are right! I realigned back and fixed the issue with boolean, now I assign PER_SESSION and PER_CONNECTION accordingly, when null it defaults to PER_CONNECTION

ludoo: awesome, thanks Simone!

modules/net-ilb/main.tf
```diff
@@ -80,11 +80,7 @@ resource "google_compute_region_backend_service" "default" {
         : null
       )
       idle_timeout_sec = local.bs_conntrack.idle_timeout_sec
-      tracking_mode = (
-        local.bs_conntrack.track_per_session != null
-        ? local.bs_conntrack.track_per_session
-        : null
-      )
+      tracking_mode    = try(local.bs_conntrack.track_per_session ? "PER_SESSION" : "PER_CONNECTION", null)
```
ludoo, line 83: this should probably be

`try(local.bs_conntrack.track_per_session, null) ? "PER_SESSION" : "PER_CONNECTION"`

simonebruzzechesse: The suggested code will work only if track_per_session is either true or false, in case it is null it returns error because the condition value will be null ("The condition value is null. Conditions must either be true or false."). The try statement implemented will check if the variable is true or false and assign the value accordingly, in case it is null the first expression will result in an error (same of the previous case) and the try will assign value null to the tracking_mode (null will then result in the default value of "PER_CONNECTION". Am I missing something?

ludoo: nope, you're right and I should not multitask :)

ludoo (approved):

PR #1240, opened 2023-03-11
feat: Enable populating of data directory and .sample files and update dependencies in 0-cicd-github

antonkovach:
The Readme.md files reference the data directory and .sample files, but the code did not allow for their populating. This update enables the copying of the data directory and .sample files, with the data directory being populating as a data.sample directory to prevent overwriting any existing data directory.

Additionally, dependencies have been updated by adding the depends_on section to several resources to ensure that the dependencies are in the correct order. This update addresses some states that were not being handled previously.

There is a minor known issue with Pull Request creation in the current state of the code. The Pull Request is only created after the first run has occurred. A fix for this issue is currently being worked on and will be addressed in a separate Pull Request. However, this issue does not affect the main functionality of the code.

Please review the changes and suggest any possible improvements or approve the pull request.

### DIFF
--- a/fast/extras/0-cicd-github/main.tf
+++ b/fast/extras/0-cicd-github/main.tf
@@ -20,11 +20,13 @@ locals {
       for f in concat(
         [for f in fileset(path.module, "${v.populate_from}/*.svg") : f],
         [for f in fileset(path.module, "${v.populate_from}/*.md") : f],
+        [for f in fileset(path.module, "${v.populate_from}/*.sample") : f],
+        [for f in fileset(path.module, "${v.populate_from}/data/**/*.*") : f],
         [for f in fileset(path.module, "${v.populate_from}/*.tf") : f]
         ) : {
         repository = k
         file       = f
-        name       = replace(f, "${v.populate_from}/", "")
+        name       = replace(replace(f, "${v.populate_from}/", ""), "data/", "data.sample/")
       }
     ] if v.populate_from != null
   ])
@@ -143,14 +145,18 @@ resource "github_branch" "default" {
     : {}
   )
   repository    = each.key
-  branch        = var.pull_request_config.head_ref
-  source_branch = var.pull_request_config.base_ref
+  branch        = var.pull_request_config.create == true ? var.pull_request_config.head_ref : "main"
+  source_branch = var.pull_request_config.create == true ? var.pull_request_config.base_ref : ""
+
+  depends_on = [
+    github_repository.default
+  ]
 }
 
 resource "github_repository_file" "default" {
   for_each   = local.modules_repo == null ? {} : local.repository_files
   repository = local.repositories[each.value.repository]
-  branch     = try(var.pull_request_config.head_ref, "main")
+  branch     = var.pull_request_config.create == true ? var.pull_request_config.head_ref : "main"
   file       = each.value.name
   content = (
     endswith(each.value.name, ".tf") && local.modules_repo != null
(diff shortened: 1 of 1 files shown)


ludoo:
Hey Anton, I'm not entirely sure we want to copy sample files: in my opinion once you have CI/CD set up you don't need samples and have the correct files in place.
On data I have mixed feelings: I think actual data for factories should live into a different folder and be referenced explicitly in the factories data path variable. This avoids conflicts between the dummy files we provide, and actual files used for customer infrastructure.

antonkovach:
I agree with your opinion. I also had mixed feelings about the idea. From my own experience, it took me several hours to deduce the functionality of the modules and variables from the source code, including the structure and contents of the data directory. Similarly, I did not expect to find terraform.tfvars.sample in the original repo. Having copies of these files in the current repo would have saved me a lot of time as I was reading the Readme.md file and getting acquainted with the provisioned repo. Perhaps we could add a note in the Readme.md files to indicate the location of such files in the original repo.

ludoo:
I see your points, and also that we have two different approaches to CI/CD.

What I and most of my colleagues do, is to initially apply the stages manually a few times until everything is working well, the custom tfvars have the desired values, etc. then transfer the files to CI/CD, at which point you already have tfvars and factory data has been customized.

For our approach, moving tfvars and data potentially creates a problem, or at least adds non-relevant files to the repo.

Why don't we do both the things you suggested, with twist? We add a notice to the READMEs, and we also implement this change but we gate it behind a boolean variable, possibly defaulting to false? And we explain in the cicd github stage that it exists, and should be used when CI/CD is used right from the start.

Would this work for you?

antonkovach:
@ludoo  I would like to check the proposal. Which version would be better?

Simple:
```hcl
repositories = {
  devops_gcp_fast_00_bootstrap = {
    name = "devops-gcp-fast-00-bootstrap"
    create_options = {
      description = "FAST bootstrap."
    }
    features = {
      issues = true
    }
    populate_from    = "../../stages/0-bootstrap"
    populate_samples = true
  }
# tftest skip
```

More future-proof:
```hcl
repositories = {
  devops_gcp_fast_00_bootstrap = {
    name = "devops-gcp-fast-00-bootstrap"
    create_options = {
      description = "FAST bootstrap."
    }
    features = {
      issues = true
    }
    populate_from    = "../../stages/0-bootstrap"
    populate_options = {
      populate_samples = true
    }
  }
# tftest skip
```

ludoo:
I prefer the first one, but either works fine if you have a strong preference :)

fast/extras/0-cicd-github/main.tf
```diff
@@ -171,6 +177,11 @@ resource "github_repository_file" "default" {
       content,
     ]
   }
+
+  depends_on = [
```
ludoo, line 181: nit: can we use refs and avoid explicit dependencies? repository already references the resources in locals, branch can be set to use head ref from the branch resource maybe?

ludoo (approved):

fast/extras/0-cicd-github/main.tf
antonkovach: @ludoo, thank you for your suggestions. Based on your feedback, I refactored the code to avoid explicit dependencies by using resource references in the `github_branch` and `github_repository_pull_request` resources. The updated code now handles the different scenarios correctly. Please let me know if you have any further concerns or suggestions.

antonkovach: I would be happy to be added as a repo user

ludoo:
check your mail, you should then be able to merge yourself

PR #1243, opened 2023-03-13
Autopilot fixes

apichick:


blueprints/gke/autopilot/vpc.tf
```diff
@@ -37,6 +37,12 @@ module "vpc" {
   ]
 }
 
+module "firewall" {
```
juliocc, line 40: do you need this?

blueprints/gke/autopilot/variables.tf
```diff
@@ -81,4 +74,17 @@ variable "region" {
   description = "Region."
   type        = string
   default     = "europe-west1"
+}
+
+variable "vpc_create" {
```
juliocc, line 79: I'd merge `vpc_create` and `vpc_name`, maybe even use the existing `cluster_network_config`

juliocc (approved):

blueprints/gke/autopilot/vpc.tf
apichick: yes, to have firewall rule created for IAP

PR #1257, opened 2023-03-16
Fixes related to boot_disk in compute-vm module

apichick:


1
ludoo, line 1: can you remove this?

apichick: what?

PR #1268, opened 2023-03-20
simple-nva: add ability to parse BGP configs as strings.

LucaPrete:
Users can now pass FRR BGP configs as strings and not just as paths to existing files. This is useful if templatefile is used in the caller module.


simonebruzzechesse (approved):
Great improvement that will simplify usage of this module

ludoo (approved):

PR #1281, opened 2023-03-27
Use unique bundle name for Cloud Function

wiktorn:
When cloud-function module is used multiple times within project and default `bundle_config.output_path` is used then all the instances try to use filename and result is undefined without guarantee to converge to desired state (i.e. multiple functions may share the same code).

### DIFF
--- a/modules/cloud-function/main.tf
+++ b/modules/cloud-function/main.tf
@@ -267,7 +267,7 @@ resource "google_storage_bucket_object" "bundle" {
 data "archive_file" "bundle" {
   type             = "zip"
   source_dir       = var.bundle_config.source_dir
-  output_path      = var.bundle_config.output_path != null ? var.bundle_config.output_path : "/tmp/bundle-${var.project_id}-${var.name}.zip"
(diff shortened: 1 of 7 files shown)


modules/cloud-function/main.tf
```diff
@@ -267,8 +267,8 @@ resource "google_storage_bucket_object" "bundle" {
 data "archive_file" "bundle" {
   type             = "zip"
   source_dir       = var.bundle_config.source_dir
-  output_path      = var.bundle_config.output_path
-  output_file_mode = "0666"
+  output_path      = var.bundle_config.output_path != null ? var.bundle_config.output_path : "/tmp/bundle-${var.project_id}-${var.name}.zip"
```
juliocc, line 270: `coalesce(var.bundle_config.output_path, "/tmp/bundle-${var.project_id}-${var.name}.zip")`

juliocc (approved):

juliocc:
> When cloud-function module is used multiple times within project and default `bundle_config.output_path` is used then all the instances try to use filename and result is undefined without guarantee to converge to desired state (i.e. multiple functions may share the same code).

Can you add a test ensuring two calls to the module generate a different bundle path?

PR #1284, opened 2023-03-27
Update Provider and Terraform variables section

gcardamone:
Updating readme so that the provider and terraform variables section is identical to the documentation in the other stages.


ludoo (approved):
Thanks a lot for this!

Issue #1277, opened 2023-03-26
Split gke module


juliocc:
Create separate modules for GKE Standard and GKE Autopilot clusters

juliocc:
Duplicate of #1316

PR #1287, opened 2023-03-28
Add Backup for GKE plans support in GKE cluster module and removing deprecated control_plane field from GKE hub module

valeriobponza:
Incompatible change:
* moving `enable_addons.gke_backup_agent` to `backup_configs.enable_backup_agent` in `modules/gke-cluster`


modules/gke-cluster/variables.tf
```diff
@@ -14,6 +14,21 @@
...
+    enable_backup_agent = optional(bool, false)
+    backup_plans = optional(map(object({
+      region                            = string
+      schedule                          = string
+      retention_policy_days             = optional(string)
+      retention_policy_lock             = optional(bool, false)
+      retention_policy_delete_lock_days = optional(string)
+    })))
+  })
+  default = null
```
wiktorn, line 29: Can we have here `default = {}` and then we can skip `try(...)` part when accessing this variable because with `{}` optional defaults will kick in.

valeriobponza: make sense, working on it (also fixing test and pushing again)

valeriobponza:
thank you @juliocc for pointing out. I made this changes on my local module for a demo and I thought it was low effort to share them. No biggie if they gets overwritten. 

also, I would be happy to contribute to #1277 if there is still work to do!

wiktorn (approved):

PR #1289, opened 2023-03-29
Network Dashboard improvements and bug fixing

simonebruzzechesse:
- add explicit requirements for Python version at least 3.9
- add constraint on Shared VPC service projects and host project to both be in the monitoring scope
- fix issue with itertools.groupby which assumes the iterator to be already sorted when grouping by keys

The latter issue was causing duplicated metric to be written on Cloud Monitoring resulting in the following error:

_One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric_

### DIFF
--- a/blueprints/cloud-operations/network-dashboard/src/README.md
+++ b/blueprints/cloud-operations/network-dashboard/src/README.md
@@ -3,9 +3,9 @@
 This tool constitutes the discovery and data gathering side of the Network Dashboard, and can be used in combination with the related [Terraform deployment examples](../), or packaged in different ways including standalone manual use.
 
 - [Network Dashboard Discovery Tool](#network-dashboard-discovery-tool)
-- [Quick Usage Example](#quick-usage-example)
-- [High Level Architecture and Plugin Design](#high-level-architecture-and-plugin-design)
-- [Debugging and Troubleshooting](#debugging-and-troubleshooting)
+  - [Quick Usage Example](#quick-usage-example)
+  - [High Level Architecture and Plugin Design](#high-level-architecture-and-plugin-design)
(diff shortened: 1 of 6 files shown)


blueprints/cloud-operations/network-dashboard/src/README.md
```diff
@@ -2,9 +2,10 @@
 
 This tool constitutes the discovery and data gathering side of the Network Dashboard, and can be used in combination with the related [Terraform deployment examples](../), or packaged in different ways including standalone manual use.
 
-- [Quick Usage Example](#quick-usage-example)
-- [High Level Architecture and Plugin Design](#high-level-architecture-and-plugin-design)
-- [Debugging and Troubleshooting](#debugging-and-troubleshooting)
+- [Network Dashboard Discovery Tool](#network-dashboard-discovery-tool)
```
ludoo, line 5: This heading in the ToC and the ToC indents do not match the hierarchy in the doc

ludoo (approved):
I don't know what you're changing with the groupings, but I trust you :)

juliocc:
@simonebruzzechesse I'm trying to target Python 3.10 for everything in the repo. Would that make sense for you?

blueprints/cloud-operations/network-dashboard/src/README.md
simonebruzzechesse: fixed indentation!

simonebruzzechesse:
> @simonebruzzechesse I'm trying to target Python 3.10 for everything in the repo. Would that make sense for you?

Sure! Function code works with Python 3.10 (tested both locally and on CF), should I explicitly pin the required version in README or is it fine >= 3.10?

PR #1312, opened 2023-04-09
add firewall enforcement variable to VPC

fawzihmouda:
since we can use both [net-vpc-firewall-policy](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/net-vpc-firewall-policy) and [net-vpc-firewall](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/net-vpc-firewall) at the same time in a VPC, this PR will add the ability to choose the enforcement order for the Firewalls Policy

Default enforcement order is ```AFTER_CLASSIC_FIREWALL```  https://cloud.google.com/vpc/docs/firewall-policies-overview#rule-evaluation


ludoo:
I don't think there's a way yet to control this via Terraform.

fawzihmouda:
yes its available in [version 4.60](https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#4600-april-4-2023), but for some reason the checks are breaking this is why i closed the PR

modules/net-vpc/main.tf
```diff
@@ -34,14 +34,16 @@ data "google_compute_network" "network" {
...
+  count                                     = var.vpc_create ? 1 : 0
+  project                                   = var.project_id
+  name                                      = var.name
+  description                               = var.description
+  auto_create_subnetworks                   = var.auto_create_subnetworks
+  delete_default_routes_on_create           = var.delete_default_routes_on_create
+  mtu                                       = var.mtu
+  routing_mode                              = var.routing_mode
+  network_firewall_policy_enforcement_order = var.firewall_enforcement_order
+
```
juliocc, line 46: remove empty line before `}`

juliocc (changes requested):
* The provider is currently pinned to version 4.59. Can you update [default-versions.tf](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/default-versions.tf) and then run `find . -name versions.tf -exec cp default-versions.tf {} \;` from the root of the repo to upgrade it?
* Please add at least [one test](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#testing-via-readmemd-example-blocks) to ensure the new variable is working correctly.
* I also left a couple of comments in variables.tf and main.tf. Please take a look.

juliocc:
Thanks for the PR and applying the requested changes @fawzihmouda

PR #1315, opened 2023-04-11
Fix stage links script for multitenant stages

ludoo:


fast/stage-links.sh
```diff
@@ -70,18 +70,34 @@ case $STAGE_NAME in
...
-  PROVIDER="providers/1-resman-tenant-providers.tf"
-  TFVARS="tfvars/0-bootstrap-tenant.auto.tfvars.json"
+  PROVIDER="tenants/$TENANT/providers/1-resman-tenant-providers.tf"
+  TFVARS="tenants/$TENANT/tfvars/0-bootstrap-tenant.auto.tfvars.json"
   ;;
 "2-networking"*)
-  PROVIDER="providers/2-networking-providers.tf"
-  TFVARS="tfvars/0-bootstrap.auto.tfvars.json
-  tfvars/1-resman.auto.tfvars.json"
+  if [[ -z "$TENANT" ]]; then
```
juliocc, line 77: can you use `if [[ ! -z ${TENAT+x} ]]; then` for those using OSX's old bash?

juliocc (changes requested):

fast/stage-links.sh
ludoo: Well, I don't have osx old or new. I just kept using the syntax we already have. Feel free to fix for unfortunate users later
 😂

PR #1326, opened 2023-04-14
Remove net-interconnect-attachment-direct

juliocc:


ludoo (approved):
Cam you also check the readmes?

juliocc:
I didn't find any references. Are you referring to one in particular?

ludoo:
Thought we were linking to it, probably we already planned to eventually
nuke it 😂

PR #1329, opened 2023-04-17
fix: Change net-glb serve_while_stale type to number

tobbbles:
In the provider, the `serve_while_stale` attribute is listed as an integer, https://github.com/hashicorp/terraform-provider-google/blob/main/google/resource_compute_backend_service.go#L373-L374

And also documented here as such: https://cloud.google.com/cdn/docs/serving-stale-content

Currently it is not possible to configure `cdn_policy` with this module and specify a `serve_whilst_stale` value.


juliocc (approved):

juliocc:
thanks @tobbbles! good catch.

PR #1331, opened 2023-04-17
IAP for Cloud Run GA

### LINKED ISSUE #1321: Error replicating Cloud Run Explore Use Case 5: IAP for Cloud Run app is not set up correctly
I have replicated the example put forward in the [Cloud Run Blueprint Use Case 5](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/blueprints/serverless/cloud-run-explore/README.md#use-case-5-cloud-run-service-protected-by-cloud-armor-and-identity-aware-proxy). 

- I have successfully applied the Terraform in my cloud environment;
- Can see all of the infrastructure that has been configured;
- The custom domain is available; and
- Redirects me to the Google login page

However, after logging in, I see the following text in my browser:

```
IAP for Cloud Run app is not set up correctly.Please follow the instructions to rectify IAP and Cloud Run set up: https://cloud.google.com/iap/docs/enabling-cloud-run
```

Another strange thing that I saw was that the OAuth consent screen did not contain any of the custom information that I specified (App name, support email etc.), but simply says _Choose an account to continue to iap.googleapis.com_.

juliodiez:
Now that IAP for Cloud Run is GA, the IAP service agent needs to be provisioned and granted access to invoke Cloud Run.
Fixes #1321

### DIFF
--- a/blueprints/serverless/cloud-run-explore/main.tf
+++ b/blueprints/serverless/cloud-run-explore/main.tf
@@ -49,7 +49,10 @@ module "cloud_run" {
     }
   }
   iam = {
-    "roles/run.invoker" = ["allUsers"]
+    "roles/run.invoker" = (local.gclb_create && var.iap.enabled
+      ? ["serviceAccount:${google_project_service_identity.iap_sa[0].email}"]
+      : ["allUsers"]
+    )
   }
   ingress_settings = var.ingress_settings
 }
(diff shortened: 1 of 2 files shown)


blueprints/serverless/cloud-run-explore/main.tf
```diff
@@ -49,7 +49,10 @@ module "cloud_run" {
     }
   }
   iam = {
-    "roles/run.invoker" = ["allUsers"]
+    "roles/run.invoker" = (local.gclb_create && var.iap.enabled
+      ? ["serviceAccount:${google_project_service_identity.iap_sa[0].email}"]
```
ludoo, line 53: did you try this branch? Terraform has the nasty habit of checking both branches regardless of condition, and this will fail if the resource does not exist; it got better lately, but I would wrap this into a `try/null` just to be extra safe

ludoo (approved):

blueprints/serverless/cloud-run-explore/main.tf
juliodiez: I tried because I was not sure of the terraform behavior (I expected it to be the opposite :/ ).  But now I realize I didn't check it correctly. Thanks, fixed.

PR #1336, opened 2023-04-18
Certificate renewal through terraform 

bjohnrl:
When renewing the certificate by updating the variable file, GCP will not let you update because it will try to delete the old certificate is been used by the target proxy. Updating lifecycle  with create_before_destroy will allow the certificate and attach to the proxy and then delete the old certificate.


juliocc (approved):

juliocc:
hi @bjohnrl, thanks the patch. Can you run your code through `terraform fmt` and sign the CLA?

Thanks!

juliocc:
The linting error is not related to your PR. I've force-merged and we'll fix the issue later.

Thanks @bjohnrl!

PR #1339, opened 2023-04-21
Deprecate plan runner fixture and all its variants

juliocc:
Migrate remaining modules to the new testing framework


fast/stages-multitenant/0-bootstrap-tenant/README.md
```diff
@@ -85,7 +85,7 @@ As shown in the script output above, the provider file is a template used as a s
 
 Note that the `outputs_location` variable is disabled by default, you need to explicitly set it in your `terraform.tfvars` file if you want output files to be generated by this stage. This is a sample `terraform.tfvars` that configures it, refer to the [org-level bootstrap stage documentation](../../stages/0-bootstrap/README.md#output-files-and-cross-stage-variables) for more details:
 
-```hcl
+```tfvars
```
ludoo, line 88: weird, I did not know tfvars was supported; it's still hcl though, any benefit in changing this?

ludoo (commented):
Deprecating the old runners is great, I'm just not sure we want to add inventory for everything: whenever the resource changes they break, and often you end up needing to remove stuff fromt he inventory, or figure out why some attribute changed. Happened twice to me lately and I lost a lot of time over something that was WAI.

fast/stages-multitenant/0-bootstrap-tenant/README.md
juliocc: It's mostly cosmetic: you don't have to add `tftest skip` to prevent that snippet from being parsed as a test

juliocc:
> Deprecating the old runners is great, I'm just not sure we want to add inventory for everything: whenever the resource changes they break, and often you end up needing to remove stuff fromt he inventory, or figure out why some attribute changed. Happened twice to me lately and I lost a lot of time over something that was WAI.

Agreed. Most of the new inventories reflect what we were already testing using python assertion. I did add a few here and there but I tried to keep them simple and asserting only the critical bits.

We should definitely fix/remove/improve those cases where you find things WAI that are causing false positives.

ludoo (approved):

PR #1340, opened 2023-04-22
Extend tests to use lockfile if available

juliocc:
This commints allows the test framework to use a lockfile generated within the tools/lockfile directory. This allows reusing the provider cache and also running tests in parallel.


.github/workflows/tests.yml
```diff
@@ -29,10 +29,55 @@ env:
   PYTHON_VERSION: "3.10"
   TF_PLUGIN_CACHE_DIR: "/home/runner/.terraform.d/plugin-cache"
   TF_VERSION: 1.4.4
+  TFTEST_COPY: 1
 
 jobs:
+  setup-tf-providers:
+    runs-on: ubuntu-latest
```
ludoo, line 36: can't you put this in the composite action? there's also a duplicate step

juliocc:
Yeah, I'm still trying to figure out the details and I'll clean it up once
it's working.

.github/workflows/tests.yml
juliocc: Not really, we need the other jobs to run after we compute, fetch and cache all the providers.

PR #1341, opened 2023-04-26
Removing project_id conditional from gke-cluster modules

mark1000:
Removing the project_id conditional and getting it directly from the variable.


juliocc:
@ludoo we think the project_id in the `peering_config` parameter is unnecessary since it's always going to be the project of the cluster. 

Do you remember why we had it as an argument to the gke module?

ludoo:
> @ludoo we think the project_id in the `peering_config` parameter is unnecessary since it's always going to be the project of the cluster.
> 
> Do you remember why we had it as an argument to the gke module?

If you are using Shared VPC it's the host project id AFAIK.

juliocc:
> If you are using Shared VPC it's the host project id AFAIK.

Ah of course. I'll bring back the default and provide a more reasonable default.

mark1000:
Ah, I didn't think about Shared VPC. Thanks!

ludoo:
Closing this, feel free to reopen if anything needs changing.

PR #1347, opened 2023-05-01
Fix external documentation links

bobidle:
I noticed several outdated external documentation links.

Updated them to the new redirect location. If the new location was missing the old information i tried to find the new location and anchor somehow (checked the old locations with https://archive.org/)

Please let me know if i could do something better with my changes and all.

There are still some more external links to fix, but i wanted to push my changes for now, i am not sure when i have time to continue updating the docs. Also i hope my changes are not outdated already.


ludoo:
Many thanks for this!

ludoo (approved):

PR #1349, opened 2023-05-02
Enhance GKE Backup Configuration Support

tacchino:
Adds support for more detailed GKE backup config

* non default include_volume_data
* non default include_secrets
* encryption_key
* selected_namespaces


modules/gke-cluster-standard/main.tf
```diff
@@ -372,11 +372,25 @@ resource "google_gke_backup_backup_plan" "backup_plan" {
...
+    dynamic "encryption_key" {
+      for_each = lookup(each.value, "encryption_key", null) != null ? [""] : []
+      content {
+        gcp_kms_encryption_key = each.value.encryption_key
+      }
+    }
+
+    all_namespaces = lookup(each.value, "namespaces", null) != null ? null : true
+    dynamic "selected_namespaces" {
+      for_each = lookup(each.value, "namespaces", null) != null ? [""] : []
```
ludoo, line 389: if you default to `[]` in the variable you can then omit `'lookup`

```hcl
namespaces                        = optional(list(string), [])
```

ludoo (approved):
you need to rerun tfdoc, linting check probably fails there :)

thanks for this, dropped a couple minor comments

tacchino:
Awesome, thanks for the the fast review @ludoo

PR #1354, opened 2023-05-03
adding the ability to export or not the subnet routes with public ip

itManuel:
I added the ability to avoid the exchange subnet routes with public IPv4, adding a variable. The default value will allow the module to work as until now (the default value is TRUE so if you want to avoid that you should change it to FALSE)

Regards,


modules/net-vpc-peering/variables.tf
```diff
@@ -27,6 +27,12 @@ variable "export_peer_custom_routes" {
   default     = false
 }
 
+variable "export_subnet_routes_with_public_ip" {
```
ludoo, line 30: can we rename to `export_public_ip_routes` for consistency?

ludoo (changes requested):
Dropped a minor comment, can you make sure CLA and checks pass? If you are unsure on how to do that just ping us here. Thanks for this!

modules/net-vpc-peering/variables.tf
itManuel: Sure, please check again

itManuel: I singned the CLA, which checks are you talking about?

ludoo:
hmmm
image

ludoo:
if both users are you, ping me internally :)

itManuel:
can you cancel this? I'll try to do it again BUT without multi-personality issues.

PR #1360, opened 2023-05-05
Add support for Shared VPC in Cloud Run

### LINKED ISSUE #1350: Feat req: Add support for Shared VPC in Cloud Run
Cloud Run may use a VPC access connector for egress. This connector uses either a special purpose subnet or an unused custom IP range. When in a Shared VPC the only option is to define a subnet, however this module only offers to create it defining the IP range.

juliodiez:
A VPC access connector in a Shared VPC needs to specify the subnet and subnet project (the host project).
Fixes #1350


ludoo (approved):
👍

PR #1365, opened 2023-05-08
feat(net-cloudnat): add toggle for independent endpoint mapping

JSchwerberg:


juliocc (dismissed):
Thanks for this PR.

Can you update the README using tfdoc as explained in the contribution guide?

JSchwerberg:
@juliocc Done :)

JSchwerberg:
@juliocc Actually, do not merge this yet, this will break a lot of configurations.  Let me fix first.

JSchwerberg:
@juliocc slipping in some toggles for dynamic port allocation as well -- the current module is broken, with `var.config_min_ports_per_vm` being a no-op since dynamic port allocation isn't turned on by default.  This fixes that, and adds a variable for `config_max_ports_per_vm`, as well as making sure that dynamic port allocation and independent endpoint mapping are mutually exclusive as required by the API.

modules/net-cloudnat/variables.tf
```diff
@@ -24,6 +24,22 @@ variable "config_min_ports_per_vm" {
   description = "Minimum number of ports allocated to a VM from this NAT config."
   type        = number
   default     = 64
+
+  validation {
+    condition = var.enable_dynamic_port_allocation
```
ludoo, line 29: I don't think you can check other variables from a validation condition, unless things changed recently... What about bundling both into a single variable of type object?

modules/net-cloudnat/variables.tf
```diff
@@ -102,3 +108,15 @@ variable "subnetworks" {
...
 }
+
+variable "enable_endpoint_independent_mapping" {
+  description = "Enables endpoint independent mapping if null"
+  type        = bool
+  default     = null
+}
+
+variable "enable_dynamic_port_allocation" {
+  description = "Disables dynamic port allocation if null"
```
juliocc: variable name says "enable", description days "disable". Which one is it?

modules/net-cloudnat/outputs.tf
```diff
@@ -42,3 +42,12 @@ output "router_name" {
   description = "Cloud NAT router name."
   value       = local.router_name
 }
+
+output "validate_config" {
+  value = null
+
+  precondition {
+    condition     = var.enable_dynamic_port_allocation ? var.enable_endpoint_independent_mapping == false : true
```
juliocc: Never seen this before. Does this work? In any case seems like abusing the language for something you can do with a proper "config" variable with all the required flags.

modules/net-cloudnat/variables.tf
```diff
@@ -102,3 +108,15 @@ variable "subnetworks" {
   }))
   default = []
 }
+
+variable "enable_endpoint_independent_mapping" {
+  description = "Enables endpoint independent mapping if null"
```
juliocc: double check description, "enables if null"?

juliocc (changes requested):
See inline comments, also make sure all tests and linting checks are passing.

modules/net-cloudnat/variables.tf
```diff
@@ -20,10 +20,25 @@ variable "addresses" {
...
 }
 
-variable "config_min_ports_per_vm" {
-  description = "Minimum number of ports allocated to a VM from this NAT config."
-  type        = number
-  default     = 64
+variable "port_allocation_config" {
+  description = "Configuration for how to assign ports to virtual machines. min_ports_per_vm and max_ports_per_vm have no effect unless enable_dynamic_port_allocation is set to 'true.'"
+  type = object({
+    enable_endpoint_independent_mapping = bool
```
juliocc: I'd make all these option with the default specified per field

modules/net-cloudnat/variables.tf
JSchwerberg: If `null` it's disabled, but if `true` it's enabled -- this is a simple passthrough of the Terraform resource [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_router_nat#enable_dynamic_port_allocation), and mirrors its behavior (enable_endpoint_independent_mapping defaults to true, and the two are mutually exclusive, meaning that enable_dynamic_port_allocation defaults to false).  

This is a non-issue in the updated PR with a config block.

JSchwerberg:
> See inline comments, also make sure all tests and linting checks are passing.

I'd love to, but that would mean Github would have to fix their 6 hour plus actions outage, so we'll have to re-run whenever they get their act together :grin:

modules/net-cloudnat/variables.tf
```diff
@@ -20,10 +20,25 @@ variable "addresses" {
...
+  })
+  default = {
+    enable_endpoint_independent_mapping = true
+    enable_dynamic_port_allocation      = false
+    min_ports_per_vm                    = 64
+    max_ports_per_vm                    = 65536
+  }
+
+  validation {
+    condition     = var.port_allocation_config["enable_dynamic_port_allocation"] ? var.port_allocation_config["enable_endpoint_independent_mapping"] == false : true
```
juliocc: use dotted notation (`var.port_allocation_config.enable_dynamic_port_allocation`)

modules/net-cloudnat/outputs.tf
JSchwerberg: It does work, and in fact is suggested by Terraform devs as a workaround to cross-variable validation not working. But yes, refactored into a config block.

modules/net-cloudnat/variables.tf
```diff
@@ -20,10 +20,25 @@ variable "addresses" {
...
+  type = object({
+    enable_endpoint_independent_mapping = bool
+    enable_dynamic_port_allocation      = bool
+    min_ports_per_vm                    = number
+    max_ports_per_vm                    = number
+  })
+  default = {
+    enable_endpoint_independent_mapping = true
+    enable_dynamic_port_allocation      = false
+    min_ports_per_vm                    = 64
```
JSchwerberg, line 34: I'm not sure if we want to keep this minimum, as the provider will default to 32? 

I've kept it here for the sake of not boiling the ocean, but setting this (and `max_ports_per_vm`) to null would allow us to validate that these are not being set when `enable_dynamic_port_allocation` is false.

JSchwerberg:
This is now passing linting and tests. Some test are failing on my machine locally, but are ENTIRELY unrelated to this PR and have to do with upstream changes in the firewall API, afaict.

cc @juliocc @ludoo

juliocc (approved):

PR #1371, opened 2023-05-12
added ability to lookup projects by folders or project ids

alx13:


fast/stages/2-security/vpc-sc.tf
```diff
@@ -19,38 +19,117 @@ locals {
...
+      from = v.from
+      to = merge([
+        v.to,
+        {
+          resources = sort(toset(flatten([
+            v.to.resources,
+            [for f in v.to.folder_ids : local._vpc_sc_resources_projects_in_folders[f]],
+            [for f in v.to.project_ids : local._vpc_sc_resources_projects_by_ids[f]],
+          ])))
+          project_ids = null
```
juliocc, line 87: I think this is ignored by the [provider](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/access_context_manager_service_perimeter). Looks lit it only accepts resources in the form of projects/number

alx13: Yes, I nulled it in case of underlying module changes

fast/stages/2-security/README.md
```diff
@@ -14,21 +14,30 @@ The following diagram illustrates the high-level design of created resources and
...
+    - [Running the stage](#running-the-stage)
+  - [Customizations](#customizations)
+    - [KMS keys](#kms-keys)
+    - [VPC Service Controls configuration](#vpc-service-controls-configuration)
+      - [Dry-run vs. enforced](#dry-run-vs-enforced)
+      - [Access levels](#access-levels)
+      - [Lookup variables](#lookup-variables)
+      - [Ingress and Egress policies](#ingress-and-egress-policies)
+      - [Perimeters](#perimeters)
+    - [Brdiges](#brdiges)
```
ludoo, line 36: Typo (not yours probably but worth fixing)

fast/stages/2-security/README.md
```diff
@@ -185,12 +194,13 @@ The script will create one keyring for each specified location and keys on each
 
 A set of variables allows configuring the VPC SC perimeters described above:
 
-- `vpc_sc_perimeter_projects` configures project membership in the three regular perimeters
+- `vpc_sc_perimeter_projects` configures project membership in regular perimeters
+- `vpc_sc_bridges` configures bridges between perimters
```
ludoo, line 198: typo

fast/stages/2-security/README.md
```diff
@@ -258,24 +288,55 @@ Access levels and egress/ingress policies are referenced in perimeters via keys.
...
+    resources            = ["projects/0000000000"]
+    resources            = ["projects/2222222222"]
+    project_ids          = ["xxx-project-id"]
+    folder_ids           = ["555555555"]
+    folder_ids_recursive = ["666666666"]
   }
 }
 ```
 
+### Brdiges
```
ludoo, line 319: typo

ludoo (approved):

ludoo:
@juliocc @drebes this will disrupt our standard VPC SC design, by making the stage more flexible. While I see the benefits, we will need to change all the supporting materials before a release is cut, including workflow slides.

ludoo:
Won't the diagram and architecture also need to change, as the variables do not enforce a standard design anymore?

juliocc:
I can't pinpoint a specific issue with this PR but I don't like the implementation. I **do like** the idea behind it but the code just looks messy to me. 

I particularly dislike the new `vpc-sc-resources.tf` but it could be that I can't follow what's  going on.

ludoo:
Agree, can we move the code to its own module, which can be used to resolve ids to names?

alx13:
`vpc-sc-resources.tf` is doing three things:

1. it puts all references to project by ids, folders to fetch projects inside recursive and non-recursive to the respective set, to deduplicate API lookups
2. creates lookup tables: `project id => project number`, `folder id => projects number` (recursive and not)
3. running lookups to form project lists for different inputs of VPC SC

ludoo:
> Agree, can we move the code to its own module, which can be used to resolve ids to names?

alx13:
I can, but let's get @juliocc onboard with this.

ludoo:
> I can, but let's get @juliocc onboard with this.

we're usually very much aligned...

rosmo:
I think the code looks fine, for what's it worth. "I don't like it" doesn't sound good.

ludoo:
> I think the code looks fine, for what's it worth. "I don't like it" doesn't sound good.

it does instead, if the good code makes the whole stage harder to understand or modiify; anyway, I think Julio and I are aligned, let me close this and feel free to reopen as a standalone module, which I would very much like to have in Fabric...

PR #1375, opened 2023-05-14
Several updates

bluPhy:
Several updates


juliocc (approved):
Thanks for all these fixes! 

Can you ensure the linting test passes? I think you're modifying some of the input/output tables in the READMEs without updating the corresponding terraform file.

juliocc (approved):
Thanks for all these fixes! 

Can you ensure the linting test passes? I think you're modifying some of the input/output tables in the READMEs without updating the corresponding terraform file.

bluPhy:
@juliocc Spend some time trying to understand why the linting fails on just a few files where there is no changes to the variable names but in the comments or descriptions, at the end I could not get to the bottom of it. Decided to revert changes only to the problematic files to allow most of the changes to be committed and will work only on the problematic ones later.

juliocc:
@bluPhy I'm merging this. Can you open a PR with the remaining changes and I'll try to fix the linting issues.

PR #1376, opened 2023-05-15
fixed permissions for security stage SA

alx13:
it should be able to use automation project
as a quota project, hence it needs `serviceusage.serviceUsageConsumer` role


ludoo:
Thanks!!!

PR #1385, opened 2023-05-21
Add conditional org admin role to sandbox SA

ludoo:


juliocc (approved):

gustavovalverde:
I'm having an error after this change, when doing a `terraform apply` on `1-resman`

```
│ Error: Invalid index
│ 
│   on branch-sandbox.tf line 68, in resource "google_organization_iam_member" "org_policy_admin_sandbox":
│   68:   member = module.branch-sandbox-sa.0.iam_email
│     ├────────────────
│     │ module.branch-sandbox-sa is empty tuple
│ 
```

PR #1386, opened 2023-05-22
Support CMEK encryption in logging-bucket module

afda16:
We have a use case, Local Controls, that requires to configure CMEK with Logging buckets. This commit adds an optional variable to configure CMEK in the `logging-bucket` module. By default the Logging bucket won't use CMEK encryption.

To configure CMEK for Logging buckets it's also required to add the correct permissions to the bucket service account. For more information and a Terraform example, see:

https://cloud.google.com/logging/docs/routing/managed-encryption-storage

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/logging_project_bucket_config#example-usage

### DIFF
--- a/modules/logging-bucket/main.tf
+++ b/modules/logging-bucket/main.tf
@@ -21,6 +21,13 @@ resource "google_logging_project_bucket_config" "bucket" {
   retention_days = var.retention
   bucket_id      = var.id
   description    = var.description
+
+  dynamic "cmek_settings" {
+    for_each = var.kms_key_name == null ? [] : [""]
+    content {
+      kms_key_name = var.kms_key_name
+    }
+  }
 }
 
 resource "google_logging_folder_bucket_config" "bucket" {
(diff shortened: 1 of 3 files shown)


modules/logging-bucket/main.tf
```diff
@@ -21,6 +21,13 @@ resource "google_logging_project_bucket_config" "bucket" {
   retention_days = var.retention
   bucket_id      = var.id
   description    = var.description
+
+  dynamic "cmek_settings" {
+    for_each = var.kms_key_name == null ? {} : { kms_key_name = var.kms_key_name }
```
ludoo, line 26: you can also do something like

```hcl
for_each = var.kms_key_name == null ? [] : [""]
content {
  kms_key_name = var.kms_key_name
}
```

Which is easier to read

ludoo (approved):

modules/logging-bucket/main.tf
afda16: updated

PR #1391, opened 2023-05-23
fix(stages): only add sandbox SA when `sandbox` feature is enabled

gustavovalverde:
If you have the `project_factory` feature enabled, but not the `sandbox` feature (as it's not a requirement on your org), when doing a `terraform apply` on `1-resman` it raises this errors as it's expecting the wrong feature when creating the sandbox SA

```
│ Error: Invalid index
│ 
│   on branch-sandbox.tf line 68, in resource "google_organization_iam_member" "org_policy_admin_sandbox":
│   68:   member = module.branch-sandbox-sa.0.iam_email
│     ├────────────────
│     │ module.branch-sandbox-sa is empty tuple
│ 
```

### DIFF
--- a/fast/stages/1-resman/branch-sandbox.tf
+++ b/fast/stages/1-resman/branch-sandbox.tf
@@ -62,7 +62,7 @@ module "branch-sandbox-sa" {
 }
 
 resource "google_organization_iam_member" "org_policy_admin_sandbox" {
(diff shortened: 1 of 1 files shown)


gustavovalverde:
CLA signed

ludoo (approved):
Thanks for this, stupid mistake on our side :)

PR #1393, opened 2023-05-24
Update README.md

juliocc:


modules/cloudsql-instance/README.md
```diff
@@ -27,7 +27,6 @@ module "vpc" {
   name       = "my-network"
   psa_config = {
     ranges = { cloud-sql = "10.60.0.0/16" }
-    routes = null
```
ludoo, line 30: should we make that optional? isn't it already?

juliocc: `routes` is wrong we already have two optional arguments now: https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/net-vpc/variables.tf#L93

ludoo: god I'm blind, you're removing it

ludoo (approved):

PR #1395, opened 2023-05-24
allow to configure stack type in GKE autopilot

NitriKx:


ludoo:
Thanks for this! Would you mind also doing the same for the standard GKE module?

ludoo:
Thanks a lot! setting this to auto-merge :)

ludoo (approved):

PR #1400, opened 2023-05-26
Add default googleapi route creation to net-vpc

juliocc:


modules/net-vpc/routes.tf
```diff
@@ -88,3 +88,43 @@ resource "google_compute_route" "vpn_tunnel" {
   tags                = each.value.tags
   next_hop_vpn_tunnel = each.value.next_hop
 }
+
+resource "google_compute_route" "private" {
```
ludoo, line 92: an we leverage the existing route resource, by defining a local with the private ones, and merging with the routes variable?

ludoo (approved):

modules/net-vpc/routes.tf
```diff
@@ -17,7 +17,42 @@
 # tfdoc:file:description Route resources.
 
 locals {
-  _routes = var.routes == null ? {} : var.routes
+  _default_routes = {
```
ludoo, line 20: you can simplify with a for_each maybe? only names and ranges change

drebes:
All good just maybe change the var name? These are not default routes (in the networking sense).

drebes:
One more point, a bit more controversial. I see today we don't default to true delete_default_routes_on_create. We could consider also adding a TF managed "default" route in addition to these in the module (create_special_routes), or at least do that just for FAST.

juliocc:
> One more point, a bit more controversial. I see today we don't default to true delete_default_routes_on_create. We could consider also adding a TF managed "default" route in addition to these in the module (create_special_routes), or at least do that just for FAST.

Good point. Can you open an issue for this?

juliocc:
> All good just maybe change the var name? These are not default routes (in the networking sense).

You right, "default routes" is a terrible name. Any suggestions?

PR #1404, opened 2023-05-30
Add trigger SA for Cloud Run

wiktorn:
Allows setting service account for triggers in Cloud Run

### DIFF
--- a/modules/cloud-run/README.md
+++ b/modules/cloud-run/README.md
@@ -218,6 +218,57 @@ module "cloud_run" {
 # tftest modules=1 resources=2 inventory=audit-logs.yaml
 ```
 
+#### Using custom service accounts for triggers
+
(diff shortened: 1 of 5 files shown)


juliocc (approved):
Great PR! Thanks @wiktorn

modules/cloud-run/README.md
```diff
@@ -218,6 +218,57 @@ module "cloud_run" {
 # tftest modules=1 resources=2 inventory=audit-logs.yaml
 ```
 
+#### Using custom service accounts for triggers
```
juliocc, line 221: Can you update the TOC at the beginning of the README?

wiktorn: TOC is only for `###`  headers, so no need to update TOC unless we want to expand it and add 4th level headers.

PR #1406, opened 2023-05-31
moved NVA landing subnets from YAMLs

alx13:


fast/stages/2-networking-c-nva/main.tf
```diff
@@ -18,14 +18,7 @@
...
   custom_roles = coalesce(var.custom_roles, {})
-  # combine all regions from variables and subnets
-  regions = distinct(concat(
-    values(var.regions),
-    values(module.dev-spoke-vpc.subnet_regions),
-    values(module.landing-trusted-vpc.subnet_regions),
-    values(module.landing-untrusted-vpc.subnet_regions),
-    values(module.prod-spoke-vpc.subnet_regions),
-  ))
+  regions      = distinct(values(var.regions))
```
ludoo, line 21: this will fail if subnets from the factory use different regions which are not in variables, can you revert this change?

fast/stages/2-networking-e-nva-bgp/main.tf
```diff
@@ -19,13 +19,7 @@
...
   custom_roles = coalesce(var.custom_roles, {})
   # combine all regions from variables and subnets
-  regions = distinct(concat(
-    values(var.regions),
-    values(module.dev-spoke-vpc.subnet_regions),
-    values(module.landing-trusted-vpc.subnet_regions),
-    values(module.landing-untrusted-vpc.subnet_regions),
-    values(module.prod-spoke-vpc.subnet_regions),
-  ))
+  regions = distinct(values(var.regions))
```
ludoo, line 22: same as previous comment

fast/stages/2-networking-e-nva-bgp/nva.tf
```diff
@@ -58,18 +58,21 @@ locals {
       gcp_prod_primary                = var.gcp_ranges.gcp_prod_primary
       gcp_prod_secondary              = var.gcp_ranges.gcp_prod_secondary
       # The IPs of cross-region NVA VMs in the untrusted VPC (x.y.w.z)
-      ip_neighbor_cross_region_nva_0 = cidrhost(module.landing-untrusted-vpc.subnet_ips["${local._regions_cross[v.0]}/landing-untrusted-default-${local.region_shortnames[local._regions_cross[v.0]]}"], 101)
-      ip_neighbor_cross_region_nva_1 = cidrhost(module.landing-untrusted-vpc.subnet_ips["${local._regions_cross[v.0]}/landing-untrusted-default-${local.region_shortnames[local._regions_cross[v.0]]}"], 102)
+      ip_neighbor_cross_region_nva_0 = cidrhost(module.landing-untrusted-vpc.subnet_ips[local.landing_subnet_ids[local._regions_cross[v.0]].untrusted], 101)
```
ludoo, line 61: can you break long lines, here and in the other places?

ludoo (changes requested):
Alex, thanks for doing this. Dropped a couple minor comments but this looks ok to me once those are addressed.

fast/stages/2-networking-c-nva/main.tf
alx13: It would not fail, because:
1. `local.regions` is only used in [`local._region_tokens`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/fast/stages/2-networking-c-nva/regions.tf#L29)
2. `local._region_tokens` is only used in [`local.region_shortnames`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/fast/stages/2-networking-c-nva/regions.tf#LL32C3-L32C20)
3. and `local.region_shortnames` is only used to lookup regions which are in var.regions: [proof](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fcloud-foundation-fabric+path%3Afast%2Fstages%2F2-networking-c-nva+local.region_shortnames&type=code)

ludoo: it's nice to be proven wrong on our own code :) thanks for being thorough and persistent!

ludoo: Actually, I'm right: customer adds subnets in new regions via factory. New NAT gateways are created in the new regions. Poof!

[proof](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/fast/stages/2-networking-a-peering/spoke-dev.tf#L76)

alx13: This is different design, it doesn't involve creating required subnets, because we are not putting NVA NICs in them

alx13: We can avoid the cycle just having separate LUT which will only use `var.regions` as source and use it in required subnet creation

ludoo: what is a LUT? anyway, this is becoming too complex: there's no need to break everything in pieces to move 3 subnets in the code; I'm closing this, @juliocc feel free to reopen ofc if you disagree

alx13: By LUT a meant lookup table, but in our case it's `map` `local.region_shortnames`
We can just have another map, which will not have outputs as dependencies. But again, we are not using `local.region_shortnames` to lookup anything which comes from factories.

Referring to something which is outside of terraform code by file naming conventions is actually making this thing too complex and brings a lot of unnecessary copy-paste of `"${each.value.region}/landing-untrusted-default-${each.value.shortname}"`.

ludoo: well, FAST has worked great for the past year and a half for a lot of customers, partners, and PSO people; I'm sorry you're finding it too complex but I don't think we need to start dismembering it

alx13: I'm saying that in this particular case it's unnecessary tight coupling, which makes code hard to refactor and adapt.

If you want to create a table with short names for regions, why not to use [`google_compute_regions`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_regions) data source, which does not have any dependencies except for automation project.

alx13: You can mock them for testing without credentials.

alx13: And regarding coupling on the file name to contain short name for the region - it is even not mentioned in the documentation how to derive that short name. So “Poof” in NAT can happen if someone will just put a wrong file name.

alx13: And one thing:
having `var.regions` and `module.landing-trusted-vpc.subnet_regions` in the same set doesn't make sense, because `module.landing-trusted-vpc.subnet_regions` includes `var.regions` by design.

PR #1412, opened 2023-06-02
Add VPN monitoring alerts to 2-networking and VPN usage chart

afda16:
The Fast stage 2-networking-* currently adds a monitoring dashboard for VPN metrics. This change adds an additional chart to monitor the usage of the VPN bandwidth.
    
This change also adds the following monitoring alerts:
    
* VPN tunnel established
* [VPN bandwidth](https://cloud.google.com/network-connectivity/docs/vpn/how-to/viewing-logs-metrics#define-bandwidth-alerts)
    
To configure the alerts, there is a new `alert_config` variable with defined default values.
    
The alerts are created in the stage `b` by default. In the stages a, c, d, and e, the alerts are created if the user creates the On-prem VPN.
    
To disable the creation of alerts, add the following to `terraform.tfvars`:
    
```
alert_config = {
  vpn_tunnel_established = null
  vpn_tunnel_bandwidth = null
}
```


fast/stages/2-networking-b-vpn/variables.tf
```diff
@@ -199,3 +199,25 @@ variable "vpn_onprem_primary_config" {
...
+variable "alert_config" {
+  description = "Configuration for monitoring alerts."
+  type = object({
+    vpn_tunnel_established = optional(object({
+      auto_close            = optional(string, null)
+      duration              = optional(string, "120s")
+      enabled               = optional(bool, true)
+      notification_channels = optional(list(string), [])
+      user_labels           = optional(map(string), {})
+    }), null)
```
ludoo, line 212: you don't need `null` as a default, it's implied :)

ludoo (approved):
Super nice, thanks Ana! The problem with networking stages is, this should be rolled out to all other stages too :(

fast/stages/2-networking-b-vpn/variables.tf
afda16: Thanks, updated

afda16:
Thanks @juliocc , I only copied the variables.tf but obviously it doesn't make sense to have a variable present if the implementation is not there. 

I can add the same monitoring to the other stages, but currently the dashboards are not replicated as well. Does it make sense to have monitoring (dashboards + alerts) copied to all the networking stages, or should it be a separate module?

aurelienlegrand (approved):
LGTM
I was able to test the VPN.json dashboard and it works fine. I wasn't able to test the full solution in a fabric fast deployment though because I don't have one at hand right now.

PR #1415, opened 2023-06-03
Add notice to net-ilb module on routes

ludoo:
This adds a notice to the README file on how to prevent a circular dependency, when using the module to create ILBaNH routes in the VPC module..


juliocc (approved):
Thanks for the changes :)

PR #1417, opened 2023-06-05
Remove hardcoded description from instance groups created under net-ilb

LucaPrete:


modules/net-ilb/groups.tf
```diff
@@ -21,7 +21,7 @@ resource "google_compute_instance_group" "unmanaged" {
   project     = var.project_id
   zone        = each.value.zone
   name        = each.key
-  description = "Terraform-managed."
+  description = var.description
```
ludoo, line 24: can we keep the hardcoded if var description is null?

ludoo (approved):
discussed with Luca

PR #1418, opened 2023-06-05
Network Load Balancer module

ludoo:


modules/net-nlb/health-check.tf
```diff
@@ -0,0 +1,119 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# tfdoc:file:description Health check resource.
+
+locals {
+  hc       = var.health_check_config
+  hc_grpc  = try(local.hc.grpc, null) != null
```
juliocc, line 21: non-nullable and remove trys?

modules/net-nlb/outputs.tf
```diff
@@ -0,0 +1,77 @@
...
+}
+
+output "groups" {
+  description = "Optional unmanaged instance group resources."
+  value       = google_compute_instance_group.unmanaged
+}
+
+output "health_check" {
+  description = "Auto-created health-check resource."
+  value       = try(google_compute_health_check.default.0, null)
```
juliocc, line 61: IMO `one()` instead of `try()` would make more sense here

juliocc (approved):
two small comments but LGTM

modules/net-nlb/outputs.tf
ludoo: it might not be defined but passed in

PR #1420, opened 2023-06-05
Move net-dedicated-vlan-attachment module to net-vlan-attachment and …

apichick:
…added support for Partner interconnect


modules/net-vlan-attachment/main.tf
```diff
@@ -48,7 +48,7 @@ resource "google_compute_interconnect_attachment" "default" {
   vlan_tag8021q            = var.vlan_tag
   admin_enabled            = var.admin_enabled
   encryption               = local.ipsec_enabled ? "IPSEC" : null
-  type                     = "DEDICATED"
```
wiktorn, line 51: Will `type` auto-magically default to `PARTNER` when `edge_availability_domain` is provided?

modules/net-vlan-attachment/variables.tf
```diff
@@ -22,9 +22,9 @@ variable "admin_enabled" {
 
 variable "bandwidth" {
   # Possible values @ https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_interconnect_attachment#bandwidth
-  description = "The bandwidth assigned to the VLAN attachment (e.g. BPS_10G)."
+  description = "The bandwidth assigned to the VLAN attachment (e.g. BPS_10G). Only applicable for Dedicated Interconnect."
```
wiktorn, line 25: Just loose comment on this. Maybe it's worth to(comment above) group Partner Interconnect related parameters under one variable, and Dedicated Interconnect related parameters under the other? We could use this to infer `type` for `google_compute_interconnect_attachment` and keep default bandwidth

sruffilli:
Two general comments:

1. (my ignorance) I'm unclear about what happens when creating a partner interconnect via terraform, and whether the "pending"-state vlan attachment can be used ad a resource. If not, everything that depends on those VLAN attachment would fail (I'm especially thinking about the blueprint here)
2. Before merging we need to test these changes with in the real world :( unfortunately while I've had the chance to test a dedicated interconnect, testing a partner interconnect involves external stakeholders and I'm not sure we'll be able to test it internally.

Besides that, good job!

modules/net-vlan-attachment/variables.tf
apichick: actually that's a good idea, I will do a few changes.

wiktorn: Thanks a lot @apichick, LGTM

modules/net-vlan-attachment/README.md
```diff
@@ -0,0 +1,518 @@
...
+  source      = "./fabric/modules/net-vlan-attachment"
+  network     = "mynet"
+  project_id  = "myproject"
+  region      = "europe-west8"
+  name        = "vlan-attachment-a"
+  description = "interconnect-a vlan attachment 0"
+  peer_asn    = "65000"
+  router_config = {
+    create = false
+    name   = google_compute_router.interconnect-router.id
```
wiktorn, line 118: I guess it should be here `google_compute_router.interconnect-router.name`, this applies to many examples below too.

modules/net-vlan-attachment/README.md
```diff
@@ -0,0 +1,518 @@
...
+
+```hcl
+resource "google_compute_router" "interconnect-router" {
+  name    = "interconnect-router"
+  network = "mynet"
+  project = "myproject"
+  region  = "europe-west8"
+  bgp {
+    advertise_mode    = "CUSTOM"
+    asn               = 64514
```
wiktorn, line 17: When testing that I get:
```
Error: Error creating InterconnectAttachment: googleapi: Error 400: Invalid value for field 'resource.router': '[...]/routers/interconnect-router'. The Cloud Router used by PARTNER type interconnect attachments must be assigned a local ASN of '16550', invalid
```

wiktorn:
> Two general comments:
> 
> 1. (my ignorance) I'm unclear about what happens when creating a partner interconnect via terraform, and whether the "pending"-state vlan attachment can be used ad a resource. If not, everything that depends on those VLAN attachment would fail (I'm especially thinking about the blueprint here)

I just tested. After creating the attachment it looks as follows:
```tf
$ terraform state show 'module.example-va.google_compute_interconnect_attachment.default'
# module.example-va.google_compute_interconnect_attachment.default:
resource "google_compute_interconnect_attachment" "default" {
    admin_enabled             = true
    creation_timestamp        = "2023-06-07T05:34:31.372-07:00"
    description               = "Example vlan attachment"
    edge_availability_domain  = "AVAILABILITY_DOMAIN_ANY"
    encryption                = "NONE"
    id                        = "projects/***/regions/europe-west1/interconnectAttachments/vlan-attachment"
    ipsec_internal_addresses  = []
    mtu                       = "1500"
    name                      = "vlan-attachment"
    pairing_key               = "***/europe-west1/any"
    private_interconnect_info = []
    project                   = "***"
    region                    = "https://www.googleapis.com/compute/v1/projects/***/regions/europe-west1"
    router                    = "https://www.googleapis.com/compute/v1/projects/***/regions/europe-west1/routers/interconnect-router"
    self_link                 = "https://www.googleapis.com/compute/v1/projects/***/regions/europe-west1/interconnectAttachments/vlan-attachment"
    state                     = "PENDING_PARTNER"
    type                      = "PARTNER"
    vlan_tag8021q             = 0
}
```
Looks like perfectly usable for use as reference in other resources and the blueprint should be fine.

> 2. Before merging we need to test these changes in the real world :( unfortunately while I've had the chance to test a dedicated interconnect, testing a partner interconnect involves external stakeholders and I'm not sure we'll be able to test it internally.

Agreed. Though I think we can test quite a lot even without having access to test partner interconnect to iron this out.

modules/net-vlan-attachment/README.md
apichick: A true need to update the readme.

apichick: changed

wiktorn (approved):

PR #1427, opened 2023-06-07
Updating hub-and-spoke peering blueprint to use HA VPN.

mark1000:
Updating terraform to use HA VPN. Updating README to reflect the change and to incorporate some of the text from the reference architecture.


ludoo:
We have an example of a hub and spoke via HA VPN in FAST. THanks for this PR, but I don't think we should change this blueprint to switch from peering to VPN. Maybe pointing to the FAST example from the README is enough, all FAST stages are usable in isolation.

mark1000:
Hi Ludovico,
Understood. However, unless I goofed, this PR doesn't change the
hub-and-spoke configuration itself. All it does is change the transitive
routing bit from static VPN to HA VPN setup. I did this because we're
trying to get away from static VPN, if I understand the product goals
correctly.
I definitely don't want to change the core goal of the blueprint.
Additional thoughts on that?
Thanks.
Mark

ludoo:
Ooops sorry, I had a quick look and got a completely wrong idea :( It should teach me never to do things in a hurry.

Do you need help with the failing checks?

mark1000:
Thanks Ludovico!
Julio has offered to help on the failing checks. I understand one of them
now. I think I can get through the others, but I'll definitely ask for help
if I get stuck. Thanks!
Mark

juliocc (approved):

PR #1434, opened 2023-06-12
Add subnets id output, expand net-address outputs

juliocc:


modules/net-address/outputs.tf
```diff
@@ -18,54 +18,38 @@ output "external_addresses" {
   description = "Allocated external addresses."
   value = {
     for address in google_compute_address.external :
-    address.name => {
-      address   = address.address
-      self_link = address.self_link
-    }
+    address.name => address
```
ludoo, line 21: self links are used in a few places, can we add a second set of outputs for ids?

ludoo (approved):

modules/net-address/outputs.tf
ludoo: god I'm dumb... :)

PR #1445, opened 2023-06-16
Bump TF version in all workflow templates to coincide with module requirements

kthhrv:
Bootstrap was bumped in #1414

my networking repo GH action was failing 
image


ludoo (changes requested):
THanks for spotting this, can you do the same for all templates across all FAST stages?

ludoo (approved):

PR #1446, opened 2023-06-19
fixup(project-factory): Use the correct KMS Service Agents attribute …

alloveras:
## Intent

To fix a bug affecting the `project-factory` blueprint that was preventing the `kms_service_agents`  entries in `**/*.yaml` from being propagated into the project blueprint module.

## Problem

The file `fast/stages/3-project-factory/dev/main.tf` seems to expect the YAML key to be named `kms` instead of `kms_service_agents`. The key lookup for that attribute is wrapped in a `try()` clause which causes the naming mismatch to go unnoticed because the execution always falls back to the `{}` default value and, thus, any values for `kms_service_agents` are being discarded.

## Solution
To update the `fast/stages/3-project-factory/dev/main.tf` and `README.md` files to use the correct attribute name for KMS Service Agents configurations.

### DIFF
--- a/blueprints/factories/project-factory/README.md
+++ b/blueprints/factories/project-factory/README.md
@@ -67,7 +67,7 @@ module "projects" {
   folder_id              = each.value.folder_id
   group_iam              = try(each.value.group_iam, {})
   iam                    = try(each.value.iam, {})
-  kms_service_agents     = try(each.value.kms, {})
+  kms_service_agents     = try(each.value.kms_service_agents, {})
   labels                 = try(each.value.labels, {})
   org_policies           = try(each.value.org_policies, {})
   prefix                 = each.value.prefix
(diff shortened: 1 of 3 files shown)


ludoo (approved):

alloveras:
Hey @ludoo,
can you provide some guidance on how to get the unit tests working? I tried to get them working by adding the new resources that will be created now that KMS Service Agents are handled properly but, for some reason, I am still getting test failures reporting a mismatch between expected and actual resource counts 😢 

Thank you in advance!

juliocc:
Just change [this line](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/blueprints/factories/project-factory/README.md?plain=1#L79) to reflect `resources=37`

As for the local tests, you're not selecting the right tests. Try with `pytest -vv -k 'blueprints and project-factory'`

alloveras:
> Just change [this line](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/blueprints/factories/project-factory/README.md?plain=1#L79) to reflect `resources=34`
> 
> As for the local tests, you're not selecting the right tests. Try with `pytest -vv -k 'blueprints and project-factory'`

Right, thank you for the pointer 🙇  Now I wonder if the changes that I did before to `//tests/blueprints/factories/project_factory/examples/example.yaml` are actually required?

PR #1449, opened 2023-06-19
Added iam for DNS managed zone to dns module

apichick:


modules/dns/main.tf
```diff
@@ -178,6 +178,14 @@ resource "google_dns_managed_zone" "public" {
   }
 }
 
+resource "google_dns_managed_zone_iam_binding" "iam_bindings" {
+  for_each     = coalesce(var.iam, {})
+  project      = var.project_id
+  managed_zone = (var.type == "public") ? google_dns_managed_zone.public[0].name : google_dns_managed_zone.non-public[0].name
```
ludoo, line 184: nit: can you split this line?

```hcl
  managed_zone = (
    var.type == "public"
    ? google_dns_managed_zone.public[0].name
    : google_dns_managed_zone.non-public[0].name
  )

```

ludoo (approved):

PR #1453, opened 2023-06-20
data-platform-minimal - correct typo

kthhrv:
`container-egine` errors

```
│ Error: Invalid index
│ 
│   on ../../../../modules/project/shared-vpc.tf line 74, in resource "google_project_iam_member" "shared_vpc_host_robots":
│   74:     : "serviceAccount:${local.service_accounts_robots[each.value.service]}"
│     ├────────────────
│     │ each.value.service is "container-egine"
│     │ local.service_accounts_robots is object with 177 attributes
│ 
│ The given key does not identify an element in this collection value.
```


lcaggio (approved):
Thanks for spotting this!

PR #1455, opened 2023-06-21
match readme groups with variables file

CanburakTumer:
I faced error during my deployment even though I created `data-engineers group`. Then I found out that variables file uses `gcp-data-engineers` as default. The same is valid for `data-security`.

reference lines in `variables.tf`:
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/261ad646a821851fb4c1e42d26fefa41e8aa371f/blueprints/data-solutions/shielded-folder/variables.tf#L68

https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/261ad646a821851fb4c1e42d26fefa41e8aa371f/blueprints/data-solutions/shielded-folder/variables.tf#L69

### DIFF
--- a/blueprints/data-solutions/shielded-folder/README.md
+++ b/blueprints/data-solutions/shielded-folder/README.md
@@ -45,8 +45,8 @@ User groups provide a stable frame of reference that allows decoupling the final
 
 We use groups to control access to resources:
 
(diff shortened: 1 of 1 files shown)


ludoo (approved):
Thanks for spotting this!

PR #1456, opened 2023-06-21
add missing variable image_uri

jose-bermudez-digitalfemsa:
For master_config, the variable image_uri is missing

PR #1457, opened 2023-06-21
Added missing variable image_uri

jlbermudez-imagenevolutiva:
for master_config the variable image_uri is missing


wiktorn:
Let's track htis in #1456

PR #1463, opened 2023-06-23
Compute vm snapshot policy attachment (issue 1462)

eandresr:
Here we able to configure disk snapshots, for just boot (or not), or other attached disks, each one can have it's own.


modules/compute-vm/variables.tf
```diff
@@ -35,14 +35,21 @@ variable "attached_disk_defaults" {
   }
 }
 
+variable "boot_disk_snapshot_policy" {
```
ludoo, line 38: can we move this to the `boot_disk` variable?

eandresr: much better!

modules/compute-vm/main.tf
```diff
@@ -116,6 +116,44 @@ resource "google_compute_region_disk" "disks" {
   }
 }
 
+# Disks snapshot handler
+data "google_compute_instance" "instance" {
```
ludoo, line 120: Why do you need a data source?

ludoo:
I am closing this, implementing this in the module should support defining the policy with support for instance and snapshot schedule, and avoid using datasources. I will try and send a PR for this in the next few days.

PR #1466, opened 2023-06-23
Rename network load balancer modules

ludoo:
https://cloud.google.com/load-balancing/docs


sruffilli:
There are still quite a few instances where "ILB" and "GLB" are referred to, mostly as descriptions or module names.
https://gist.github.com/sruffilli/fb617854cf1ef5a632f5312943fc3233

I guess we're ok with that - just making sure.

ludoo:
Yep this is only the code part. READMEs will take a bit more care and can be done incrementally imho.

sruffilli:
LGTM with some additional changes which can be postponed
https://gist.github.com/sruffilli/6f7d35c15cc907326e5ffbad637a09e0

sruffilli (approved):

juliocc (approved):
There are more instances of the old names but let's merge this PR and we update the rest as we go along

PR #1475, opened 2023-06-28
Minimal Data Platform - Shared VPC

lcaggio:
Fixing the logic on how shared VPC informations are computed and handled.


blueprints/data-solutions/data-platform-minimal/02-processing.tf
```diff
@@ -50,12 +50,12 @@ locals {
   processing_subnet = (
     local.use_shared_vpc
     ? var.network_config.subnet_self_link
-    : module.processing-vpc.0.subnet_self_links["${var.region}/${var.prefix}-processing"]
+    : try(module.processing-vpc.0.subnet_self_links["${var.region}/${var.prefix}-processing"], null)
```
juliocc, line 53: why do you need this `try`?

juliocc (approved):

blueprints/data-solutions/data-platform-minimal/02-processing.tf
lcaggio: Mobile.processing-vpc.0 does not exists when a shared vpc is configured. I can't find a better way to fix it. Merging this PR, we can improve in a future PR.

PR #1479, opened 2023-06-29
Update ncc-spoke-ra module to explicity request ncc hub id when referencing existing hubs

simonebruzzechesse:
Modified module input variable to deal with ncc hub creation (passing the name of the hub to create) as well as referencing existing hub via id (instead of the name). Only specifying the name of the hub (even if in the same project) led to strange terraform behaviour (name of the hub being interpreted as the project id somehow), at least in the latest version of the provider.

### DIFF
--- a/modules/ncc-spoke-ra/variables.tf
+++ b/modules/ncc-spoke-ra/variables.tf
@@ -21,7 +21,7 @@ variable "data_transfer" {
 }
 
 variable "hub" {
-  description = "The name of the NCC hub to create or the id for an existing hub to use."
+  description = "The NCC hub. You should either provide an existing hub id or a hub name if create is true."
   type = object({
     create      = optional(bool, false)
     description = optional(string)
--- a/modules/ncc-spoke-ra/variables.tf
+++ b/modules/ncc-spoke-ra/variables.tf
@@ -29,7 +29,7 @@ variable "hub" {
     name        = optional(string)
   })
   validation {
-    condition     = var.hub.create && var.hub.name != null || !var.hub.create && var.hub.id != null
+    condition     = var.hub.create && var.hub.name != null || var.hub.create == false && var.hub.id != null
     error_message = "Name is required for configuring new ncc hub while referencing existing hub requires id."
   }
 }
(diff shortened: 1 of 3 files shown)


modules/ncc-spoke-ra/variables.tf
```diff
@@ -21,12 +21,17 @@ variable "data_transfer" {
 }
 
 variable "hub" {
-  description = "The name of the NCC hub to create or use."
+  description = "The name of the NCC hub to create or the id for an existing hub to use."
```
LucaPrete, line 24: Should we fix a bit this description? It seems we're passing an object rather than a string.

Example:
The NCC hub. You should either provide an existing hub id or a hub name if create is true.

modules/ncc-spoke-ra/variables.tf
```diff
@@ -21,12 +21,17 @@ variable "data_transfer" {
...
   type = object({
     create      = optional(bool, false)
     description = optional(string)
-    name        = string
+    id          = optional(string)
+    name        = optional(string)
   })
+  validation {
+    condition     = var.hub.create && var.hub.name != null || !var.hub.create && var.hub.id != null
```
LucaPrete, line 32: While it's a bit more lengthy I'd try to be more explicit.
btw I think hub name should be populated if hub create is true. Am I missing something?

For example:
`var.hub_create && var.hub.name == null || var.hub.create == false && var.hub.id == null`
```

LucaPrete:
Thanks @simonebruzzechesse . LGTM in general, besides minor comments

modules/ncc-spoke-ra/variables.tf
simonebruzzechesse: Updated description as suggested!

modules/ncc-spoke-ra/variables.tf
simonebruzzechesse: Made the condition more explicit replacing ! with == false, overall the condition LGTM (tested as well locally)

sruffilli (approved):

LucaPrete (approved):

PR #1492, opened 2023-07-05
Peering dashboard

aurelienlegrand:
Adding a monitoring dashboard for key VPC and VPC peering group network quotas that most of our customers care about.

I added this only in the networking-a-peering folder, should I also add it for other networking architecture options in fast?
Should I also add alerting policies when your utilization is above 80% for any of these quotas?


ludoo:
> Adding a monitoring dashboard for key VPC and VPC peering group network quotas that most of our customers care about.
> 
> I added this only in the networking-a-peering folder, should I also add it for other networking architecture options in fast? Should I also add alerting policies when your utilization is above 80% for any of these quotas?

Yes, it's best to have it in all networking stages. Alerts would be nice, why not. :)

ludoo (approved):

aurelienlegrand:
I will add the alerts in a future PR but the dashboard is ready to go.

PR #1496, opened 2023-07-07
Allow using a separate resource for boot disk in compute-vm module

ludoo:
This is an interesting use case brought up by @drebes where the boot disk is managed via a separate resource from the instance. This allows recreating the instance from Terraform (eg by tainting it) while still preserving the disk.

### DIFF
--- a/modules/compute-vm/main.tf
+++ b/modules/compute-vm/main.tf
@@ -63,6 +63,27 @@ locals {
   termination_action = var.options.spot ? coalesce(var.options.termination_action, "STOP") : null
 }
 
+resource "google_compute_disk" "boot" {
+  count   = !var.create_template && var.boot_disk.use_independent_disk ? 1 : 0
+  project = var.project_id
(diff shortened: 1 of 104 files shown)


juliocc (approved):

modules/compute-vm/main.tf
```diff
@@ -172,12 +193,26 @@ resource "google_compute_instance" "default" {
   }
 
   boot_disk {
-    auto_delete             = var.boot_disk.auto_delete
-    source                  = var.boot_disk.source
-    disk_encryption_key_raw = var.encryption != null ? var.encryption.disk_encryption_key_raw : null
-    kms_key_self_link       = var.encryption != null ? var.encryption.kms_key_self_link : null
+    auto_delete = var.boot_disk.auto_delete
```
drebes, line 196: I think here we can always have auto_delete = false if it is an independent_disk.

PR #1497, opened 2023-07-09
Implement proper support for data access logs in resource manager modules

### LINKED ISSUE #1384: Data access logs support
Add data access logs support to folders and projects

ludoo:
Fixes #1384

- [x] organization
- [x] folder
- [x] project


modules/folder/README.md
```diff
@@ -301,6 +303,56 @@ module "folder-sink" {
...
+This example shows how to set a non-authoritative access log configuration:
+
+```hcl
+module "folder" {
+  source = "./fabric/modules/folder"
+  parent = "folders/657104291943"
+  name   = "my-folder"
+  logging_data_access = {
+    allServices = {
+      ADMIN_READ = ["group:organization-admins@example.org"]
```
juliocc, line 319: I think we should mention somewhere that these are excluded identities. It's not clear at first glance what they mean.

modules/folder/variables.tf
```diff
@@ -89,12 +89,33 @@ variable "iam_additive_members" {
   nullable    = false
 }
 
+variable "iam_policy" {
```
juliocc, line 92: `iam_policy` is mutually exclusive with the other iam variables? If so, I think we should mention it the the description

juliocc (approved):
Looks great. Left a couple of comments but you can address them later if you want.

modules/folder/README.md
ludoo: good point, addressed in the most recent commit

modules/folder/variables.tf
ludoo: good point, addressed in the most recent commit

PR #1499, opened 2023-07-10
Add support for custom description in net-address

simonebruzzechesse:
Added support for custom description in net-address for internal, psc and psa addresses.


juliocc (approved):

wiktorn (approved):

wiktorn:
What about `var.external_addresses` ? Can you add there too?

PR #1501, opened 2023-07-11
Fix in nb-glb-psc-neg-sb-psc-ilbl7-hybrid-neg blueprint

apichick:


ludoo (approved):
that name makes my brain hurt :)

Issue #1291, opened 2023-03-29
Update Cloud Function


juliocc:
- [x] Leverage tf1.3 optionals 
- [ ] Add any missing features
- [ ] Try to make interface similar to Cloud Run
- [x] split into v1 and v2 modules
- [x] Remove `_http` suffix for v2
- [x] Make IAM grants for trigger SA at function level instead of project level

ludoo:
- [ x] split into v1 and v2 modules
- [x ] Remove `_http` suffix for v2
- [ -] Leverage tf1.3 optionals  (partial support)
- [ ] Add any missing features
- [ ] Try to make interface similar to Cloud Run
- [ ] Make IAM grants for trigger SA at function level instead of project level

ludoo:
Closing this as most of the tasks have been completed. We will tackle any missing feature if/when separate issues are opened for those.

PR #1505, opened 2023-07-13
Align vpc_connector on Cloud Run and Cloud Functions v1 and v2

wiktorn:
Modify a bit interface in Cloud Run for vpc_connector (remove vpc information from `revision_annotations`, remove `vpc_connector_create` variable), so it is easier to align the interface with Cloud Functions.

Adopt the same interface for Cloud Functions

See: #1291 

* [ ] Try to make interface similar to Cloud Run (half-ticked)


modules/cloud-function-v1/variables.tf
```diff
@@ -154,22 +154,35 @@ variable "trigger_config" {
 }
 
 variable "vpc_connector" {
-  description = "VPC connector configuration. Set create to 'true' if a new connector needs to be created."
+  description = "Populate this to create a VPC connector."
```
ludoo, line 157: description is not accurate, what if I populate with `create = false`, is it still creating?

modules/cloud-run/variables.tf
```diff
@@ -160,9 +160,7 @@ variable "revision_annotations" {
       max_scale = number
       min_scale = number
     }))
-    cloudsql_instances  = optional(list(string), [])
```
ludoo, line 163: I'm not sure we need to do this. For CR the right place for VPC connector is the annotations block. Maybe this part of the interface needs to stay different.

wiktorn: For Cloud Run using V1 API yes, it lands in annotations, but if we check V2 API, it lands in dedicated fields. This is also the case for trigger configuration, which if we migrate CR to v2 API, we could have same API for CR and CF-v2

ludoo: But we decided to ignore v2 API since it's just a different API interface to the same thing, which is irrelevant when using the provider instead of naked API calls. Right? :)

I feel like we're chasing a red herring, like we did with GKE initially when using the same module for standard and autopilot.

wiktorn:
Closing, as per discussion - worsening Cloud Run interface for similarity with Cloud Functions is not worth it.

PR #1516, opened 2023-07-20
WIP data-platform-minimal - optional google analytics landing project

kthhrv:
Would you be interested in adding an optional google analytics landing project?

This project sets DRS set to allow=All as required

Its best practice to have GA landing in its own project

still todo
- docs
- refactor to support multiple GA properties
- make region for datasets configurable
- grant role/editor on project to analytics-processing-dev@system.gserviceaccount.com as required by GA


lcaggio:
From what I see, the main difference between `land` and `ga-land` is the `tag_bindings`. Why not having the tag as something you can configure in the land project? 

To cover the multiple landingzone scenario: it looks to be setup specific to the use case. It can be handled forking the repo and customizing it to the specific use case. WDYT?

lcaggio:
I see your point and I agree on having a separate project is the right approach. What about having a count in the creation of landing projects and curated projects? with the count that should be >0 (to have always at leas one)?.

@wiktorn WDYT?

lcaggio:
I like the idea to make landing and curated configurable. At that point I would also make buckets and dataset configurable within the project.

@kthhrv WDYT? is it going to cover your use case?

(I will be OOO for the next 3 weeks, not sure I will be able to continue the conversation)

kthhrv:
If the projects are in the output so they can be referenced from the module calling data-platform-minimal then I could do the steps of adding GA service account and disabling DRS from there, so yes I think `making landing and curated projects configurable` would work for me

kthhrv:
can the DRS be set after the project is created rather than as its created
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/50725f4156a627ca7773a5d062cf40c2c697a0ea/blueprints/data-solutions/data-platform-minimal/01-ga-landing.tf#L54-L56

kthhrv:
Closing

Issue #1517, opened 2023-07-21
In project-factory, add support for granting access to host project Serverless VPC Access connectors


ChrisF987:
To enable Cloud Run, Cloud Functions and App Engine to use [Serverless VPC Access connectors](https://cloud.google.com/vpc/docs/serverless-vpc-access) their respective Service Agent requires [Serverless VPC Access User](https://cloud.google.com/iam/docs/understanding-roles#vpcaccess.user) permissions. 

The connectors can be deployed in host or service projects. When deployed in the host project, it would be useful to support enabling this role assignment via a project's yaml configuration, in the same way as host project role assignments are already supported for the GKE Service Agent.

ChrisF987:
Implemented in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1542

PR #1518, opened 2023-07-21
Project factory support for Serverless VPC Access connector permissions

ChrisF987:
Proposed project-factory update to support enabling access for Cloud Run, Cloud Functions and App Engine service agents to use Serverless VPC Access connectors that are deployed in a shared vpc host project. 

Relates to issue https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/1517

### DIFF
--- a/blueprints/factories/project-factory/README.md
+++ b/blueprints/factories/project-factory/README.md
@@ -76,7 +76,7 @@ module "projects" {
   service_identities_iam = try(each.value.service_identities_iam, {})
   vpc                    = try(each.value.vpc, null)
 }
-# tftest modules=7 resources=36 inventory=example.yaml
+# tftest modules=7 resources=38 inventory=example.yaml
 ```
 
 ### Projects configuration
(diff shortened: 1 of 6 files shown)


juliocc:
@ChrisF987 what if you expose the whole `shared_vpc_service_config.service_identity_iam` as a field in the yaml? This would allow using any service identity and not just the serverless one.

wiktorn:
Hi @ChrisF987,

I did a bit different implementation of this feature. One thing that I miss, is that you are also granting `roles/vpcaccess.user` for App Engine Standard robot. I would expect this to be mentioned [here](https://cloud.google.com/appengine/docs/standard/shared-vpc-host-project) but this doc is only mentioning that the role is needed for the user that is doing the deploy. Is this a bug in official docs?

ChrisF987:
> Hi @ChrisF987,
> 
> I did a bit different implementation of this feature. One thing that I miss, is that you are also granting `roles/vpcaccess.user` for App Engine Standard robot. I would expect this to be mentioned [here](https://cloud.google.com/appengine/docs/standard/shared-vpc-host-project) but this doc is only mentioning that the role is needed for the user that is doing the deploy. Is this a bug in official docs?

Hi @wiktorn @juliocc , thanks for your comments, and for the implementation of service identity iam via your other PR. Re if `roles/vpcaccess.user` is necessary for App Engine Standard robot, I've reviewed the docs again and yea I just misread it whilst doing this PR.

ChrisF987:
Implemented in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1542

PR #1519, opened 2023-07-21
Improve Dataplex

lcaggio:
Improve dataplex module:
- Add IAM support
- support different type of assets


modules/project/service-agents.yaml
```diff
@@ -155,6 +155,7 @@
   service_agent: "service-%s@gcp-sa-datapipelines.iam.gserviceaccount.com"
 - name: "dataplex"
   service_agent: "service-%s@gcp-sa-dataplex.iam.gserviceaccount.com"
+  jit: true
```
wiktorn, line 158: Please add role information here: `roles/dataplex.serviceAgent` (and also to [table in the README](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/project#service-identities-requiring-manual-iam-grants))

modules/cloud-dataplex/main.tf
```diff
@@ -52,28 +82,42 @@ resource "google_dataplex_zone" "basic_zone" {
...
+}
 
-  project = var.project_id
+resource "google_dataplex_zone_iam_binding" "binding" {
+  for_each = {
+    for zone_role in local.zone_iam : "${zone_role.zone}-${zone_role.role}" => zone_role
+  }
+  project       = var.project_id
+  location      = var.region
+  lake          = local.zone_resources[each.value.zone].lake
```
wiktorn, line 93: In other places we simplified that to:
```
lake     = google_dataplex_lake.lake.name
```

modules/cloud-dataplex/variables.tf
```diff
@@ -45,8 +53,10 @@ variable "zones" {
...
     type      = string
     discovery = optional(bool, true)
+    iam       = optional(map(list(string)), null)
     assets = map(object({
-      bucket_name            = string
+      resource_name          = string
+      resource_project       = optional(string)
       cron_schedule          = optional(string, "15 15 * * *")
       discovery_spec_enabled = optional(bool, true)
       resource_spec_type     = optional(string, "STORAGE_BUCKET")
```
wiktorn, line 62: Shouldn't we add now conditional validation on types of resources? That only `STORAGE_BUCKET`  and `BIGQUERY_DATASET` are valid?

modules/cloud-dataplex/main.tf
```diff
@@ -52,28 +82,42 @@ resource "google_dataplex_zone" "basic_zone" {
...
+  dataplex_zone = local.zone_resources[each.value.zone].name
```
wiktorn, line 94: Doesn't this equals to `each.value.zone`? (and then `zone_resources` are not needed anymore?)

lcaggio: I created `zone_resources` variable to guarantee dependency between `google_dataplex_zone` and `google_dataplex_zone_iam_binding` and avoid `depends_on`. Do you see any better way?

modules/cloud-dataplex/main.tf
lcaggio: see comment below.

modules/cloud-dataplex/main.tf
wiktorn: Yes, you're right. Though I guess, I'd still prefer direct `google_dataplex_zone.zone[each.value.zone].name` here. And as zone is dependent on lake, then no direct dependency on lake is required.

modules/cloud-dataplex/main.tf
lcaggio: done

wiktorn (approved):

PR #1523, opened 2023-07-25
Fix in event_filters of trigger_config

apichick:


wiktorn:
Can we also add example with event_filters? Would prevent re-introducing such bugs in the future, thanks!

PR #1524, opened 2023-07-26
Change export_public_ip_routes to false by default fixes #1522

### LINKED ISSUE #1522: Peering module shouldn't export public ip routes by default
[export_public_ip_routes](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/net-vpc-peering/variables.tf#L30)

arvindag07:
@juliocc Please review this PR


ludoo (commented):
We have a bug open and the module interface will change.

arvindag07:
Thanks for the update Ludovico.

Issue #1521, opened 2023-07-24
Missing requirements in data-platform-foundations README


danieldeleo:
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/blueprints/data-solutions/data-platform-foundations is missing a step instructing users to enable the following APIs before running Terraform:
* Cloud Billing API 
* Cloud Resource Manager API
* Cloud Pub/Sub API
* Identity and Access Management (IAM) API

lcaggio:
Hi Daniel, thanks for flagging missing details in the README, could you send a PR to improve it? Thanks in advance for your help!

ludoo:
@wiktorn want to take a stab at this since Lorenzo is OOO?

wiktorn:
Added fix for missing groups in #1573.

Not describing anything about enabling API's in default project as discussed in #1525 (i.e. it's part of setting up the service account for terraform that is generic, not specific to this specific blueprint).

PR #1525, opened 2023-07-26
Add terraform to activate required APIs on the provider project

danieldeleo:
Fixes https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/1521


juliocc:
Can you explain the purpose of this PR? 

In any case, we don't use bare data sources as they break our testing framework.

wiktorn:
Is this to address #1521 ?

Two things here:
* is the problem that these services needs to be enabled in provider default project /  application-default quota project?
* there are multiple projects created in the blueprint, if the above is not the case, then those services should be added to those projects

danieldeleo:
Yes this addresses https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/1521 
Those APIs need to be enabled in the provider default project, I was trying to make it easier for the user than asking them to manually enable as a step in the README.

danieldeleo:
> Can you explain the purpose of this PR?
> 
> In any case, we don't use bare data sources as they break our testing framework.

I did notice the breakage and then ran out of time yesterday investigating how the testing works and a looking for a potential fix. Is there any recommendation you have for how to reference the provider project id?

wiktorn:
I think that the best way forward here is to provide this info in README if at all. This is the case for many services / modules and I guess the overall assumption is that you have all of them enabled.

drebes:
> Yes this addresses #1521 Those APIs need to be enabled in the provider default project, I was trying to make it easier for the user rather than asking them to manually enable as a step in the README.

The provider configuration is not part of the blueprint, so I don't think this belongs here. In fact, IUC, this is only because you are using a service account from the provider project that you didn't properly setup (independent of this blueprint). When using Terraform as your regular user account this is not required (and wouldn't even work as the default project in that case is managed by Google).

PR #1528, opened 2023-07-27
Delete default route on vpc creation

### LINKED ISSUE #1401: Always delete default route on VPC module and create an explicit one
To make it easier to add/remove a default route (to the default gateway) on the VPC module, we could always delete the auto-created default route and create a TF managed one that can be created/deleted via flag.

juliocc:
Fixes #1401

### DIFF
--- a/modules/net-vpc/routes.tf
+++ b/modules/net-vpc/routes.tf
@@ -17,6 +17,15 @@
 # tfdoc:file:description Route resources.
 
 locals {
+  _default_route = var.vpc_create && var.delete_default_routes_on_create ? {} : {
+    "${var.name}-default" = {
+      dest_range    = "0.0.0.0/0"
+      next_hop      = "default-internet-gateway"
+      next_hop_type = "gateway"
+      priority      = 1000
+      tags          = null
+    }
+  }
   _googleapis_ranges = {
     private      = "199.36.153.8/30"
     private-6    = "2600:2d00:0002:2000::/64"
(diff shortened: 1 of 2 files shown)


juliocc:
I myself am ambivalent. This will break the VPCs of anyone upgrading from previous versions.

modules/net-vpc/routes.tf
```diff
@@ -17,6 +17,15 @@
 # tfdoc:file:description Route resources.
 
 locals {
+  _default_route = var.vpc_create && var.delete_default_routes_on_create ? {} : {
```
wiktorn, line 20: This will create default_route also when `var.vpc_create` is set to false. Is this intended?

juliocc: should be fixed now. PTAL

wiktorn:
Tested migration scenario. So the end result is that I'm left with default route created by default and additional default route created by module.

It looks like the `delete_default_routes_on_create` is implemented in provider and only when creating the network. So the proper migration path is to remove manually the auto-created one.

I think it will only affect those, who already removed the default route manually.

ludoo:
TBH I don't see the need to this: it covers a use case which is practically non-existent (forget to set delete default route at vpc creation), and can be addressed with a single click in the console.

The fact that it creates an apocalypse for all existing users of the module is a huge red flag, and the two combines make me *not* want this change at all.

ludoo:
discussed it with Julio, closing this

PR #1538, opened 2023-07-28
Extend tfdoc to generate TOCs

juliocc:


tools/tfdoc.py
```diff
@@ -322,15 +323,39 @@ def format_variables(items, show_extra=True):
     yield format
 
 
+def create_toc(readme):
+  'Create a Markdown table of contents a Markdown table of contents.'
```
ludoo, line 327: nit: "a...a"

tools/tfdoc.py
```diff
@@ -322,15 +323,39 @@ def format_variables(items, show_extra=True):
...
+  return "\n".join(lines)
+
+
 # replace functions
 
 
 def get_doc(readme):
   'Check if README file is marked, and return current doc.'
-  m = re.search('(?sm)%s\n(.*)\n%s' % (MARK_BEGIN, MARK_END), readme)
+  m = re.search("(?sm)%s(.*)%s" % (MARK_BEGIN, MARK_END), readme)
```
ludoo, line 347: why are you changing quotes?

tools/tfdoc.py
```diff
@@ -373,24 +398,34 @@ def get_readme(readme_path):
     raise SystemExit(f'Error opening README {readme_path}: {e}')
 
 
-def replace_doc(readme_path, doc, readme=None):
+def render_doc(readme, doc):
   'Replace document in module\'s README.md file.'
-  readme = readme or get_readme(readme_path)
   result = get_doc(readme)
-  if not result:
-    raise SystemExit(f'Mark not found in README {readme_path}')
```
ludoo, line 381: I would keep this exit message

ludoo (approved):
❤️

tools/tfdoc.py
juliocc: For no reason. Reverted

tools/tfdoc.py
juliocc: Fixed

tools/tfdoc.py
juliocc: The meaning changed a bit from the previous version. This function now only ties to render the document (vars, outputs, etc) or leaves it as is. After this is done we do the TOC.

I don't want this function to raise, otherwise we would never render the TOC.

ludoo: But then if we don't have a block we never notice...

juliocc: should we fail in that case? I mean, is the mark mandatory in all READMEs?

ludoo: yep, otherwise there's no point running tfdoc :)

PR #1541, opened 2023-07-28
Added dependency on id output for additive and authorizative IAM bind…

apichick:
…ings


ludoo (changes requested):
I remember doing this in the past and this breaking several things. I would not merge this.

ludoo:
It's simple to test: create a project, a service account in that project, and assign a role to the service account. It will result in a circular dependency, and it's a pattern we use everywhere and need to support.

juliocc:
> It's simple to test: create a project, a service account in that project, and assign a role to the service account. It will result in a circular dependency, and it's a pattern we use everywhere and need to support.

this deserves an ADR

Issue #1527, opened 2023-07-27
[tentative] Automate shared VPC role grants for service identities


juliocc:
Investigate if there's a reasonable approach to simplify the process of granting service identities their required roles in shared VPC host projects.

[discussed offline with @wiktorn and @ludoo]

PR #1526, opened 2023-07-27
feat: network user for dataflow service agent

ericyz:
Add the IAM permission to support dataflow workers running on shared VPC host network

PR #1542, opened 2023-07-28
Grant IAM rights to service identities in host project

wiktorn:
Allow granting IAM permissions in Shared VPC host project based on list of services to be used in service project.
* [x] leverage this functionality in `project-factory` blueprint

Closes: #1527, #1526, #1518


modules/project/shared-vpc.tf
```diff
@@ -17,15 +17,25 @@
 # tfdoc:file:description Shared VPC project-level configuration.
 
 locals {
+  _shared_vpc_agent_config = yamldecode(file("${path.module}/sharedvpc-agent-iam.yaml"))
+  _shared_vpc_agent_config_filtered = [
+    for config in local._shared_vpc_agent_config : config
+    if contains(var.shared_vpc_service_config.grant_iam_for_services, config.service)
```
juliocc, line 23: I think you should also filter this with the apis that are enabled in the project. In other words, something like `contains(stintersection(var.services, var.shared_vpc_service_config.grant_iam_for_services), config.service)`

modules/project/shared-vpc.tf
```diff
@@ -34,11 +44,9 @@ locals {
       try(var.shared_vpc_host_config.service_projects, null), []
     )
   }
-  svpc_service_config = coalesce(var.shared_vpc_service_config, {
-    host_project = null, service_identity_iam = {}
-  })
+
   svpc_service_iam = {
-    for b in local._svpc_service_iam : "${b.role}:${b.service}" => b
+    for b in distinct(concat(local._svpc_service_iam, local._shared_vpc_agent_grants)) : "${b.role}:${b.service}" => b
```
juliocc, line 49: `distinct+concat`, or `setunion`

juliocc (commented):
I like it

juliocc (approved):

modules/project/shared-vpc.tf
```diff
@@ -59,7 +67,7 @@ resource "google_compute_shared_vpc_service_project" "service_projects" {
 
 resource "google_compute_shared_vpc_service_project" "shared_vpc_service" {
   provider        = google-beta
-  count           = local.svpc_service_config.host_project != null ? 1 : 0
+  count           = var.shared_vpc_service_config.host_project != null ? 1 : 0
```
juliocc, line 70: One more thing. Can you add a precondition here to ensure that `grant_iam_for_services` only contains APIs mentioned in the yaml?

modules/project/shared-vpc.tf
ludoo: we might not have the APIs here Julio

modules/project/variables.tf
```diff
@@ -292,12 +292,24 @@ variable "shared_vpc_host_config" {
...
 variable "shared_vpc_service_config" {
   description = "Configures this project as a Shared VPC service project (mutually exclusive with shared_vpc_host_config)."
-  # the list of valid service identities is in service-accounts.tf
+  # the list of valid service identities is in service-agents.yaml
   type = object({
-    host_project         = string
-    service_identity_iam = optional(map(list(string)))
+    host_project           = string
+    service_identity_iam   = optional(map(list(string)), {})
+    grant_iam_for_services = optional(list(string), [])
```
ludoo, line 299: can we rename to `service_identity_iam_grants`?

modules/project/README.md
```diff
@@ -252,6 +252,29 @@ module "service-project" {
...
+  }
+}
+
+module "service-project" {
+  source = "./fabric/modules/project"
+  name   = "my-service-project"
+  shared_vpc_service_config = {
+    host_project = module.host-project.project_id
+    grant_iam_for_services = [
+      "container.googleapis.com",
```
ludoo, line 271: I would skip `.googleapis.com`.

ludoo (approved):
Very nice, dropped a couple comments but I really like this!

modules/project/README.md
juliocc: I kind of like the googleapis suffix as it kind of conveys that the field only makes sense if you're using that API (i.e. you wouldn't put container here if you don't enable the container API)

ludoo: I see your point, it's fine it's not like typing it 2-3 times would be a problem

modules/project/variables.tf
wiktorn: Sound good until I realize that I have `service_identity_iam` just above. Maybe `services_iam_grants`?

modules/project/shared-vpc.tf
wiktorn: We can make it more complex like:
```hcl
    if contains(var.shared_vpc_service_config.services_iam_grants, config.service) && (
      var.services == null || contains(var.services, config.service)
    )
```

If you add more services here, than you activate, the should fail on missing service identity. Filtering here, will make sure that there is no failure at the cost of not letting the user now, that his intent was not fulfilled. 

So I prefer not to add such filtering.

modules/project/variables.tf
ludoo: `service_iam_grants` ? :)

modules/project/shared-vpc.tf
wiktorn: One of the usage patterns that I would like to allow is something like that:
```hcl
module "service-project" {
  source          = "./fabric/modules/project"
  services = [
    "alloydb.googleapis.com",
    "run.googleapis.com",
  ]
  shared_vpc_service_config = {
    host_project = module.project.project_id
    grant_iam_for_services = module.service-project.services
  }
}
```

I know that I'm contradicting here a bit myself from the above comment, but ability not to repeat myself, and not having to check in YAML file, which services are allowed here, is reducing overhead when using this module.

Let's allow user to have magic in their own code 🧙

modules/project/shared-vpc.tf
wiktorn: Thanks, corrected.

modules/project/variables.tf
wiktorn: Thanks

blueprints/factories/project-factory/main.tf
```diff
@@ -131,19 +144,18 @@ locals {
     coalesce(var.labels, {}), coalesce(try(var.defaults.labels, {}), {})
   )
   services = distinct(concat(var.services, local._services))
-  vpc = coalesce(var.vpc, {
-    host_project = null, gke_setup = null, subnets_iam = null
-  })
   vpc_cloudservices = (
-    local.vpc_gke_service_agent ||
+    var.vpc.gke_setup.enable_host_service_agent ||
     contains(var.services, "compute.googleapis.com")
```
wiktorn, line 149: Anybody can give a reference to docs why do we grant `roles/compute.networkUser` to cloudservices if using Compute Engine? 

Should this be added to `sharedvpc-agent-iam.yaml`?

ludoo: MIGs managing autoscaling VMs

PR #1543, opened 2023-07-28
initial base module implementation for Dataplex AutoDQ

thinhha:
REST Resource: projects.locations.dataScans:
https://cloud.google.com/dataplex/docs/reference/rest/v1/projects.locations.dataScans#DataScan

Raw terraform resources:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dataplex_datascan

PR #1545, opened 2023-07-29
add dataplex autodq base module

thinhha:
REST Resource: projects.locations.dataScans:
https://cloud.google.com/dataplex/docs/reference/rest/v1/projects.locations.dataScans#DataScan

Raw terraform resources:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dataplex_datascan

copied over from https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1543


tests/modules/dataplex_autodq/examples/datascan_test_inputs.tfvars
```diff
@@ -0,0 +1,105 @@
+# Copyright 2023 Google LLC
```
juliocc, line 1: I think this file is not used.

modules/dataplex-autodq/main.tf
```diff
@@ -0,0 +1,136 @@
...
+  execution_spec {
+    field = var.incremental_field
+    trigger {
+      dynamic "on_demand" {
+        for_each = try(var.execution_schedule) == null ? [""] : []
+        content {
+        }
+      }
+      dynamic "schedule" {
+        for_each = try(var.execution_schedule) != null ? [""] : []
```
juliocc, line 43: Most of these `try()` are not needed. For example, `var.execution_schedule` always succeeds even if it's value is null

modules/dataplex-autodq/variables.tf
```diff
@@ -0,0 +1,166 @@
...
+  type        = string
+}
+
+variable "row_filter" {
+  description = "A filter applied to all rows in a single DataScan job. The filter needs to be a valid SQL expression for a WHERE clause in BigQuery standard SQL syntax. Example: col1 >= 0 AND col2 < 10."
+  type        = string
+  default     = null
+}
+
+variable "rules" {
```
juliocc, line 108: Would it makes sense to explode some of the nested values as shown below?

```hcl
list(object({
    column               = optional(string)
    ignore_null          = optional(bool)
    dimension            = string
    threshold            = optional(number)
    non_null_expectation = optional(object({}))
    range_expectation = optional(object({
      min_value          = optional(number)
      max_value          = optional(number)
      strict_min_enabled = optional(bool)
      strict_max_enabled = optional(bool)
    }))
    regex_expectation = optional(string)
    set_expectation = optional(list(string))
    uniqueness_expectation = optional(object({}))
    statistic_range_expectation = optional(object({
      statistic          = string
      min_value          = optional(number)
      max_value          = optional(number)
      strict_min_enabled = optional(bool)
      strict_max_enabled = optional(bool)
    }))
    row_condition_expectation = optional(string)
    table_condition_expectation = optional(string)
  }))```

**Review by juliocc — COMMENTED** [review]
Dropped a couple of comments

File: modules/dataplex-autodq/main.tf (hunk 2 above)
Comment (thinhha) [reply modules/dataplex-autodq/main.tf:43]: good point. These `try()` was needed when the variable was a nested object -- before refactoring to be an optional string. I will remove this.

File: tests/modules/dataplex_autodq/examples/datascan_test_inputs.tfvars (hunk 1 above)
Comment (thinhha) [reply tests/modules/dataplex_autodq/examples/datascan_test_inputs.tfvars:1]: I think I am missing the tftest.yaml file to turn this into a test according to this guide https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#testing-via-tfvars-and-yaml-aka-tftest-based-tests. Will look into adding this

File: modules/dataplex-autodq/variables.tf (hunk 3 above)
Comment (thinhha) [reply modules/dataplex-autodq/variables.tf:108]: Hey @juliocc, good point. My rationale for not simplifying this structure at the moment is to ensure compatibility / keep a 1:1 mapping for the rules configurations to the API protos and the input format for the gcloud argument `--data-quality-spec-file` in https://cloud.google.com/sdk/gcloud/reference/alpha/dataplex/datascan/create. One example I will add is how to pass in the rules as a yaml input file -- and we would like this yaml input file to be compatible with the REST API / gcloud. Hope that makes sense.

File: tests/modules/dataplex_autodq/examples/datascan_test_inputs.tfvars (hunk 1 above)
Comment (juliocc) [reply tests/modules/dataplex_autodq/examples/datascan_test_inputs.tfvars:1]: Indeed. If you want to run a test with this tfvars you'll have to create a tftest.yaml. Ping me if you need some help with it.

File: modules/dataplex-autodq/variables.tf (hunk 3 above)
Comment (juliocc) [reply modules/dataplex-autodq/variables.tf:108]: Makes sense. I just mentioned this because there are places where we try to simplify the interface to make it easier to use the module (e.g. organization policies). 

You'll be the primary user of this so go with what you feel is right.

File: modules/dataplex-autodq/variables.tf
Comment (saurabh-net) [inline modules/dataplex-autodq/variables.tf:1]: can we make all inputs camelCase instead of snake_case? 

That way, users can reuse the same YAML they use in gcloud for Terraform as well.
I checked to see if gcloud accepts snake_case input, and it doesn't. 


ERROR: (gcloud.alpha.dataplex.datascan.create) Failed to parse value(s) in protobuf [GoogleCloudDataplexV1DataQualitySpec]:
  GoogleCloudDataplexV1DataQualitySpec.rules[0].non_null_expectation
  GoogleCloudDataplexV1DataQualitySpec.rules[1].{ignore_null,set_expectation}

Example gcloud input--

rules:
- nonNullExpectation: {}
  column: word
  dimension: COMPLETENESS
  threshold: 1
- setExpectation : 
    values :
    - 'INR'
    - 'USD'
  column : word
  ignoreNull : true
  dimension : VALIDITY
  threshold : 1
- uniquenessExpectation: {}
  column: word_count
  dimension: UNIQUENESS
- regexExpectation:
    regex:  '^\d{8}[a-zA-Z]{16}$'
  column : word
  ignoreNull : true
  dimension : VALIDITY
  threshold : 1

Comment (ludoo) [reply modules/dataplex-autodq/variables.tf:1]: We don't want camel case in our variables :)

Comment (ludoo) [reply modules/dataplex-autodq/variables.tf:1]: I don't think this is up for debate: variables in our modules are *not* camelcase. Feel free to implement a conversion function from camelcase source files, of course.

Comment (juliocc) [reply modules/dataplex-autodq/variables.tf:1]: I don't see having another input as a problem. In fact, this is exactly what we do with factories in other modules (e.g. net-vpc, net-vpc-firewall-etc).

Terraform conventions is to use snake case for all variables. Camel case is a no-go.

File: modules/dataplex-autodq/README.md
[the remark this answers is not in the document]
Comment (thinhha) [reply modules/dataplex-autodq/README.md:1]: Hey @saurabh-net thanks for your feedback. Looking at our internal proto docs, I think the data profile spec will evolve further, and we also plan to have other types of data scan specs in the future, so I think it make sense to refactor this interface to be compatible with the API.

File: modules/dataplex-autodq/variables.tf
Comment (thinhha) [reply modules/dataplex-autodq/variables.tf:1]: @saurabh-net what do you think about the update change?

File: modules/dataplex-autodq/README.md [hunk 4]
```diff
@@ -221,10 +233,12 @@ module "dataplex-autodq" {
...
     resource = "//bigquery.googleapis.com/projects/bigquery-public-data/datasets/austin_bikeshare/tables/bikeshare_stations"
   }
-  sampling_percent  = 100
-  row_filter        = "station_id > 1000"
   incremental_field = "modified_date"
-  rules             = [for rule in yamldecode(file("config/rules_camel_case.yaml")) : { for k, v in rule : join("_", [for word in flatten(regexall("((?:[A-Z]|[0-9]+)[a-z]*)", title(k))) : lower(word)]) => try({ for kk, vv in v : join("_", [for word in flatten(regexall("((?:[A-Z]|[0-9]+)[a-z]*)", title(kk))) : lower(word)]) => vv }, v) }]
+  data_quality_spec = {
+    sampling_percent = 100
+    row_filter       = "station_id > 1000"
+    rules            = [for rule in yamldecode(file("config/rules_camel_case.yaml")) : { for k, v in rule : join("_", [for word in flatten(regexall("((?:[A-Z]|[0-9]+)[a-z]*)", title(k))) : lower(word)]) => try({ for kk, vv in v : join("_", [for word in flatten(regexall("((?:[A-Z]|[0-9]+)[a-z]*)", title(kk))) : lower(word)]) => vv }, v) }]
```
saurabh-net, line 240: LGTM. This looks reasonably simple. V quick turn-around :)

minor nits:
(1) It would be good to have the example have the YAML at the data quality spec level instead of the rule level, to align 1:1 with the gcloud yaml input.

(2) In a future PR, let's explore a more integrated solution for the yaml support. Perhaps a separate optional argument for it, which inlines the yamldecode and the flatten code.

ludoo: this example in the doc is pretty much unreadable :)

can we internalize the external YAML decode in the module, via a variable that takes a rules file and optional boolean for the conversion?

at the very least, line 240 needs to be broken into multiple lines for legibility

thinhha: sounds good. I'll add a separate variable `data_quality_spec_yaml_file` which expects a file as an input and has a variable switch for converting from camel case to snake case.

modules/dataplex-autodq/variables.tf
saurabh-net: The changes look good for the most part.

I like @ludoo's suggestion to internalize the camelCase to snake_case conversion as a separate variable and inlined logic.

If the user specifies both the rules via the object variable and the file, do we just reject it? Prefer the rule-object one? Merge them?

modules/dataplex-autodq/README.md
ludoo: Another possibly simpler option would be to move the transformation to a `locals` block in the example, using separate local variables for each step.

thinhha: @ludoo that's a great idea. Wish I saw this before doing all the refactoring to add yaml parsing support as part of the module.

Let me know what you think of the latest change.

modules/dataplex-autodq/variables.tf
thinhha: I've implemented this suggestion. Let me know what you think of the latest change.

> If the user specifies both the rules via the object variable and the file, do we just reject it? Prefer the rule-object one? Merge them?

I've added a validation to check that only one option is provided.

modules/dataplex-autodq/main.tf
```diff
@@ -0,0 +1,187 @@
...
+    } :
+  null)
+  _data_quality_spec_string = (
+    var.data_quality_spec != null ?
+    jsonencode(var.data_quality_spec) : (
+      try(var.data_quality_spec_file.convert_camel_case, false) == true ?
+      jsonencode(local._parsed_data_quality_spec_convert_camel_case) :
+      jsonencode(local._parsed_data_quality_spec)
+    )
+  ) # using jsonencode as workaround for "Inconsistent conditional result types" error due to mismatching attributes case
```
juliocc, line 45: IIUC you're keeping either the data from the file or the data form the variable. I would merge `var.data_quality_spec` with the data in `var.data_quality_spec_file`

Also, this jsonencode/jsondecode back-and-forth seems a bit hacky

modules/dataplex-autodq/main.tf
```diff
@@ -0,0 +1,187 @@
...
+          join("_", [for word in flatten(regexall("((?:[A-Z]|[0-9]+)[a-z]*)", title(k))) : lower(word)])
+          => try({
+            for kk, vv in v :
+            join("_", [for word in flatten(regexall("((?:[A-Z]|[0-9]+)[a-z]*)", title(kk))) : lower(word)])
+            => vv
+          }, v)
+        }
+      ]
+    } :
+  null)
```
juliocc, line 37: the else part should be `local._parsed_data_quality_spec` no?

modules/dataplex-autodq/main.tf
```diff
@@ -0,0 +1,187 @@
...
+  _parsed_data_quality_spec = try(yamldecode(file(var.data_quality_spec_file.path)), null)
+  _parsed_data_quality_spec_convert_camel_case = (
+    try(var.data_quality_spec_file.convert_camel_case, false) ?
+    {
+      sampling_percent = try(local._parsed_data_quality_spec.samplingPercent, null)
+      row_filter       = try(local._parsed_data_quality_spec.rowFilter, null)
+      rules = [
+        for rule in local._parsed_data_quality_spec.rules : {
+          for k, v in rule :
+          join("_", [for word in flatten(regexall("((?:[A-Z]|[0-9]+)[a-z]*)", title(k))) : lower(word)])
```
juliocc, line 28: Personally I'm not a fan of this type of string manipulation. I'd rather try to build the exact data structure that you need. I know it's more verbose but explicit is better than implicit.

It would look somethig like this:
```hcl
_parsed_data_quality_spec_convert_camel_case = {
  // outer fields
  rules = [
    for rule in local._parsed_data_quality_spec.rules : {
      column = rule.column
      range_expectation = can(rule.rangeExpectation) ? {
        min_value          = try(rule.rangeExpectation.minValue, null)
        max_value          = try(rule.rangeExpectation.maxValue, null)
        strict_min_enabled = try(rule.rangeExpectation.strictMinEnabled, null)
        strict_max_enabled = try(rule.rangeExpectation.strictMaxEnabled, null)
      } : null
      // other fields...
    }
  ]
}
```


Here's a couple modules where we do something similar: [organization](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/organization/organization-policies.tf#L20), [netf-vpc subnets](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/net-vpc/subnets.tf#L25)

thinhha: fair point. I've refactored the parsing logic accordingly.

modules/dataplex-autodq/main.tf
thinhha: This has been refactored away.

modules/dataplex-autodq/main.tf
thinhha: This has been refactored away.

modules/dataplex-autodq/README.md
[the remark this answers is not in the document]
thinhha: This is complete.

modules/dataplex-autodq/variables.tf
thinhha: this is done

modules/dataplex-autodq/main.tf
thinhha: Let me know what you think of the updated logic.

modules/dataplex-autodq/README.md
thinhha: Yaml parsing support added as discussed. Let's consolidate this discussion to the latest thread started by Julio.

modules/dataplex-autodq/main.tf
thinhha: This is completed.

tests/modules/dataplex_autodq/examples/datascan_test_inputs.tfvars
thinhha: This is completed.

modules/dataplex-autodq/README.md
```diff
@@ -0,0 +1,443 @@
+# Dataplex AutoDQ
+
+This module manages the creation of Dataplex AutoDQ DataScan resources.
```
saurabh-net, line 3:
```suggestion
This module manages the creation of Dataplex DataScan resources.
```

modules/dataplex-autodq/README.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Data Profiling
+
+This example shows how to create a Data Profiling scan using AutoDQ. To create an AutoDQ Data Profiling scan, provide the `data_profile_spec` input arguments as documented in https://cloud.google.com/dataplex/docs/reference/rest/v1/DataProfileSpec.
```
saurabh-net, line 7:
```suggestion
This example shows how to create a Data Profiling type DataScan. To create a Data Profiling scan, provide the `data_profile_spec` input arguments as documented in https://cloud.google.com/dataplex/docs/reference/rest/v1/DataProfileSpec.
```

modules/dataplex-autodq/outputs.tf
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+output "data_scan_id" {
+  description = "Dataplex AutoDQ DataScan ID."
+  value       = google_dataplex_datascan.datascan.data_scan_id
+}
+
+output "id" {
+  description = "A fully qualified Dataplex AutoDQ DataScan identifier for the resource with format projects/{{project}}/locations/{{location}}/dataScans/{{data_scan_id}}."
```
saurabh-net, line 23:
```suggestion
  description = "A fully qualified Dataplex DataScan identifier for the resource with format projects/{{project}}/locations/{{location}}/dataScans/{{data_scan_id}}."
```

modules/dataplex-autodq/variables.tf
```diff
@@ -0,0 +1,165 @@
...
+
+variable "group_iam" {
+  description = "Authoritative IAM binding for organization groups, in {GROUP_EMAIL => [ROLES]} format. Group emails need to be static. Can be used in combination with the `iam` variable."
+  type        = map(list(string))
+  default     = {}
+  nullable    = false
+}
+
+variable "iam" {
+  description = "Dataplex AutoDQ  IAM bindings in {ROLE => [MEMBERS]} format."
```
saurabh-net, line 108:
```suggestion
  description = "Dataplex Datascan  IAM bindings in {ROLE => [MEMBERS]} format."
```

modules/dataplex-autodq/variables.tf
```diff
@@ -0,0 +1,165 @@
...
+
+variable "labels" {
+  description = "Resource labels."
+  type        = map(string)
+  default     = {}
+  nullable    = false
+}
+
+variable "name" {
+  description = "Name of Dataplex AutoDQ Scan."
```
saurabh-net, line 147:
```suggestion
  description = "Name of Dataplex DataScan."
```

modules/dataplex-autodq/variables.tf
```diff
@@ -0,0 +1,165 @@
...
+  type        = string
+}
+
+variable "prefix" {
+  description = "Optional prefix used to generate Dataplex AutoDQ DataScan ID."
```
saurabh-net, line 152: remove AutoDQ across the file

README.md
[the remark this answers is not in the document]
thinhha: That's ok -- good thing we caught it now! I've pushed a change to remove all references of autodq.

saurabh-net (approved):

modules/dataplex-datascan/variables.tf
```diff
@@ -0,0 +1,165 @@
...
+    resource = optional(string)
+  })
+  validation {
+    condition     = length([for k, v in var.data : v if contains(["resource", "entity"], k) && v != null]) == 1
+    error_message = "Datascan data must specify one of 'entity', 'resource'."
+  }
+}
+
+variable "data_profile_spec" {
+  description = "DataProfileScan related setting. Variables description are provided in https://cloud.google.com/dataplex/docs/reference/rest/v1/DataProfileSpec."
```
ludoo, line 30: Nit:" `variables description is` or `variable descriptions are` :) Same in other variables.

modules/dataplex-datascan/variables.tf
```diff
@@ -0,0 +1,165 @@
...
+  })
+}
+
+variable "data_quality_spec" {
+  description = "DataQualityScan related setting. Variables description are provided in https://cloud.google.com/dataplex/docs/reference/rest/v1/DataQualitySpec."
+  default     = null
+  type = object({
+    sampling_percent = optional(number)
+    row_filter       = optional(string)
+    rules = list(object({
```
ludoo, line 44: are expectations mutually exclusive or can they be specified together in the same rule?

modules/dataplex-datascan/main.tf
```diff
@@ -0,0 +1,213 @@
...
+ * limitations under the License.
+ */
+
+locals {
+  prefix                      = var.prefix == null ? "" : "${var.prefix}-"
+  _file_data_quality_spec_raw = var.data_quality_spec_file != null ? yamldecode(file(var.data_quality_spec_file.path)) : tomap({})
+  _file_data_quality_spec = var.data_quality_spec_file == null ? null : {
+    sampling_percent = try(local._file_data_quality_spec_raw.samplingPercent, local._file_data_quality_spec_raw.sampling_percent, null)
+    row_filter       = try(local._file_data_quality_spec_raw.rowFilter, local._file_data_quality_spec_raw.row_filter, null)
+    rules = [
```
ludoo, line 23: Nit: I would move the rules block into its own local, and use a `rules.tf` file for it to keep this file more readable.

ludoo (approved):
Awesome work. I dropped a couple of nits, feel free to self-resolve them and merge when ready.

modules/dataplex-datascan/main.tf
thinhha: Thanks for your review Ludo. Will submit a quick change to move rules out to a separate file, and then merge.

modules/dataplex-datascan/variables.tf
thinhha: each rule object may only have one expectation. I've created a precondition check for this https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1545/files#diff-c2745954b00105f8387b607ccbe9413b234e75a8c7b20dad05c713e91e0cc86cR194

modules/dataplex-datascan/variables.tf
thinhha: Good catch!

modules/dataplex-datascan/variables.tf
ludoo: ah right, sorry I was expecting to see it as a validation rule in the variable... it's ok, we can move it to validation later

thinhha: It was in the variable before :) Now with the additional option to provide the spec as a file with `var.data_quality_spec_file`, I moved the validation step into the module so it is applied regardless of whether the user supplied the input via `var.data_quality_spec` or `var.data_quality_spec_file`

ludoo: okok ignore me :)

juliocc (approved):

modules/dataplex-autodq/main.tf
```diff
@@ -0,0 +1,232 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  prefix                    = var.prefix == null ? "" : "${var.prefix}-"
+  _parsed_data_quality_spec = try(yamldecode(file(var.data_quality_spec_file.path)), null)
```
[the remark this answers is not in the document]
juliocc: ignore, stale comment from previous review.

PR #1553, opened 2023-07-31
Added module for Regional Internal Proxy Load Balancer

apichick:


modules/net-lb-proxy-int/README.md
```diff
@@ -0,0 +1,308 @@
+# Internal Application Load Balancer Module
```
ludoo, line 1: Miren, can you link the module from root and module-level READMEs?

modules/net-lb-proxy-int/variables.tf
```diff
@@ -0,0 +1,279 @@
...
+  nullable = false
+  validation {
+    condition = (var.backend_service_config == null || contains(["NONE", "CLIENT_IP"],
+      var.backend_service_config.session_affinity
+    ))
+    error_message = "Invalid session affinity value."
+  }
+}
+
+variable "backends" {
```
ludoo, line 55: Can you use a single `backend_service_config` variable, which includes a `backends` attribute?

ludoo: You modeled this onto the ILB module but newer ones embed backends into the backend service config.

ludoo (approved):

modules/net-lb-proxy-int/README.md
apichick: Done

modules/net-lb-proxy-int/variables.tf
apichick: Done

PR #1556, opened 2023-08-01
autodq parsing proposal

juliocc:
@thinhha this is how I'd handle the cameCase/snake_case fields in the spec files. Basically first try cameCase and if it fails then you try snake_case. What do you think?

@ludoo wdyt?


thinhha:
This is simpler for the user. I'm fine with this approach.

Some cons of the approach is that it feels a bit magical and also allow end users to mix casing in their spec.

Perhaps I'd still keep the input type as an object. This is because we may want to expand this in the future to allow users to supply an additional file containing default values for variables like ignore_null, row_filters, threshold, etc... as well as a collection of rule templates that they can use in the spec.

ludoo (approved):

PR #1557, opened 2023-08-01
renaming net-vpc-swp to net-swp

skalolazka:


ludoo (approved):
Can you also edit top-level README?

skalolazka:
> Can you also edit top-level README?

Ah, sure, I see it was missing there. Done.

ludoo (approved):

PR #1561, opened 2023-08-02
Removed unused attribute in peer_gateway_config variable

apichick:


modules/net-ipsec-over-interconnect/main.tf
```diff
@@ -54,7 +54,7 @@ resource "google_compute_external_vpn_gateway" "default" {
   name            = var.name
   project         = var.project_id
   description     = var.peer_gateway_config.description
-  redundancy_type = var.peer_gateway_config.redundancy_type
+  redundancy_type = try(length(var.peer_gateway_config.interfaces) == 2, false) ? "TWO_IPS_REDUNDANCY" : "SINGLE_IP_INTERNALLY_REDUNDANT"
```
wiktorn, line 57: Can we make 
```hcl
 interfaces      = optional(list(string), [])
```
And then we could skip the try here?

apichick: done

wiktorn (approved):

PR #1564, opened 2023-08-02
Fixed error of inconsistent conditional result types when evaluating …

apichick:
…local peer_gateway variable

```
I was getting the following error:

│ Error: Inconsistent conditional result types
│
│   on .terraform/modules/nonprod_hub.ipsec_over_interconnect/modules/net-ipsec-over-interconnect/main.tf line 20, in locals:
│   19:     var.peer_gateway_config.create
│   20:     ? try(google_compute_external_vpn_gateway.default[0], null)
│   21:     : var.peer_gateway_config
│     ├────────────────
│     │ google_compute_external_vpn_gateway.default[0] is object with 9 attributes
│     │ var.peer_gateway_config is object with 5 attributes
│     │ var.peer_gateway_config.create is true
│
│ The true and false result expressions must have consistent types. The 'true' value includes object attribute "interface", which is
│ absent in the 'false' value.
```

### DIFF
--- a/modules/net-ipsec-over-interconnect/main.tf
+++ b/modules/net-ipsec-over-interconnect/main.tf
@@ -15,11 +15,6 @@
  */
 
 locals {
-  peer_gateway = (
-    var.peer_gateway_config.create
-    ? try(google_compute_external_vpn_gateway.default[0], null)
-    : var.peer_gateway_config
-  )
   peer_gateway_id = (
     var.peer_gateway_config.create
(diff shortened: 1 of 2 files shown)


wiktorn:
Usual approach that we have in such cases would be to have data resource, that would return the whole structure of `google_compute_external_vpn_gateway` instead of using config. But such data resource does not exists :scream: 

The other approach to fix this is to add to output description, that it returns gateway resource only if autocreated.

PR #1568, opened 2023-08-05
Add support for ipv6 to net-vpc module

### LINKED ISSUE #1530: Net-vpc module doesnot support IPV6
The current "net-vpc" module does not support dual stack subnets. It supports only IPV4 at the moment. There is no attribute to add "STACK_TYPE". 

[stack_type](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_subnetwork.html#stack_type) - (Optional) The stack type for this subnet to identify whether the IPv6 feature is enabled or not. If not specified IPV4_ONLY will be used. Possible values are: IPV4_ONLY, IPV4_IPV6.

ludoo:
Fixes #1530

### DIFF
--- a/modules/net-vpc/README.md
+++ b/modules/net-vpc/README.md
@@ -19,6 +19,7 @@ This module allows creation and management of VPC networks including subnetworks
   - [Custom Routes](#custom-routes)
   - [Private Google Access routes](#private-google-access-routes)
   - [Allow Firewall Policy to be evaluated before Firewall Rules](#allow-firewall-policy-to-be-evaluated-before-firewall-rules)
+  - [IPv6](#ipv6)
 - [Variables](#variables)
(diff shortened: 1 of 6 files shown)


Avinash-jayakumar (commented):
Change looks good. the "private_ipv6_google_access" does not take true or false value. If we give true to false, terraform plan shows a change in resource everytime we run the code. Setting it to null or not having that attribute should work fine. I Could see you have commented it.

ludoo:
> Change looks good. the "private_ipv6_google_access" does not take true or false value. If we give true or false, terraform plan shows a change in resource everytime we run the code. Setting it to null or not having that attribute should work fine. I Could see you have commented it.

Yep. If you look at the code it's a string, but there's no value that can be set.

juliocc (approved):

Issue #1571, opened 2023-08-05
allow_net_admin unsupported in gke-cluster-autopilot


eunanhardy:
Unable to enable NET_ADMIN feature in autopilot module
- expose allow_net_admin in google_container_cluster
- add variable for  allow_net_admin  to gke-cluster-autopilot

[Feature](https://github.com/GoogleCloudPlatform/magic-modules/blob/1680e1587de634f37c567d7c724087abe50b1953/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb#L907) in magic modules

Code Example
```hcl
# gke-cluster-autopilot/main.tf
resource "google_container_cluster" "cluster" {
  provider        = google-beta
  project         = var.project_id
  name            = var.name
  description     = var.description
  location        = var.location
  allow_net_admin = var.allow_net_admin
  node_locations = (
    length(var.node_locations) == 0 ? null : var.node_locations
  )
...
```
```hcl
# gke-cluster-autopilot/variables.tf 
variable "allow_net_admin" {
  description = "Whether to enable NET_ADMIN feature."
  type        = bool
  default     = false
}
```

juliocc:
This field was added in version [4.76.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v4.76.0) of the provider

eunanhardy:
Can I be given permission to raise this as a pull request?

juliocc:
Please do, you don't need any permissions to open a PR.

eunanhardy:
Ive been unable to push my branch eunanhardy/add-net-admin-autopilot. keeps telling me I dont have permission and should fork. Ive verified that I am logged in on all git clients and tried multiple times any suggestions?

juliocc:
Yes, [create a fork and submit your pull request from there](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork).

PR #1575, opened 2023-08-07
Expose allow_net_admin feature in gke-cluster-autopilot module

eunanhardy:
Raised in this issue #1571 

Summary of changes
- Added allow_net_admin variable to gke-cluster-autopilot.

### DIFF
--- a/modules/gke-cluster-autopilot/variables.tf
+++ b/modules/gke-cluster-autopilot/variables.tf
@@ -211,3 +211,9 @@ variable "vpc_config" {
   })
   nullable = false
 }
+
+variable "allow_net_admin" {
+  description = "Enable NET_ADMIN feature on autopilot cluster."
+  type        = bool
+  default     = false
+}
--- a/modules/gke-cluster-autopilot/main.tf
+++ b/modules/gke-cluster-autopilot/main.tf
@@ -32,6 +32,7 @@ resource "google_container_cluster" "cluster" {
   initial_node_count       = 1
 
   enable_autopilot = true
+  allow_net_admin = var.allow_net_admin
 
   addons_config {
     http_load_balancing {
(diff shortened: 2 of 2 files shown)


modules/gke-cluster-autopilot/variables.tf
```diff
@@ -211,3 +211,9 @@ variable "vpc_config" {
   })
   nullable = false
 }
+
+variable "allow_net_admin" {
```
juliocc, line 215: Please move this into the `enable_features` variable.

juliocc (changes requested):
Besides the inline comments, please make sure to:

1. Update the README.md using the tfdoc script. Review the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md) for more details.
2. Bump the provider version in [default-versions.tf](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/default-versions.tf) and copy it to versions.tf in every module.
3. Sign the CLA

modules/gke-cluster-autopilot/variables.tf
ludoo: for the provider, one quick way is to update `default-versions.tf` then run `find . -name versions.tf -exec cp default-versions.tf "{}" \;`

eunanhardy: Is it recommended to update the beta and ga provider or just the ga in this instance. @ludoo

ludoo: we keep them both at the same version to avoid issues

juliocc (approved):

juliocc:
Tests should be fixed now. Once they pass we can go ahead and merge this.

Thanks @eunanhardy!

PR #1578, opened 2023-08-08
Fix: Instance level stateful disk config 

beardedsamwise:
There are a couple of issues with the compute-mig module with respect to instance level stateful disks:

1 - The example and documentation is misleading and inaccurate. You don't need to know the instance name to create an instance level stateful config, you define the instance you want to create within the MIG and attach any disks you want as stateful
2 - There is a bug in the stateful config where the ID is passed for the instance group manager which results in a 404 when applying. 

`The requested URL <code>/compute/v1/projects/redactedproject/zones/australia-southeast1-b/instanceGroupManagers/projects/redactedproject/zones/australia-southeast1-b/instanceGroupManagers/mongodb/createInstances?alt=json</code> was not found on this server.`


ludoo:
Can you fix tests?

modules/compute-mig/README.md
```diff
@@ -356,23 +358,18 @@ module "nginx-template" {
...
-      type         = "PERSISTENT"
-    }
-  }]
   create_template = true
   metadata = {
     user-data = module.cos-nginx.cloud_config
   }
 }
 
+resource "google_compute_disk" "test_disk" {
```
ludoo, line 367: why are you replacing the module-level additional disk with a resource?

beardedsamwise: The stateful virtual machine needs to reference a disk when it's created. I might have missed some functionality within the module that already does this.

beardedsamwise: The module might need some logic like what is in the `compute-vm` module - https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/c5a77ebfe3d44c2a51e23ffb104bb112c9b78ec8/modules/compute-vm/main.tf#L66C11-L66C30

beardedsamwise: I should add, creating a per instance config (stateful instance in a MIG) you need to source a previously created disk (this is how the instance becomes stateful).

ludoo: Yes, but examples should not use resources generally.

I just tried your id/name change and yes, that is needed thanks. :)

For the examples, we can have one with a MIG managing a single instance with autohealing and a stateful config, this is what I just tried

```hcl
module "cos-nginx" {
  source = "../cloud-foundation-fabric/modules/cloud-config-container/nginx"
}

module "nginx-template" {
  source        = "../cloud-foundation-fabric/modules/compute-vm"
  project_id    = "tf-playground-svpc-gce"
  name          = "nginx-template"
  zone          = "europe-west8-b"
  tags          = ["http-server", "ssh"]
  instance_type = "e2-small"
  network_interfaces = [{
    network    = var.vpc_self_links.dev-spoke-0
    subnetwork = var.subnet_self_links.dev-spoke-0["europe-west8/gce"]
  }]
  boot_disk = {
    initialize_params = {
      image = "projects/cos-cloud/global/images/family/cos-stable"
    }
  }
  attached_disks = [{
    name = "data-1"
    size = 10
  }]
  create_template = true
  metadata = {
    user-data = module.cos-nginx.cloud_config
  }
}

module "nginx-mig" {
  source            = "../cloud-foundation-fabric/modules/compute-mig"
  project_id        = "tf-playground-svpc-gce"
  location          = "europe-west8-b"
  name              = "mig-test"
  target_size       = 1
  instance_template = module.nginx-template.template.self_link
  stateful_config = {
    instance-1 = {
      minimal_action                 = "NONE",
      most_disruptive_allowed_action = "REPLACE"
      preserved_state = {
        disks = {
          persistent-disk-1 = {
            source = "projects/tf-playground-svpc-gce/zones/europe-west8-b/disks/data-1"
          }
        }
        metadata = {
          foo = "bar"
        }
      }
    }
  }
}
```

And your example where the disk id is static (in real use it might be a variable, or a resource).

WDYT?

ludoo: Sent the patch to your fork, approving and merging once tests pass. Thanks for flagging this.

PR #1584, opened 2023-08-11
add support for object upload to gcs module

ehorning:
Allow users to upload files to GCS as part of GCS module


juliocc:
Hi @ehorning,

I was discussing with @ludoo and while we're a bit ambivalent about this PR, we're willing to give it a shot as long as others don't have any strong objections (@thinhha, @lcaggio what do you think about this PR).

That being said, the [`google_storage_bucket_object`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_object) resource has a lot more options. If we're going to allow the module to create objects, we should expose all/most of those features. Otherwise, we should just recommend to use the resource directly.

ehorning:
> Hi @ehorning,
> 
> I was discussing with @ludoo and while we're a bit ambivalent about this PR, we're willing to give it a shot as long as others don't have any strong objections (@thinhha, @lcaggio what do you think about this PR).
> 
> That being said, the [`google_storage_bucket_object`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_object) resource has a lot more options. If we're going to allow the module to create objects, we should expose all/most of those features. Otherwise, we should just recommend to use the resource directly.

Is the ambivalence just around the question of whether or not users should be able to upload objects through terraform? I can see how there might be an argument against this in many cases due to the possibility that it might increase the likelihood of sensitive information getting stored in terraform source control. The use case I encountered was around streamlining the upload of sample data via terraform, but I would buy the argument that this might not be a great idea in general so if others think there isn't much need for this functionality, I'm ok with discarding this PR. 

If others do think this could be valuable in certain use cases, I will add those other options.

ludoo:
It's more due to our general approach as infra/net people that modules should be used to create the underlying infra, and we don't really care about workload specifities. Which could be just selective blindness from our side, so that isn't really an issue especially once a few of the data people contributing to this repo chime, as we suspect that supporting objects will make a lot of sense to them.

The real block is doing it properly by supporting all or most of the attributes exposed by the resource. Do you feel like giving it a try? We can help sketch out the variable interface if you want.

ehorning:
> It's more due to our general approach as infra/net people that modules should be used to create the underlying infra, and we don't really care about workload specifities. Which could be just selective blindness from our side, so that isn't really an issue especially once a few of the data people contributing to this repo chime, as we suspect that supporting objects will make a lot of sense to them.
> 
> The real blocker is doing it properly by supporting all or most of the attributes exposed by the resource. Do you feel like giving it a try? We can help sketch out the variable interface if you want.

Sure I can do that. Point about workloads makes sense to me as someone who largely works on infra as well so curious what thoughts others have.

lcaggio:
I see the need to have the ability to upload files for a demo/example purpose.
 
What about having a separate module to upload files into existing bucket? this would make it more flexible.

Note on writing blueprints with demo data: I would suggest to have the demo data deployed as optional, controlled by a bool variable (e.g. `deploy_demo_data`).

modules/gcs/variables.tf
```diff
@@ -132,6 +132,40 @@ variable "notification_config" {
...
+    kms_key_name        = optional(string)
+    customer_encryption = optional(object({
+      encryption_algorithm = optional(string)
+      encryption_key       = string
+    }))
+  }))
+  validation {
+    condition = alltrue([
+      for k, v in var.objects_to_upload :
+      v.content != null
```
juliocc, line 160: This condition would allow both content and source to be specified. I think you want something like `(v.content != null) != (v.source != null)`. In any case I think you this is checked by the provider, you don't have to duplicate it here.

juliocc (approved):
Thanks for the changes! This LGTM now.

One more thing to consider, but this can be a separate PR: an "objects" output with the created object details (self_link, name, hash, etc)

modules/gcs/variables.tf
ehorning: removed validation

modules/gcs/variables.tf
```diff
@@ -132,6 +132,31 @@ variable "notification_config" {
...
+    temporary_hold      = optional(bool)
+    detect_md5hash      = optional(string)
+    storage_class       = optional(string)
+    kms_key_name        = optional(string)
+    customer_encryption = optional(object({
+      encryption_algorithm = optional(string)
+      encryption_key       = string
+    }))
+  }))
+  default = {}
```
juliocc, line 157: One last thing: I'd make this `nullable = false`

juliocc (approved):
LGTM now.

modules/gcs/outputs.tf
```diff
@@ -48,6 +48,18 @@ output "notification" {
...
+output "objects" {
+  description = "Objects in GCS bucket."
+  value = { for k, v in google_storage_bucket_object.objects : k => {
+    crc32c      = v.crc32c
+    md5hash     = v.md5hash
+    self_link   = v.self_link
+    output_name = v.output_name
+    media_link  = v.media_link
+    }
+  }
```
juliocc, line 60: Does
```
  value = google_storage_bucket_object.objects
```
work?

ehorning: This is what you get if you do that:

```
│ Error: Output refers to sensitive values
│
│   on main.tf line 100:
│  100: output "bucket-1-objects" {
│
│ To reduce the risk of accidentally exporting sensitive data that was intended to be only internal, Terraform requires that any root module output containing sensitive data be explicitly marked as sensitive, to confirm your intent.
│
│ If you do intend to export this data, annotate the output value as sensitive by adding the following argument:
│     sensitive = true
```

I would assume because this would export any customer encryption key info inputted as well as everything else

juliocc: Ok then what you have is fine

PR #1588, opened 2023-08-14
feat: 🎸 (modules/cloudsql-instance): enable require_ssl cert support

erabusi:


modules/cloudsql-instance/variables.tf
```diff
@@ -167,6 +167,12 @@ variable "replicas" {
   default = {}
 }
 
+variable "require_ssl" {
+  description = "Enable SSL connections only."
+  type        = bool
+  default     = false
```
ludoo, line 173: Nit: we usually set default to null, which is equivalent to not specifying an optional attribute. And you can also pass on the value directly, without ternary.

ludoo (approved):
Thanks for this!

PR #1590, opened 2023-08-14
GCVE module first release

eliamaldini:
Hi Team,
this is the module for a GCVE Private Cloud deployment.


modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,92 @@
...
+
+variable "description" {
+  description = "Private cloud description."
+  type        = string
+  default     = "Terraform-managed."
+}
+
+variable "management_cidr" {
+  description = "vSphere/vSAN subnets CIDR range."
+  type        = string
```
ludoo, line 25: is there a sensible default that can be provided here?

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,92 @@
...
+  description = "Private cloud name."
+  type        = string
+}
+
+variable "private_connections" {
+  description = "VMWare private connections configuration. It is used to create the gcloud command printed as output."
+  type = map(object({
+    name              = string
+    description       = optional(string, "Terraform-managed.")
+    network_self_link = string
```
ludoo, line 52: put required attributes first, optional after

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,92 @@
...
+    error_message = "Type must be one of DELL_POWERSCALE, NETAPP_CLOUD_VOLUMES, PRIVATE_SERVICE_ACCESS, THIRD_PARTY_SERVICE."
+  }
+}
+
+variable "project_id" {
+  description = "Project id."
+  type        = string
+}
+
+variable "vmwareengine_network_create" {
```
ludoo, line 79: can we find a better name for this?

modules/gcve-private-cloud/main.tf
```diff
@@ -0,0 +1,74 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  region = join("-", slice(split("-", "${var.zone}"), 0, 2))
+  vmwareengine_network = (
```
ludoo, line 19: can we find a better name? something like `vmw_network`

modules/gcve-private-cloud/main.tf
```diff
@@ -0,0 +1,74 @@
...
+    var.vmwareengine_network_create
+    ? try(google_vmwareengine_network.private-cloud-network.0, null)
+    : try(data.google_vmwareengine_network.private-cloud-network.0, null)
+  )
+  psa_peering = {
+    for k, v in data.google_compute_network_peering.psa_peering : k => slice(split("/", "${v.peer_network}"), 6, 7)[0]
```
ludoo, line 25: nits:

- I would break up the line after `:`
- using a regexp might be clearer, or alternatively put a sample value in a comment above the line

modules/gcve-private-cloud/main.tf
```diff
@@ -0,0 +1,74 @@
...
+}
+
+resource "google_vmwareengine_network" "private-cloud-network" {
+  count       = var.vmwareengine_network_create ? 1 : 0
+  provider    = google-beta
+  project     = var.project_id
+  name        = "${local.region}-default"
+  location    = local.region
+  type        = "LEGACY"
+  description = "Private cloud ${var.name} network."
```
ludoo, line 73: description should be configurable by users

ludoo (changes requested):
A few minor comments, then it's good to go.

Remember to also link the module from

- the modules folder README
- the top-level repository folder

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+  description = "Private cloud name."
+  type        = string
+}
+
+variable "private_connections" {
+  description = "VMWare private connections configuration. It is used to create the gcloud command printed as output."
+  type = map(object({
+    name              = string
+    network_self_link = string
+    peering           = string
```
sruffilli, line 52: nit: I would rename this to `peering_name`

ludoo: it's an object with `name` as attribute, maybe `psa_peering`?

eliamaldini: fixed

ludoo: ach, agree `peering_name` is the best choice, I thought we were discussing the variable not the attribute

modules/gcve-private-cloud/variables.tf
eliamaldini: here I would recommend to not add any default and force the user to set a proper value. This CIDR needs to be designed according to the expected scalability of the environment and any change will require the cluster destruction and re-creation.

ludoo: I would then note this in the variable description.

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+}
+
+variable "private_connections" {
+  description = "VMWare private connections configuration. It is used to create the gcloud command printed as output."
+  type = map(object({
+    name              = string
+    network_self_link = string
+    peering_name      = string
+    description       = optional(string, "Terraform-managed.")
+    type              = optional(string, "REGIONAL")
```
juliocc, line 54: "REGIONAL" is not one of the values allowed in the validation below

modules/gcve-private-cloud/main.tf
```diff
@@ -0,0 +1,75 @@
...
+data "google_vmwareengine_network" "private-cloud-network" {
+  count    = var.vmw_network_create ? 0 : 1
+  provider = google-beta
+  project  = var.project_id
+  name     = "${local.region}-default"
+  location = local.region
+}
+
+
+data "google_compute_network_peering" "psa_peering" {
```
juliocc, line 39: You're only using it to build an output, do you really need this data source? Can you just have an input with the data you need?

We tend to avoid having unconditional data sources in modules.

modules/gcve-private-cloud/README.md
```diff
@@ -0,0 +1,100 @@
...
+- Stretched private cloud
+
+## Basic Private Cloud Creation
+
+```hcl
+module "gcve-pc" {
+  source          = "./fabric/modules/gcve-private-cloud"
+  name            = "gcve-pc"
+  project_id      = "gcve-test-project"
+  zone            = "asia-southeast1-a"
```
juliocc, line 29: default to a zone en europe ;)

modules/gcve-private-cloud/README.md
```diff
@@ -0,0 +1,100 @@
...
+  management_cidr = "192.168.0.0/24"
+
+  private_connections = {
+    transit-conn1 = {
+      name              = "transit-conn1",
+      network_self_link = "projects/test-prj-elia-01/global/networks/default",
```
juliocc, line 35: we love you but don't put your name in the example. Use a generic name for the project id

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+    routing_mode      = optional(string, "PRIVATE_SERVICE_ACCESS")
+  }))
```
juliocc, line 56: Do you need all these fields? You not using most of them.

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+    custom_core_count = null
+  }
+}
+
+variable "name" {
+  description = "Private cloud name."
+  type        = string
+}
+
+variable "private_connections" {
```
juliocc, line 47: Should this be `nullable = false`?

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+  type        = string
+  default     = "Terraform-managed."
+}
+
+variable "management_cidr" {
+  description = "vSphere/vSAN subnets CIDR range. To undersatnd the limits, please refer to [GCVE network requirements](https://cloud.google.com/vmware-engine/docs/quickstart-networking-requirements)."
+  type        = string
+}
+
+variable "management_cluster_config" {
```
juliocc, line 28: Should this be `nullable = false`?

modules/gcve-private-cloud/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+ * limitations under the License.
+ */
+
+variable "description" {
+  description = "Private cloud description."
+  type        = string
+  default     = "Terraform-managed."
+}
+
+variable "management_cidr" {
```
juliocc, line 23: Why don't you put this in `management_cluster_config`

juliocc (changes requested):

modules/gcve-private-cloud/variables.tf
eliamaldini: fixed

modules/gcve-private-cloud/main.tf
eliamaldini: it is a workaround to facilitate the extraction of the PSA tenant host project ID reuired by a gcloud command. this is a temporary solution waiting for the GCVE api update.
If it is ok I would leave it to help the user to setup the private cloud connection

modules/gcve-private-cloud/README.md
eliamaldini: fixed

modules/gcve-private-cloud/variables.tf
eliamaldini: yes I agree, fixed

modules/gcve-private-cloud/main.tf
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  region = join("-", slice(split("-", "${var.zone}"), 0, 2))
```
sruffilli, line 18: nit: you're using string interpolation syntax for var.zone where it's not required.

also, alternate approach could be
`regex("([a-z]*-[a-z]*[0-9]{1,2})-([a-z])", var.zone)[0]`

modules/gcve-private-cloud/variables.tf
eliamaldini: it is not used only by the management cluster, but it is used also by some other GCVE core componets. I used the same name of the underlying resource input, but I agree it creates confusion.
I rename it as just "cidr"

modules/gcve-private-cloud/outputs.tf
```diff
@@ -0,0 +1,71 @@
...
+  description = "The private cloud resource"
+  value       = google_vmwareengine_private_cloud.private-cloud
+}
+
+output "vcenter" {
+  description = "Details about a vCenter Server management appliance."
+  value       = google_vmwareengine_private_cloud.private-cloud.vcenter
+}
+
+output "state" {
```
sruffilli, line 52: nit: sort outputs alphabetically

modules/gcve-private-cloud/variables.tf
eliamaldini: actually all og the are used and required in a output variable (where I print the gcloud command). they are not "optional" but more required with a default value

modules/gcve-private-cloud/main.tf
juliocc: Actually, your two tests are failing because of this data source. Consider that tests only run a plan and use fake credentials. You'll never be able to query GCP APIs.

I suggest you try build an example where the peering/psa is created using the `net-vpc` module and show how to use the outputs from the vpc module to extract the data you need as input (without using data sources)

juliocc (approved):

ludoo (approved):

PR #1596, opened 2023-08-16
Seperating GKE standard and Autopilot

avinashkumar1289:
* Change the default cluster created in the different blueprints from GKE Std. to GKE Autopilot.
* There is currently an Autopilot folder. Remove this folder as we change all the blueprints to create GKE Autopilot
* Move all existing blueprints, which are GKE Std., to new folder called GKE-Standard


ludoo:
Can you provide a description of what you're trying to do? If you're cloning all gke blueprints this is not the way to go. 😀

ludoo:
Let's discuss about this. We are also going to publish some new gke blueprints in the next few weeks, this change should wait and once we define an approach, it should land then.

Closing this for now.

PR #1597, opened 2023-08-17
fix null object exception in bootstrap output when using cloudsource

sm3142:
When using Cloud Source repos with FAST, there is no workflow identity federation configuration, so `local.cicd_providers[v["identity_provider"]]` is null. When the bootstrap stage's output module tries to populate the `audience` local, this causes a null object exception.

There may be other ways to address this, but wrapping the call in a try block seemed the most painless to me.


ludoo (approved):
Thanks a lot for the quick fix!

juliocc:
@sm3142 I fixed the linting error. All tests should pass now

PR #1599, opened 2023-08-18
Fixing some typos

bluPhy:


juliocc:
Thanks for fixing these typos!

I've manually fixed a couple of linting errors but for the future, note that the variables/outputs tables are automatically generated using the tools/tfdoc.py script. The easiest path is to update the variable/output description in terraform and then run tfdoc.

juliocc (approved):

PR #1600, opened 2023-08-18
fix(cloud-run): move cpu boost annotation to revision

LiuVII:


modules/cloud-run/main.tf
```diff
@@ -75,6 +72,9 @@ locals {
     var.gen2_execution_environment ? {
       "run.googleapis.com/execution-environment" = "gen2"
     } : {},
+    var.startup_cpu_boost ? {
+      "run.googleapis.com/startup-cpu-boost" = "true"
+    } : {},
```
LiuVII, line 77: according to https://cloud.google.com/run/docs/reference/rest/v1/ObjectMeta must be in revision

modules/cloud-run/main.tf
```diff
@@ -120,9 +120,12 @@ resource "google_vpc_access_connector" "connector" {
...
   min_throughput = var.vpc_connector_create.throughput.min
-  subnet {
-    name       = var.vpc_connector_create.subnet.name
-    project_id = var.vpc_connector_create.subnet.project_id
+  dynamic "subnet" {
+    for_each = var.vpc_connector_create == {} ? [] : [""]
+    content {
+      name       = var.vpc_connector_create.subnet.name
+      project_id = var.vpc_connector_create.subnet.project_id
+    }
```
LiuVII, line 128: without there's a drift is subnet is not specified
```
      + subnet { # forces replacement
          + project_id = (known after apply)
        }
```

LiuVII: actually this still doesn't solve the drift 🤔 maybe the issue is in the provider

juliocc (approved):

PR #1602, opened 2023-08-18
Cloud DNS will become the default DNS provider for new GKE Autopilot …

danielmarzini:
…clusters

Starting August 2023, Cloud DNS will become the default DNS provider for new GKE Autopilot clusters created with version 1.25.9-gke.400 or later, or version 1.26.4-gke.500 or later (effectively replacing kube-dns). The rollout will be gradual and expected to be completed by August 11, 2023.

// code change
The change to the code is to set a default but also let users that have already something to keep their values.


ludoo (approved):

modules/gke-cluster-autopilot/variables.tf
```diff
@@ -93,7 +93,11 @@ variable "enable_features" {
     vertical_pod_autoscaling = optional(bool, false)
   })
   default = {
-
+    dns = {
+      provider = "CLOUD_DNS"
```
ludoo, line 97: Can we use `optional` here? What other providers are supported? If just one, we need to remove the attribute. If two, provider needs to be a `bool`, if more than two we need validation on the value.

ludoo (changes requested):
Dropped a comment on the variable.

danielmarzini:
Perfect, thank you, lemme check about that

gustavovalverde:
Maybe this was expected, based on the `incompatible change` label. I'd like to keep up with the repo changes, but after re-applying  stage`0-bootstrap` I got this error and I'm not completely sure what's required to _migrate_ to the new approach. 

```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.organization.google_organization_iam_member.bindings["roles/compute.orgFirewallPolicyAdmin-group:gcp-network-admins@organization.com"],
│ provider "provider[\"registry.terraform.io/hashicorp/google\"]" produced an unexpected new value: Root resource was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Error applying IAM policy for organization "***REDACTED***": Error setting IAM policy for organization "***REDACTED***": googleapi: Error 400: The role name must be in the form "roles/{role}", "organizations/{organization_id}/roles/{role}", or "projects/{project_id}/roles/{role}"., badRequest
│ 
│   with module.organization.google_organization_iam_binding.bindings["sa_resman_delegated_iam"],
│   on ../../../modules/organization/iam.tf line 51, in resource "google_organization_iam_binding" "bindings":
│   51: resource "google_organization_iam_binding" "bindings" {
```

juliocc:
@danielmarzini can you share how to reproduce that behavior? I just created a new autopilot cluster. After creation I did another apply and I'm not being forced to recreate.

olliefr:
@juliocc @danielmarzini could this be related to the different versions of `google` provider?

[Terraform plans to recreate Autopilot cluster due to `dns_config` change](https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-dns#tf-dns-config)

The behaviour Daniel describes appears to have been fixed in `4.80.0`.

If the updated provider version works as expected, there should be no need to change the default configuration for the module - it would pick the right option by default.

juliocc:
Indeed I'm using version 4.80.0. 

@danielmarzini can you try this version and see if you still get the same behavior?

danielmarzini:
Yep, I'll get back to you later today

danielmarzini:
Closing this one since it happens with older provider only.

PR #1605, opened 2023-08-21
The role roles/run.invoker cannot be granted at the cloud function v2…

apichick:
When you try to apply it at the function level you get the following error:

│
│   with module.apigee_x["int"].module.instance_status_reporter_function[0].google_cloudfunctions2_function_iam_binding.default["roles/run.invoker"],
│   on .terraform/modules/apigee_x.instance_status_reporter_function/modules/cloud-function-v2/main.tf line 157, in resource "google_cloudfunctions2_function_iam_binding" "default":
│  157: resource "google_cloudfunctions2_function_iam_binding" "default" {
│
╵
╷
│ Error: Error applying IAM policy for cloudfunctions2 function "projects/apim-apigee-x-dev-0/locations/europe-west3/functions/instance-monitor": Error setting IAM policy for cloudfunctions2 function "projects/apim-apigee-x-dev-0/locations/europe-west3/functions/instance-monitor": googleapi: Error 400: Invalid argument: 'An invalid argument was specified. Please check the fields and try again.'
│ Details:
│ [
│   {
│     "@type": "type.googleapis.com/google.rpc.DebugInfo",
│     "detail": "generic::invalid_argument: Role roles/run.invoker is not supported for this resource.",
│     "stackEntries": [

PR #1606, opened 2023-08-21
Fix IAM resource and use authoritative/additive IAM bindings for run invoker role in cf2 and cr modules

ludoo:
Keep the injected invoker role for trigger SA, but use additive if role is not present in authoritative roles. Alternative to #1605


modules/cloud-function-v2/main.tf
```diff
@@ -155,12 +138,33 @@ resource "google_cloudfunctions2_function" "function" {
...
+    role != "roles/run.invoker" || !var.trigger_config.service_account_create
+    ? each.value
+    # if invoker role is present and we create trigger sa, add it as member
+    : concat(
+      each.value, ["serviceAccount:${google_service_account.trigger_service_account[0].email}"]
+    )
+  )
+}
+
+resource "google_cloudfunctions2_service_iam_member" "member" {
```
apichick, line 156: I don't think this resource exists. Shouldn't it be google_cloud_run_service_iam_member?

ludoo: this is the cloud function module

ludoo:
Discussed with Miren, closing this she will reopen a new one.

PR #1608, opened 2023-08-21
minor updates to IAM in dataplex-datascan docs

thinhha:
The docs became somewhat outdated after the IAM update. I changed the docs a little bit to clarify what people should be using.

### DIFF
--- a/modules/dataplex-datascan/README.md
+++ b/modules/dataplex-datascan/README.md
@@ -380,15 +380,46 @@ module "dataplex-datascan" {
 
 ## IAM
 
-IAM is managed via several variables that implement different features and levels of control:
+IAM is managed via several variables that implement different levels of control:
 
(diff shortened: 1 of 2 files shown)


modules/dataplex-datascan/README.md
```diff
@@ -380,15 +380,46 @@ module "dataplex-datascan" {
 
 ## IAM
 
-IAM is managed via several variables that implement different features and levels of control:
+IAM is managed via several variables that implement different levels of control:
```
ludoo, line 383: Can you update the description in all modules that use the IAM interface? We have intentionally aligned them and I would like to avoid having a few modules diverge...

ludoo: I would do all in one go or we will start drifting :)

ludoo:
Thinh, I'm going to close this. Feel free to reopen if you manage to align all modules in a single go.

PR #1609, opened 2023-08-21
Use cloud run bindings for cf v2 invoker role, refactor iam handling in cf v2 and cloud run

ludoo:
CF v2 uses cloud run resources to set invoker role on the underlying run service, as [explained here](https://cloud.google.com/sdk/gcloud/reference/functions/add-invoker-policy-binding#DESCRIPTION).

Thanks @apichick for unraveling this.


modules/cloud-function-v2/main.tf
```diff
@@ -24,28 +24,17 @@ locals {
...
     var.service_account_create
     ? google_service_account.service_account[0].email
     : var.service_account
   )
-  trigger_service_account_email = (
-    try(var.trigger_config.service_account_create, false)
-    ? google_service_account.trigger_service_account[0].email
-    : try(var.trigger_config.service_account_email, null)
+  trigger_sa_create = (
+    try(var.trigger_config.service_account_create, false) == true
```
apichick, line 34: why do we need == true?

ludoo: because it could be passed in as `null`, pass through the `try`, and raise a type error as it would not be a boolean

juliocc (approved):

PR #1613, opened 2023-08-23
Cloud SQL activation policy selectable

cmvalla:
This PR make the Cloud SQL activation policy changeable from the default ALWAYS value. it yet does validation on input and is defaulted in order to don't make regressions on previous module. This PR make possible to manage databases that are not alway started (Eg. non production envs, historical data storage etc)


modules/cloudsql-instance/variables.tf
```diff
@@ -13,6 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+variable "activation_policy" {
+  description = "(Optional) This specifies when the instance should be active. Can be either ALWAYS, NEVER or ON_DEMAND. Default is ALWAYS"
+  default  = "ALWAYS"
+  validation {
+    condition     = var.activation_policy == "NEVER" || var.activation_policy == "ON_DEMAND" || var.activation_policy == "ALWAYS" 
+    error_message = "activation_policy must be ALWAYS, NEVER or ON_DEMAND"
```
juliocc, line 21: I'd keep null as the default.

Error message must end in period and the first letter must be uppercase.

modules/cloudsql-instance/variables.tf
```diff
@@ -13,6 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+variable "activation_policy" {
+  description = "(Optional) This specifies when the instance should be active. Can be either ALWAYS, NEVER or ON_DEMAND. Default is ALWAYS"
```
juliocc, line 17: please add type to the variable

juliocc, line 17: Remove "(Optional)"

juliocc (changes requested):
Thanks for this PR. 

Please check the contributing guide to check the linting and test errors. You'll also have to regenerate the README file for the module.

modules/cloudsql-instance/variables.tf
cmvalla: on GCP provider side that null is defaulted to ALWAYS. I Defaulted to ALWAYS for this reason (transparency). you cannot use a null value on provider, I'lll correct the error message

cmvalla:
@juliocc please could you review/Approve in order to check subsequent wf status? thx

modules/cloudsql-instance/main.tf
```diff
@@ -119,6 +119,7 @@ resource "google_sql_database_instance" "primary" {
     }
   }
   deletion_protection = var.deletion_protection
+  activation_policy   = var.activation_policy
```
juliocc, line 122: `activation_policy` is not a top-level field but a field of the `settings`.

Should you also set the same setting for the replicas?

juliocc (changes requested):
I tied the module and I'm getting the following error:

```
Error running command plan: 1
Error: Unsupported argument

  on fabric/modules/cloudsql-instance/main.tf line 122, in resource "google_sql_database_instance" "primary":
 122:   activation_policy   = var.activation_policy

An argument named "activation_policy" is not expected here.
```

Please make sure to test the proposed changes.

juliocc (approved):

cmvalla:
Thank you @juliocc for your valuable work and the patience.

Issue #1617, opened 2023-08-23
feat: Support for NAT IPs in Apigee Instances


billabongrob:
We should be able to add NAT IPs for Apigee X.  This will enable organizations whose backend **does** require IP allow-listing to leverage this module without external modifications.

> Activates the NAT address. The Apigee instance can now use this for Internet egress traffic.

They're supported [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/apigee_nat_address).

PR #1619, opened 2023-08-23
Adding support for NAT in Apigee

billabongrob:
#1617


modules/apigee/main.tf
```diff
@@ -100,6 +100,13 @@ resource "google_apigee_instance" "instances" {
   consumer_accept_list     = each.value.consumer_accept_list
 }
 
+resource "google_apigee_nat_address" "apigee-nat" {
+  for_each    = local.instances
```
juliocc, line 104: I think this doesn't work. You can't use `for_each` and `count` together.

Yo might want to filter `local.instances` with a comprehension.

juliocc (changes requested):

billabongrob:
@juliocc would you mind seeing if the recent changes would meet the requirements?  I'm not sure how to trigger the actions again.

juliocc (approved):

PR #1595, opened 2023-08-16
IAM interface refactor

ludoo:

PR #1621, opened 2023-08-24
fix(organization): use correct role for bindings

gustavovalverde:
Previously, setting a binding was using the binding `key` as the the role, which does not necessarily have the correct format.

Causing errors like:

```
│ Error: Error applying IAM policy for organization "***REDACTED***": Error setting IAM policy for organization "***REDACTED***": googleapi: Error 400: The role name must be in the form "roles/{role}", "organizations/{organization_id}/roles/{role}", or "projects/{project_id}/roles/{role}"., badRequest
│
│   with module.organization.google_organization_iam_binding.bindings["sa_resman_delegated_iam"],
│   on ../../../modules/organization/iam.tf line 51, in resource "google_organization_iam_binding" "bindings":
│   51: resource "google_organization_iam_binding" "bindings" {
```


juliocc:
@ludoo this is related to [this discussion](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1595#issuecomment-1691611972) related to FAST custom roles. I think the problem is with FAST and not with this module

ludoo:
Well, changing the variable type is not a good solution. 😀

ludoo:
> @ludoo this is related to [this discussion](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1595#issuecomment-1691611972) related to FAST custom roles. I think the problem is with FAST and not with this module

let's close this PR and figure out what's happening with FAST, BTW I'm using the new IAM code and don't have the above issue

gustavovalverde:
For FAST this seemed like the right approach based on how the the bindings were defined there (which seems correct). But I certainly can't scope the implications on other modules, even though it solved the problem.

ludoo:
Discussed with Juliio, and you surfaced an error in FAST which was introduced in the last IAM refactor. The role should be the key here

image

but it cannot be as that would trigger an error in the internal loop since the role is custom and its id is dynamic.

Julio came up with a workaround which is similar to the one we use in the project and service account modules' outputs. Alternatively, we will just go back to using a resource for that IAM binding in FAST like before.

Thanks for surfacing this, a fix should land shortly. :)

ludoo:
As a quick workaround, you could use `iam_bindings_additive` for that, and it would work (replace members with member and make sure to merge it with the other bindings).

juliocc:
So we actually already have the fix in place (i.e. to [build the role id statically](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/organization/outputs.tf#L28)).

Need to finish a couple of things first but I'll send a fix ASAP. As a separate issue, we need to [update the `iam_bindings` interface to allow multiple conditions on the same role](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/1622)

PR #1625, opened 2023-08-27
gke-cluster-autopilot: add logging configuration

olliefr:
Although one cannot disable Cloud Logging and Cloud Monitoring integration in GKE Autopilot clusters, one has some flexibility over which control plane components' logs should be ingested: [Available logs](https://cloud.google.com/stackdriver/docs/solutions/gke/installing#available-logs)

This PR adds a new variable and a block to the cluster resource to facilitate that.

I have attempted to imitate the style of the original code to the best of my abilities. I'm happy to change it if I got it wrong.

I run Terraform format; followed by `tfdoc` to generate the README.

Feedback is welcome 🙃 

PS Against my better judgement, I also fixed a few typos in comments elsewhere, I hope it's OK 😈

PPS If this PR is of interest, I could also do the monitoring config to the same standard in a separate PR...


juliocc:
thanks @olliefr for the patch and fixing the typos!

Reading the documentation, it looks like this variable would make sense for standard clusters too. Would you mind porting it to the [`gke-cluster-standard`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/gke-cluster-standard) module too?

> PPS If this PR is of interest, I could also do the monitoring config to the same standard in a separate PR...

Please do!

olliefr:
Hi @juliocc 

I believe `gke-cluster-standard` already has support for logging configuration: https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/c63884d52e8e8450d83e3c6cbe2efa861fe77775/modules/gke-cluster-standard/variables.tf#L143

So what I did, I updated my PR to match the coding style and added input variable validation on top of that so now there is going to be consistent interface between the standard and autopilot modules.

ludoo (approved):

olliefr:
@juliocc I happen to prefer the original interface too! 😆 I reverted the interface to use an object in the latest commit.

I also added a test for `logging_config` to the `README` file. I haven't actually run it because I could not figure out how to fully set up the test environment. But I meticulously compared my test to the existing ones and read the `CONTRIBUTING` file several times over, so I hope I got it right. Any feedback on the test is welcome!

The `logging_config` for `gke-cluster-standard` is a bit more tricky because the validation required there is more complex. But I have already built a little prototype and I think I figured out what to do with the validation. I'll submit a separate PR for `gke-cluster-standard`.

olliefr:
I also just searched the whole `cloud-foundation-fabric` repo for references to `gke-cluster-autopilot` to make sure that introducing `logging_config` does not break any existing code.  Because there was no `logging_config` for `gke-cluster-autopilot` before, there is no existing code to fix 🙃

PR #1630, opened 2023-08-28
[Fix] Add explicit dependency between CR peers and NCC RA spoke creation

LucaPrete:
Currently CR peers may be created before the NCC-RA spoke is created (so before the VM is attached to an NCC spoke).
This errors out because we can't reference a VM as a peer before it gets added to an NCC spoke.

Adding an explicit dependency between the CR peer and the NCC RA spoke.

[tried to add an IMPLICIT dependency but makes the code much more complex and unreadable]


simonebruzzechesse (approved):
LGTM

PR #1631, opened 2023-08-28
Allow single hfw policy association in folder and organization modules

juliocc:


ludoo (approved):

fast/stages/2-networking-c-nva/main.tf
```diff
@@ -46,8 +46,9 @@ module "folder" {
   name          = "Networking"
   folder_create = var.folder_ids.networking == null
   id            = var.folder_ids.networking
-  firewall_policy_associations = {
-    default = module.firewall-policy-default.id
+  firewall_policy_ = {
```
LucaPrete, line 49: As the failed test says :)

LucaPrete (changes requested):

fast/stages/2-networking-c-nva/main.tf
juliocc: and that's why we have them ;)

PR #1632, opened 2023-08-28
Vpc sc allow null for identity type

LudovicEmo:
Hello,

modules/vpc-sc is already supporting **null** for **ingress_policies.identity_type**.
This PR adds support for null for **egress_policies.identity_type**.

**Example**
This is required for example to implement an egress rule for using **Connected sheets** with **Bigquery** for a selected list of identities (refer to [1]).

In the following example, **egressFrom.identityType** is absent (hence null), and the list of identities is managed through identities.

```
- egressTo:
    operations:
    - serviceName: 'bigquery.googleapis.com'
      methodSelectors:
      - permission: 'bigquery.vpcsc.importData'
    resources:
    - projects/628550087766 # Sheets-owned Google Cloud project
  egressFrom:
    identities:
    - serviceAccount: sa1@[REDACTED]
    - serviceAccount: sa2@[REDACTED]
```

As mentioned here [2], the **egress_from.identity_type** is **optional**,  specifies the type of identities that are allowed access to outside the perimeter. **If left unspecified, then members of identities field will be allowed access**.
 
**Changes**
> updated vpc_sc/variables.tf variables default value & validation
> updated the vpc_sc test result for the null capability 
> tfdoc for vpc_sc
> updated fmt for cloudsql-instance/variables.tf (missing a blank line after comment)

**References**
[1] https://cloud.google.com/bigquery/docs/connected-sheets#vpc-service-controls
[2] https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/access_context_manager_service_perimeter


modules/vpc-sc/variables.tf
```diff
@@ -89,10 +89,11 @@ variable "egress_policies" {
...
   validation {
     condition = alltrue([
-      for k, v in var.egress_policies : contains([
+      for k, v in var.egress_policies :
+      v.from.identity_type == null || contains([
         "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
-        "ANY_USER", "ANY_SERVICE_ACCOUNT", ""
-      ], v.from.identity_type)
+        "ANY_USER", "ANY_SERVICE_ACCOUNT"
+      ], coalesce(v.from.identity_type, "-"))
```
juliocc, line 96: I think this coalesce is unnecessary. `v.from.identity_type == null` is already checked explicitly.

juliocc (approved):
Thanks @LudovicEmo. I left a comment below.

Can you please run `tools/tfdoc.py` to update the readme?

juliocc:
@LudovicEmo can you provide a code example where the module fails today. The example you listed in (1) doesn't use null but `ANY_USER_ACCOUNT`

modules/vpc-sc/variables.tf
LudovicEmo: @juliocc just reused the same validation than the one already existing for ingress_policies.from.identity_type, that was already supporting null usage, need to change ?

LudovicEmo:
@juliocc 

> Thanks @LudovicEmo. I left a comment below.
> 
> Can you please run `tools/tfdoc.py` to update the readme?

not sure what is needed here, already updated README with tfdoc in commit 65b32ea

LudovicEmo:
> @LudovicEmo can you provide a code example where the module fails today. The example you listed in (1) doesn't use null but `ANY_USER_ACCOUNT`

@juliocc 

The example here: https://cloud.google.com/bigquery/docs/connected-sheets#vpc-service-controls uses identityType: ANY_USER_ACCOUNT.
However the derived example in my description is using a list of identities instead, with the identities list only, identityType in this case mut not be present, hence must be null/absent.

modules/vpc-sc/variables.tf
juliocc: The the coalesce is not needed in the ingress either.

@ludoo this conditions comes originally from you. Is there a reason to have the coalesce there?

juliocc:
> However here we are not opening for any user account, we are controlling the list of identities.

Understood. Thanks for the context, I think it makes sense. 

> > Thanks @LudovicEmo. I left a comment below.
> > Can you please run `tools/tfdoc.py` to update the readme?
> 
> not sure what is needed here, already updated README with tfdoc in commit [65b32ea](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/commit/65b32eaa61a4094cbc394b35dc201c16660b65c8)

Run it again because the linter is complaining.

juliocc (approved):

modules/cloudsql-instance/variables.tf
```diff
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
```
juliocc, line 16: This is empty line is what is causing the linter to fail. Revert the file and you'll be fine

modules/vpc-sc/variables.tf
LudovicEmo: @juliocc 

did some testing, @ludoo's solution or the below test 3 solution are needed.

**test 1:** 
- with == null
- without the coalesce

result is **KO**
analysis: terraform will still evaluate the second part after ||, as coalesce is not there it is rejected

test code:
    v.from.identity_type == null || contains([
          "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
          "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT"
        ], v.from.identity_type)
terraform result:
Error: v.from.identity_type is null

** test 2:**
- without == null,
- using the same coalesce:

result is **KO**
analysis: coalesce will replace null with "-", however "-" in not in the contains list

test code:
contains([
        "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
        "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT"
      ], coalesce(v.from.identity_type, "-"))

terraform result:
    Invalid `from.identity_type` value in egress policy.

**test 3:**
- without == null
- with coalesce changed with a trick using one of the permitted value when null

result is **ok**, however the trick seems too tricky :)

test code:
contains([
        "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
        "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT"
      ], coalesce(v.from.identity_type, "IDENTITY_TYPE_UNSPECIFIED"))

terraform result:
    No changes. Your infrastructure matches the configuration.

**proposal:**
stick with ludoo's initial solution which is still tricky but less :)

    v.from.identity_type == null || contains([
          "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
          "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT"
        ], coalesce(v.from.identity_type, "-"))

juliocc: you win :)

modules/cloudsql-instance/variables.tf
LudovicEmo: oh
done

ludoo:
Thank you for being thorough and making our job easy! :)

PR #1636, opened 2023-08-29
Delete api gateway blueprint

juliodiez:
API Gateway is not recommended anymore. The direction is towards Apigee.


juliocc (approved):
LGTM.

I think there's a broken link in the changelog.

juliodiez:
Yes, but isn't it auto-generated? Should I do anything? It's a link that existed at the time of the recorded change...

juliocc:
I think you can just delete it manually

PR #1645, opened 2023-09-03
gke-cluster-autopilot: add validation for release_channel input variable

olliefr:
This PR adds validation for the `release_channel` input variable. This is to align the code with the requirements for Autopilot clusters and to make the module more resilient to invalid input.

Relevant Google Cloud docs:

* [Clusters created in the Autopilot mode are pre-configured to use release channels](https://cloud.google.com/kubernetes-engine/docs/concepts/types-of-clusters#version-choices).
* [For Autopilot clusters, you cannot unsubscribe from a release channel](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels#unsubscribing_from_a_release_channel).

The validation should not have any adverse effect on the rest of the codebase because the input variable name or type has not changed and the default value is provided. The default value I picked is the same value that the API uses if no value is specified. It is also the same value that is used by Cloud Console's form to create a new Autopilot cluster.

I also added a bit of context to the input variable's documentation string.

---
**Checklist**


ludoo (approved):
Thank you for this PR!

PR #1646, opened 2023-09-03
gke-cluster-autopilot: add monitoring configuration

olliefr:
Although one cannot disable Cloud Monitoring integration in GKE Autopilot clusters, one has some flexibility over which control plane components' metrics should be ingested: [manage GKE metrics](https://cloud.google.com/stackdriver/docs/solutions/gke/managing-metrics).

This PR adds a new input variable `monitoring_config` and a block to the cluster resource to facilitate that. The variable is optional with a sensible default value.

The interface is similar in style to #1625. Since there was no monitoring configuration in place for this module previously, it is not likely for this PR to break any existing code 🙏 

I've added an example to the README as well.

Feedback is welcome 🙃

**Checklist**

### DIFF
--- a/modules/gke-cluster-autopilot/README.md
+++ b/modules/gke-cluster-autopilot/README.md
@@ -87,6 +87,35 @@ module "cluster-1" {
 # tftest modules=1 resources=1 inventory=logging-config.yaml
 ```
 
+### Monitoring configuration
+
+> **Note**
+> [Cloud Monitoring integration](https://cloud.google.com/stackdriver/docs/solutions/gke) is pre-configured and cannot be disabled for Autopilot clusters. GKE **system metrics** collection is always on and cannot be disabled.
+
+> **Warning**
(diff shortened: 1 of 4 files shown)


juliocc (approved):
@olliefr thanks for this!

Does it make sense to have the same variable for standard clusters?

olliefr:
@juliocc the standard cluster module already has it: [`monitoring_config`](https://github.com/devil-mice-labs/cloud-foundation-fabric/blob/42ebbccad9ffb325a1dec3e225011982487f1f66/modules/gke-cluster-standard/variables.tf#L199).

Like the `logging_config` before, it's not an object interface though. My plan is to implement the object interface for the standard clusters module as well but I'd like to do it in a separate PR because the surface of work is much greater for the standard clusters - the configuration is more flexible for standard clusters and there are existing uses of the module, just like we had with logging.

juliocc:
> @juliocc the standard cluster module already has it: [`monitoring_config`](https://github.com/devil-mice-labs/cloud-foundation-fabric/blob/42ebbccad9ffb325a1dec3e225011982487f1f66/modules/gke-cluster-standard/variables.tf#L199).
> 
> Like the `logging_config` before, it's not an object interface though. My plan is to implement the object interface for the standard clusters module as well but I'd like to do it in a separate PR because the surface of work is much greater for the standard clusters - the configuration is more flexible for standard clusters and there are existing uses of the module, just like we had with logging.

Sounds good. You can merge this one then.

PR #1648, opened 2023-09-05
add apigee non-vpc peering mode support

g-greatdevaks:
Apigee recently introduced the [Private Service Connect (Non-VPC Peering) Provisioning Mode](https://cloud.google.com/apigee/docs/api-platform/get-started/networking-options#non-vpc-peering-architecture-overview) for Apigee X provisioning; the feature is in Preview; as of the time of raising this PR. When using this provisioning mode, there is no requirement for holding any PSA IP CIDR Ranges in the Authorized Network.

- This PR focuses introducing the support for [Private Service Connect (Non-VPC Peering) Provisioning Mode](https://cloud.google.com/apigee/docs/api-platform/get-started/networking-options#non-vpc-peering-architecture-overview) Provisioning Mode for Apigee X provisioning.
- The PR also fixes a lifecycle issue with Apigee Environment resource. The issue was occurring because `try` function was being used for setting the name of the Apigee Environment. `lifecycle ignore_changes` configuration was added to overcome the issue. The issue force replaces the Apigee Environment and the same, if not fixed, will end up causing outage on each `terraform apply`.
- The Apigee module has been tested for both VPC Peering and Non-VPC Peering mode before raising the PR.

---
**Checklist**
I acknowledge that I have:


juliocc:
this is marked as fract. Is this ready for review?

g-greatdevaks:
@juliocc Fixed broken tests and added some more. Kindly review.

g-greatdevaks:
There seems to be some issue with one test case.
Will work on it and close in some time.

g-greatdevaks:
The PR is ready for review, Julio 🙂

modules/apigee/main.tf
```diff
@@ -23,11 +23,12 @@ resource "google_apigee_organization" "organization" {
   count                                = var.organization == null ? 0 : 1
   analytics_region                     = var.organization.analytics_region
   project_id                           = var.project_id
-  authorized_network                   = var.organization.authorized_network
+  authorized_network                   = var.organization.disable_vpc_peering ? null : var.organization.authorized_network
```
juliocc, line 26: Instead of changing the value here I'd add validation in the variable ensuring `var.organization.authorized_network = null` when `var.organization.disable_vpc_peering` is truee

modules/apigee/main.tf
```diff
@@ -91,7 +92,7 @@ resource "google_apigee_instance" "instances" {
   description              = each.value.description
   location                 = each.key
   org_id                   = local.org_id
-  ip_range                 = "${each.value.runtime_ip_cidr_range},${each.value.troubleshooting_ip_cidr_range}"
+  ip_range                 = var.organization.disable_vpc_peering ? null : "${each.value.runtime_ip_cidr_range},${each.value.troubleshooting_ip_cidr_range}"
```
juliocc, line 95: Same as before, I'd add a validation to variable (or maybe here you need a precondition check since it's s different variable)

modules/apigee/main.tf
```diff
@@ -117,6 +118,9 @@ resource "google_apigee_instance_attachment" "instance_attachments" {
   instance_id = google_apigee_instance.instances[each.value.region].id
   environment = try(google_apigee_environment.environments[each.value.environment].name,
   "${local.org_id}/environments/${each.value.environment}")
+  lifecycle {
```
juliocc, line 121: why is `ignore_changes` needed here?

juliocc (changes requested):

modules/apigee/main.tf
g-greatdevaks: There is an issue which is occurring because `try` function is being used for setting the name of the Apigee Environment. Because of this, the actual value for the `environment` argument cannot be determined until runtime and `terraform plan` shows `force replaces` comment for corresponding Apigee Environment. `terraform apply` force replaces the resource as well. If the same is not fixed, it will end up causing outage on each `terraform apply`. 

`lifecycle ignore_changes` configuration has been added to overcome the issue.

wiktorn: I had some doubts about this too:
- why do we need `try` here (to make `destroy` work? because otherwise it should always evaluate properly)
- is indeed `try` the issue here, and not the `for_each` that does the transformations

But in the end it will work as expected, because if you change the environment name, you change the key in the resource, thus forcing replacement. 

I'd love to have a comment here in the code about one and the other (so save my future-self from thinking this over again)

modules/apigee/README.md
```diff
@@ -135,14 +135,24 @@ module "apigee" {
 module "apigee" {
   source     = "./fabric/modules/apigee"
   project_id = "my-project"
+  organization = {
```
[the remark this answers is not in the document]
g-greatdevaks: Added.

tests/modules/apigee/tftest.yaml
```diff
@@ -21,7 +21,9 @@ tests:
   env_only_with_api_proxy_type:
   env_only_with_deployment_type:
   envgroup_only:
-  instance_only:
+  instance_only_psc_mode:
```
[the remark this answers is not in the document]
g-greatdevaks: Added.

modules/apigee/main.tf
g-greatdevaks: The `try` function is actually not even required here as if `var.environments` is `{}` then the instance attachment resource won't even be getting created.

And on another note, `environment` attribute requires the name of the Apigee Environment and not the id.
So usage of `"${local.org_id}/environments/${each.value.environment}"` is not right and instead it should just be `each.value.environment`. However, since we don't require the `try` function here, `environment = google_apigee_environment.environments[each.value.environment].name` should be sufficient.

g-greatdevaks:
Closing the PR as an alternate implementation has been worked on: [#1653](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1653)

PR #1654, opened 2023-09-07
Fix project factory blueprint and fast stage

LucaPrete:
The PR:
* Adds the variable parent to the recently refactored project factory blueprint, so the value can be correctly passed to the project module
* Modifies the `project.yaml.sample` template file adding the parent key (as opposed to the folder key which seems to map to nothing at the moment)
* Adds the possibility to temporarily set billing id to null in the project_factory blueprint, in order to use project factory internally

### DIFF
--- a/blueprints/factories/project-factory/factory.tf
+++ b/blueprints/factories/project-factory/factory.tf
@@ -28,11 +28,11 @@ locals {
   )
   projects = {
     for k, v in local._data : k => merge(v, {
-      billing_account = coalesce(
+      billing_account = try(coalesce(
         var.data_overrides.billing_account,
         try(v.billing_account, null),
         var.data_defaults.billing_account
-      )
+      ), null)
       contacts = coalesce(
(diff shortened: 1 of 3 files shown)


ludoo (approved):

blueprints/factories/project-factory/factory.tf
```diff
@@ -28,11 +28,11 @@ locals {
   )
   projects = {
     for k, v in local._data : k => merge(v, {
-      billing_account = coalesce(
+      billing_account = try(coalesce(
```
juliocc, line 31: Isn't this the same as 
```try(var.data_overrides.billing_account,
       v.billing_account,
       var.data_defaults.billing_account,
       null)
```

juliocc (approved):

blueprints/factories/project-factory/factory.tf
juliocc: It's not

PR #1659, opened 2023-09-08
Link project factory documentation from FAST stage

### LINKED ISSUE #1628: stages: document how the new `factory_data` is meant to be used
Previously I just had to `terraform apply` with a `defaults.yaml` inside `./data/` and each project yaml inside `./data/projects`, which has been working as expected. Until this PR was merged:
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1595

Now I'm specifying the path of the projects yaml as:

```ini
factory_data = {
  data_path = "./data/projects"
}
```

But after doing a `terraform apply` the plan shows it will remove all the previous configurations.

ludoo:
Fixes #1628


juliocc (approved):

gustavovalverde:
I don't think this solves the problem if someone uses the available sample file in the stage:
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/fast/stages/3-project-factory/dev/data/projects/project.yaml.sample

Most people would commonly adapt the sample to their use case, but if the sample has the wrong data types, then it will crash for newcomers, unless they start digging into each dependency and their documentation. 

I'm using that same file structure, as shared here:
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/1628#issuecomment-1709762456

I would say adapting the file would be the missing part, as done here:
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/commit/12e78af05502812b4f39e8c3f05193f6d4a6d60d#diff-8ad613bbda56823204e080e67cdab397ff19531aecb7c1a3faa06c74858f2899

ludoo:
Good point, let me nuke the stale file out of existence. :)

PR #1666, opened 2023-09-11
added support for global proxy only subnets

dgulli:
To support the new cross-regional load balancers, i need to be able to create global managed proxy subnet types, rather than regional, as per the doco 

[here](https://cloud.google.com/load-balancing/docs/l7-internal/setting-up-l7-cross-reg-internal#configure-a-network) and 
[here](https://cloud.google.com/load-balancing/docs/proxy-only-subnets#envoy-lb)

---
**Checklist**


modules/net-vpc/subnets.tf
```diff
@@ -79,6 +79,10 @@ locals {
     { for s in var.subnets_psc : "${s.region}/${s.name}" => s },
     { for k, v in local._factory_subnets : k => v if v.purpose == "PRIVATE_SERVICE_CONNECT" }
   )
+  subnets_global_proxy_only = merge(
+    { for s in var.subnets_psc : "${s.region}/${s.name}" => s },
```
wiktorn, line 83: I guess you wanted to introduce new variable here, instead reuse `var.subnets_psc`?

ludoo:
I think we should internalize the subnet type in the subnet structure instead of having a ton of variables, which was the legacy way of doing this before `optional()`. This is ok for now, just keep in mind we will probably change the whole interface soonish.

modules/net-vpc/subnets.tf
alloveras: Also, couldn't we re-use `var.subnets_proxy_only` and have the filtering predicate have something like `contains(v.purpose, ["REGIONAL_MANAGED_PROXY", "GLOBAL_MANAGED_PROXY"])`?

modules/net-vpc/variables.tf
```diff
@@ -227,6 +227,19 @@ variable "subnets" {
   nullable = false
 }
 
+variable "subnets_global_proxy_only" {
```
juliocc, line 230: Do we really need a new variable for this? I'd rather reuse the existing `subnets_proxy_only` and add a new `global` attribute with type `optional(boolean, FALSE)`.

juliocc (changes requested):
Left a comment below. Also, I'd like to see at least one test exercising the new type of subnet

modules/net-vpc/variables.tf
dgulli: That's fair - I'll have a read how to implement optional into the existing proxy subnet.

dgulli: @juliocc  - i've refactored to use the existing proxy_only subnet, and removed all the other work done to reflect. Introduced test to create the additional subnet and regenerated the doco

modules/net-vpc/subnets.tf
```diff
@@ -130,10 +130,10 @@ resource "google_compute_subnetwork" "proxy_only" {
   ip_cidr_range = each.value.ip_cidr_range
   description = (
     each.value.description == null
-    ? "Terraform-managed proxy-only subnet for Regional HTTPS or Internal HTTPS LB."
+    ? "Terraform-managed proxy-only subnet for Regional HTTPS, Internal HTTPS or Cross-Regional HTTPS Internal LB."
     : each.value.description
   )
-  purpose = "REGIONAL_MANAGED_PROXY"
+  purpose = each.value.global != false ? "GLOBAL_MANAGED_PROXY" : "REGIONAL_MANAGED_PROXY"
```
wiktorn, line 136: This can be rewritten as:
`each.value.global ? "GLOBAL_MANAGED_PROXY" : "REGIONAL_MANAGED_PROXY"`

which is shorter

modules/net-vpc/subnets.tf
```diff
@@ -33,7 +33,7 @@ locals {
       secondary_ip_ranges   = try(v.secondary_ip_ranges, null)
       iam                   = try(v.iam, [])
       iam_members           = try(v.iam_members, [])
-      purpose               = try(v.purpose, null)
+      purpose               = try(v.global, null)
```
wiktorn, line 36: This part refers to YAML file parsing, for backward compatibility please keep `v.purpose` here.

As you are adding `purpose: GLOBAL_MANAGED_PROXY` you need to add specific handling below, when creating `subnets_proxy_only` from `local._factory_subnets`. You need to add `active` true/false depending on type of what is the purpose (global or regional proxy)

wiktorn (changes requested):

modules/net-vpc/subnets.tf
dgulli: thanks for that - i find that the way i've put this is a little easier for read for peeps without deep TF experience!

modules/net-vpc/subnets.tf
```diff
@@ -130,11 +132,16 @@ resource "google_compute_subnetwork" "proxy_only" {
...
+    ? "Terraform-managed proxy-only subnet for Regional HTTPS, Internal HTTPS or Cross-Regional HTTPS Internal LB."
     : each.value.description
   )
-  purpose = "REGIONAL_MANAGED_PROXY"
-  role    = each.value.active != false ? "ACTIVE" : "BACKUP"
+  purpose = try(
+    each.value.purpose,
+    each.value.global != false
+    ? "GLOBAL_MANAGED_PROXY"
+  : "REGIONAL_MANAGED_PROXY")
```
ludoo, line 142: nit: can you add a line break between `"REGIONAL_MANAGED_PROXY"` and the following `)` then run `terraform ftm`?

ludoo (approved):

modules/net-vpc/subnets.tf
dgulli: done!

ludoo:
hit the merge button once all is green, thanks for this!

ludoo:
nit:

```hcl
each.value.global != false
    ? "GLOBAL_MANAGED_PROXY"
    : "REGIONAL_MANAGED_PROXY"
```

will default to global if `each.value.global` is `null`, which is the opposite of the default in the variable type. I would remove the `!= false` or change it to `== true`. We can do this after merging.

dgulli:
> nit:
> 
> ```terraform
> each.value.global != false
>     ? "GLOBAL_MANAGED_PROXY"
>     : "REGIONAL_MANAGED_PROXY"
> ```
> 
> will default to global if `each.value.global` is `null`, which is the opposite of the default in the variable type. I would remove the `!= false` or change it to `== true`. We can do this after merging.

done - https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1676

PR #1675, opened 2023-09-14
GKE Autopilot module: add network tags

olliefr:
This PR adds support for [network tags](https://cloud.google.com/vpc/docs/add-remove-network-tags) to [`gke-cluster-autopilot`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/e19f332dcf91edd6b75d3beed8f5cf6b3dd0657e/modules/gke-cluster-autopilot) module.

There was a `nullable` input variable `tags` defined before I started but for some reason it was not used by the module.

In addition to making use of the variable, I made it non-nullable and added validation as well.

## Potential impact area

Estimated risk: minimal.

* I did not change the name or the type of the variable. 
* It was previously declared but was not being used so there's unlikely to be much code out there that depends on it.

## Feedback wanted 

I'm still new to Fabric so please don't hesitate to tell me if my validation is excessive 🥲 and I will trim it.

Another question I have is about `node_pool_auto_config` block in `google_container_cluster`. The Terraform resource docs say it's in "beta" although I can see the corresponding JSON block not only in [v1beta](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#nodepoolautoconfig) API but also in [v1](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#nodepoolautoconfig). Is it just the case of Terraform Google provider lagging behind the release cadence of the API or am I missing something more fundamental here? From a practical point of view, this module was already using `google-beta` provider so it's not really an issue, I'm just curious.

Lastly, I have a `FIXME` comment in the code where I had to use a `dynamic` block to work around a glitch in diff-ing the empty list of network tags. My understanding is that the API does allow an empty list but Terraform somehow can't cope with it and considers it to be a diff. My question is should I proceed with the `dynamic` block or should I try to report this upstream into the provider so that this unnecessary empty diff does not happen?

I hope this makes sense. Any other feedback is welcome as well... :shipit: 

**Checklist**

I acknowledge that I have:


olliefr:
Oh, one last thing! 

I do wonder if the input variable should be called `network_tags` rather than `tags` to avoid confusion with [Tags](https://cloud.google.com/kubernetes-engine/docs/how-to/tags)? In my line of work I often come across people who are confused by this dual use of the term "tag" on Google Cloud so I thought maybe it's an opportunity to make it a bit easier for them. Please let me know what you think...

olliefr:
Spotted a few references to GKE Standard clusters in the README. Fixed that and expanded Backup for GKE example section. Added a TOC as well as the README is getting longer

modules/gke-cluster-autopilot/README.md
```diff
@@ -1,10 +1,23 @@
-# GKE cluster Autopilot  module
+# GKE Autopilot cluster module
 
-This module allows simplified creation and management of GKE Autopilot clusters. Some sensible defaults are set initially, in order to allow less verbose usage for most use cases.
+This module offers an opinionated way to create and manage Google Kubernetes Engine (GKE) [Autopilot clusters](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview). With its sensible default settings based on best practices and authors' experience as Google Cloud practitioners, the module accommodates for many common use cases out-of-the-box, without having to rely on verbose configuration.
```
ludoo, line 3: nit: we try not to be opinionated in modules, especially compared to other GCP TF modules who tweak the cluster after creation :)

modules/gke-cluster-autopilot/variables.tf
```diff
@@ -216,9 +216,34 @@ variable "service_account" {
 }
 
 variable "tags" {
-  description = "Network tags applied to nodes."
+  description = "Applies the specified network tags to all nodes in this Autopilot cluster. Network tags are metadata on Compute Engine virtual machines (VMs) that allow you to make firewall rules and routes applicable to specific VM instances. In GKE, you can use network tags to make firewall rules or routes applicable to the nodes in your cluster."
   type        = list(string)
-  default     = null
+  default     = []
+  nullable    = false
+  validation {
```
ludoo, line 223: We tend to only validate in specific cases, for example when a string field is really an enum, or when an object configures a resource via top-level attributes and only one is needed (e.g. healthchecks).

As a further example of our approach, we tend not to validate cases where the API returns a clear error. For example with disallowed characters in strings, length of strings, etc.

While I'm not entirely against this approach (it does make the variable files harder to parse though), it's way overkill vs our standard practices, and if we decide to follow this path we need to do it consistently across all modules.

I would roll this back. :)

modules/gke-cluster-autopilot/variables.tf
```diff
@@ -216,9 +216,34 @@ variable "service_account" {
 }
 
 variable "tags" {
-  description = "Network tags applied to nodes."
+  description = "Applies the specified network tags to all nodes in this Autopilot cluster. Network tags are metadata on Compute Engine virtual machines (VMs) that allow you to make firewall rules and routes applicable to specific VM instances. In GKE, you can use network tags to make firewall rules or routes applicable to the nodes in your cluster."
```
ludoo, line 219: This description is too long, we assume people know what the attributes are and only expand when the type might not be obvious.

ludoo (changes requested):
Very nice, thanks for this. A couple areas are a bit overkill vs our traditional approach. README changes are great!

modules/gke-cluster-autopilot/README.md
olliefr: Understood. I will rollback this change

modules/gke-cluster-autopilot/variables.tf
olliefr: OK, will rollback 🙂

modules/gke-cluster-autopilot/variables.tf
olliefr: Validation was removed as requested. Conversation resolved :shipit:

modules/gke-cluster-autopilot/variables.tf
olliefr: Description was rolled back as requested. Item resolved 🎉

modules/gke-cluster-autopilot/README.md
olliefr: Documentation updated to avoid ambiguous/unwanted wording

olliefr:
Added a `tftest` because I don't think that network tags warrant their own example in README.

ludoo (approved):
thanks :)

PR #1690, opened 2023-09-18
Rename instance attachment to match versions 23 and earlier

cygnus8595:
In versions 23 and earlier, Apigee instance attachments were named <region>-<environment>. Later versions swapped that to <environment>-<region>, which will cause downtime when upgrading to the latest module as the attachments would have  to be destroyed and re-created.

This PR just swaps the names back to how they were to avoid causing downtime for just a name.
---
**Checklist**


modules/apigee/variables.tf
```diff
@@ -66,11 +66,13 @@ variable "instances" {
   type = map(object({
     name                          = optional(string)
     display_name                  = optional(string)
+    name                          = optional(string)
     description                   = optional(string, "Terraform-managed")
     runtime_ip_cidr_range         = optional(string)
     troubleshooting_ip_cidr_range = optional(string)
     disk_encryption_key           = optional(string)
     consumer_accept_list          = optional(list(string))
+    environments                  = optional(list(string))
```
juliocc, line 75: Duplicate

modules/apigee/variables.tf
```diff
@@ -66,11 +66,13 @@ variable "instances" {
   type = map(object({
     name                          = optional(string)
     display_name                  = optional(string)
+    name                          = optional(string)
```
juliocc, line 69: Duplicate

juliocc (dismissed):
@apichick can you take a look?

modules/apigee/variables.tf
cygnus8595: Sorry, must have got duplicated in a merge. I was working on this simultaneously.

modules/apigee/variables.tf
cygnus8595: Same as above, merge issue.

apichick (approved):

PR #1692, opened 2023-09-19
Allow using no service account in compute-vm

### LINKED ISSUE #1691: feat: Need to make compute-vm service_account configuration "optional"
Hello Team,

we need to have the capabilit of disable service_account configuration when creating a compute-vm as our security policies recognize it as a threat. Previously we used an ad-hoc module with a dynamic block for the service_account, can you implement this in your fabric [module](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/compute-vm/main.tf )?

Thank you.

ludoo:
This also streamlines handling of service account in the `compute-vm` module.

Fixes #1691


modules/compute-vm/variables.tf
```diff
@@ -292,9 +292,13 @@ variable "scratch_disks" {
...
+  description = "Service account email and scopes. If email is null, the default Compute service account will be used unless auto_create is true, in which case a service account will be created."
+  type = object({
+    auto_create = optional(bool, false)
+    email       = optional(string)
+    scopes      = optional(list(string))
+  })
+  default = {}
 }
 
 variable "service_account_create" {
```
wiktorn, line 304: Should `service_account_create` and `service_account_scopes`  be removed?

ludoo: Gah, good catch.

modules/compute-vm/variables.tf
```diff
@@ -292,23 +292,13 @@ variable "scratch_disks" {
...
-  default     = false
-}
-
-# scopes and scope aliases list
-# https://cloud.google.com/sdk/gcloud/reference/compute/instances/create#--scopes
-variable "service_account_scopes" {
-  description = "Scopes applied to service account."
-  type        = list(string)
-  default     = []
+  description = "Service account email and scopes. If email is null, the default Compute service account will be used unless auto_create is true, in which case a service account will be created."
```
wiktorn, line 295: Maybe worth adding here: `set to null, if you don't want any service account attached to the instance`?

ludoo: Done.

wiktorn (approved):

PR #1694, opened 2023-09-19
Fix apigee addons config conditional expression

eddern:
I assume the intended behavior of the `addons_config` in the Apigee module is to enable the addons that are set to `true`?

As it is now, I need to pass this in order to enable connectors and integration:
```
addons_config = {
    connectors_platform = false
    integration         = false
    api_security = true
    monetization = true
    advanced_api_ops = true
 }
```

---
**Checklist**


modules/apigee/main.tf
```diff
@@ -137,31 +137,31 @@ resource "google_apigee_addons_config" "addons_config" {
   org      = local.org_name
   addons_config {
     dynamic "advanced_api_ops_config" {
-      for_each = var.addons_config.advanced_api_ops ? [] : [""]
+      for_each = var.addons_config.advanced_api_ops ? [""] : []
       content {
         enabled = true
```
juliocc, line 142: can we us the actual value here? 

The following should be enough:
```hcl
      advanced_api_ops_config {
        enabled = var.addons_config.advanced_api_ops
      }
```

And same for all the other options.

juliocc (changes requested):

modules/apigee/main.tf
eddern: Sure thing. 

Much cleaner like you suggested 👍

juliocc (approved):

eddern:
@juliocc after upgrading my Apigee module to the newest version, terraform keeps wanting to update the values to false every time I plan (even though I have applied several times). Any idea why this is?

Screenshot 2023-10-23 at 09 20 06

eddern:
The same applies with all the values explicitly defined in the config as well
Screenshot 2023-10-23 at 09 31 10

juliocc:
Looks like my suggestion of using the value directly in the configuration block doesn't work. Apparently the provider is only expecting values to be true or nothing.

Do you mind reverting to the previous syntax.

mwarm2:
See #1798

PR #1698, opened 2023-09-21
FAST: move organization policies to stage 0

ludoo:
This implements the changes described [in this ADR](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/ludo/fast-bootstrap-resman-refactor/fast/docs/0-org-policies.md) and is now ready for review.


fast/stages/0-bootstrap/variables.tf
```diff
@@ -97,6 +97,14 @@ variable "custom_roles" {
   default     = {}
 }
 
+variable "factories_config" {
+  description = "Configuration for the organization policies factory."
+  type = object({
+    data_dir = optional(string, "data/org-policies")
```
juliocc, line 103: Rename the key data_dir to org_policies?

fast/stages/0-bootstrap/organization.tf
```diff
@@ -45,6 +45,11 @@ locals {
       }
     ]
   ])
+  drs_domains = concat(
+    [var.organization.customer_id],
+    var.org_policies_config.constraints.allowed_policy_member_domains
```
juliocc, line 50: `var.org_policies_config` can be null. Either check for null values or (better) make the variable non-nullable?

fast/stages/0-bootstrap/organization.tf
```diff
@@ -151,4 +156,40 @@ module "organization" {
       type                 = attrs.type
     }
   }
+  org_policies_data_path = var.factories_config.data_dir
+  org_policies = {
+    "iam.allowedPolicyMemberDomains" = {
+      rules = [
+        {
```
juliocc, line 163: `var.factories_config` can be null. Either check for null values or (better) make the variable non-nullable?

fast/stages/1-resman/variables.tf
```diff
@@ -189,6 +183,18 @@ variable "locations" {
   nullable = false
 }
 
+variable "org_policy_tags" {
```
juliocc, line 186: should this be `org_policies_tags`? I'm not sure.

fast/stages/1-resman/variables.tf
```diff
@@ -227,16 +225,14 @@ variable "prefix" {
...
+    tenant      = string
   })
   default = {
-    context      = "context"
-    environment  = "environment"
-    org-policies = "org-policies"
-    tenant       = "tenant"
+    context     = "context"
+    environment = "environment"
+    tenant      = "tenant"
```
juliocc, line 235: I'd move the default values to the fields
```hcl
type = object({
    context     = optional(string, "context")
    environment = optional(string, "environment")
    tenant      = optional(string, "tenant")
  })
  default = {}
  nullable = false
```

juliocc (commented):
good stuff.

fast/stages/0-bootstrap/variables.tf
ludoo: I was trying to be consistent with other stages, but yes it makes sense and we might want to eventuallys peciali members of that variable for other stages too (eg `factories_config.subnets`)

fast/stages/1-resman/variables.tf
ludoo: I'm never sure about those things, let's look at an English language rules reference once and for all...

fast/stages/0-bootstrap/organization.tf
ludoo: added `nullable == false`, and users can provide an empty dir (or set `data_dir` to null the module would survive)

fast/stages/1-resman/variables.tf
juliocc: actually as it is right now you have to provide all the values if you want to change a single one.

fast/stages/1-resman/variables.tf
ludoo: seems like the current version is ok

https://grammar.collinsdictionary.com/easy-learning/how-do-you-form-the-plural-of-compound-nouns-in-english#:~:text=Grammar%20%7C%20Collins%20Education-,How%20do%20you%20form%20the%20plural%20of%20compound%20nouns%20in,last%20word%20of%20the%20compound.

fast/stages/1-resman/variables.tf
ludoo: gosh, swapped your code in the comment with my code, let me fix great suggestion :)

juliocc (approved):

PR #1709, opened 2023-09-28
Add bug fix in bucket local variable

luigi-bitonti:
As you suggested, I made the change only to the local variable.
With this change, "bucket" variable will have the value of "var.bucket_name" only if the bucket has to be created.
Otherwise, it will have the name of the created bucket.
---
**Checklist**

### DIFF
--- a/modules/cloud-function-v1/README.md
+++ b/modules/cloud-function-v1/README.md
@@ -95,6 +95,7 @@ You can have the module auto-create the GCS bucket used for deployment via the `
 module "cf-http" {
   source      = "./fabric/modules/cloud-function-v1"
   project_id  = "my-project"
+  prefix      = "my-prefix"
   name        = "test-cf-http"
   bucket_name = "test-cf-bundles"
   bucket_config = {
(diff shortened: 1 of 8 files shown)


wiktorn:
Oh, now I see the bug :-) Thanks a lot.

Let me update this a bit

wiktorn:
Can you pull changes from [here](https://github.com/wiktorn/cloud-foundation-fabric/tree/luigi_bug_fix_cf) ? This adds tests that verify that the bug is actually fixed and also covers cloud functions v1.

luigi-bitonti:
I did it. I don't know if it was the correct way to merge your change, but trying to merge all the branch a I saw that there are a lots file changed because your branch is is 81 commits behind GoogleCloudPlatform:master.
So I basically copied and pasted only the file changed on your last commit.

wiktorn (approved):

wiktorn:
Thank you, I thought that I based that on your branch, but maybe I missed something. `git cherry-pick` should also work in this case.

Issue #1713, opened 2023-09-29
Add support for passing multiple protocols L3_DEFAULT through net-ilb-int


LucaPrete:
This new features requires setting different ip_protocol for the forwarding rule and the backend service.

PR #1714, opened 2023-09-29
[#1713] Support multiple protocols (L3_DEFAULT) through net-ilb-in

LucaPrete:
Relates to #1713. Adds support for L3_DEFAULT for net-lb-int

---
**Checklist**

### DIFF
--- a/modules/net-lb-int/main.tf
+++ b/modules/net-lb-int/main.tf
@@ -32,7 +32,7 @@ resource "google_compute_forwarding_rule" "default" {
   name        = var.name
   description = var.description
   ip_address  = var.address
-  ip_protocol = var.protocol # TCP | UDP
+  ip_protocol = var.protocol
   backend_service = (
     google_compute_region_backend_service.default.self_link
   )
@@ -54,7 +54,7 @@ resource "google_compute_region_backend_service" "default" {
   name                            = var.name
   description                     = var.description
   load_balancing_scheme           = "INTERNAL"
-  protocol                        = var.protocol
+  protocol                        = var.backend_service_config.protocol
   network                         = var.vpc_config.network
   health_checks                   = [local.health_check]
   connection_draining_timeout_sec = var.backend_service_config.connection_draining_timeout_sec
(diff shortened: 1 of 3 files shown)


juliocc (approved):

modules/net-lb-int/main.tf
```diff
@@ -54,7 +54,7 @@ resource "google_compute_region_backend_service" "default" {
   name                            = var.name
   description                     = var.description
   load_balancing_scheme           = "INTERNAL"
-  protocol                        = var.protocol
+  protocol                        = var.backend_service_config.protocol
```
wiktorn, line 57: Can we just specify here `UNSPECIFIED` as per [docs](https://cloud.google.com/load-balancing/docs/internal#forwarding-rule-protocols) always?

We need anyway to have agreement between protocols here, so I'm reluctant to extend interface scope if the value has to match the other value that user is providing. 

If `UNSPECIFIED` will not work, then maybe just still use `var.protocol`, but in case its value is `L3_DEFAULT` replace that with `UNSPECIFIED`?

LucaPrete: Yep. I'd go with the second option to facilitate life of users.

LucaPrete: Actually, let me take it back after double checking the docs. Let's modify it directly to UNSPECIFIED. Good point. thanks!

wiktorn (dismissed):

modules/net-lb-int/variables.tf
```diff
@@ -36,6 +36,7 @@ variable "backend_service_config" {
       ratio                     = optional(number)
     }))
     log_sample_rate  = optional(number)
+    protocol         = optional(string, "UNSPECIFIED")
```
ludoo, line 39: nit: if UNSPECIFIED is the default in the provider, I would use null as the default in the optional

ludoo (approved):

ludoo:
@LucaPrete do you need this to stay open?

LucaPrete:
I thought it was merged!
Rebasing and hopefully merging soon.

PR #1718, opened 2023-09-30
FAST: add example of custom org policy condition to bootstrap README

ludoo:
This PR adds an example that shows how to conditionally set the oslogin org policy via a custom tag value. It's needed by the data platform, and also demonstrates how to define custom tag values and rules.

@lcaggio use this in your DP setup

### DIFF
--- a/fast/stages/0-bootstrap/README.md
+++ b/fast/stages/0-bootstrap/README.md
@@ -14,11 +14,11 @@ Use the following diagram as a simple high level reference for the following sec
   <img src="diagram.svg" alt="Organization-level diagram">
 </p>
 
-## Table of contents
-
+<!-- BEGIN TOC -->
 - [Design overview and choices](#design-overview-and-choices)
   - [User groups](#user-groups)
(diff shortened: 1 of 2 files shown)


fast/stages/0-bootstrap/README.md
```diff
@@ -76,6 +84,38 @@ Management of the rest of the tag hierarchy is delegated to the resource managem
...
+      description = "Bind this tag to set oslogin to false."
+      iam = {
+        "roles/resourcemanager.tagUser" = [
+          "group:foo@example.com"
+        ]
+      }
+    }
+  }
+}
+# tftest skip
```
juliocc, line 105: nit: you use "tfvars" for the language and you can remove the tftest directive.

juliocc (approved):

fast/stages/0-bootstrap/README.md
ludoo: let's fix it after :)

PR #1726, opened 2023-10-03
Add materialized views for bigquery

devuonocar:
- Added field and partition filter to time_partitioning for bq table
- Added Materialized view
---
**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,20 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 <!-- None < 2023-09-18 07:03:09+00:00 -->
 
+## [27.0.0] - 2023-10-04
+
 ### BLUEPRINTS
 
+- [[#1730](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1730)] Minimal Data Platform - Fix ([lcaggio](https://github.com/lcaggio)) <!-- 2023-10-04 10:15:51+00:00 -->
+- [[#1725](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1725)] Fix data platform roles ([lcaggio](https://github.com/lcaggio)) <!-- 2023-10-04 05:31:41+00:00 -->
+- [[#1724](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1724)] Bump provider versions to v5.0.0 ([ludoo](https://github.com/ludoo)) <!-- 2023-10-03 12:15:36+00:00 -->
 - [[#1722](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1722)] Add support for org policies to project factory ([ludoo](https://github.com/ludoo)) <!-- 2023-10-02 14:13:57+00:00 -->
 - [[#1692](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1692)] **incompatible change:** Allow using no service account in compute-vm ([ludoo](https://github.com/ludoo)) <!-- 2023-09-19 16:56:51+00:00 -->
 
 ### DOCUMENTATION
 
+- [[#1725](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1725)] Fix data platform roles ([lcaggio](https://github.com/lcaggio)) <!-- 2023-10-04 05:31:41+00:00 -->
+- [[#1724](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1724)] Bump provider versions to v5.0.0 ([ludoo](https://github.com/ludoo)) <!-- 2023-10-03 12:15:36+00:00 -->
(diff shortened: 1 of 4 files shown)


wiktorn (approved):
Thanks a lot. It looks like you need to re-run `tfdoc`.

modules/bigquery-dataset/variables.tf
```diff
@@ -152,8 +152,10 @@ variable "tables" {
         start    = number
       }))
       time = optional(object({
-        expiration_ms = number
-        type          = string
+        expiration_ms            = optional(number)
+        field                    = optional(string)
+        type                     = string
```
ludoo, line 157: nit: since `type` is not optional, can you put it first in order?

modules/bigquery-dataset/main.tf
```diff
@@ -267,3 +268,46 @@ resource "google_bigquery_table" "views" {
     use_legacy_sql = each.value.use_legacy_sql
   }
 }
+
+resource "google_bigquery_table" "materialized_view" {
+  depends_on          = [google_bigquery_table.default]
+  for_each            = var.materialized_views
+  project             = var.project_id
+  dataset_id          = google_bigquery_dataset.default.dataset_id
+  table_id            = each.key
```
ludoo, line 277: can you use an implicit reference here to the table resource, and remove the explicit dependency above?

ludoo (changes requested):
Thanks for the nice PR. Dropped a couple comments.

modules/bigquery-dataset/main.tf
devuonocar: Shoul it be done also for the `views?` I followed the same approach.
Here the code: https://github.com/devuonocar/cloud-foundation-fabric/blob/97b8af8c17a836fc1e296454d6261893698293bf/modules/bigquery-dataset/main.tf#L260C8-L260C8

ludoo: Ah yes, we always prefer implicit dependencies

ludoo (approved):
thanks for the very nice PR!

ludoo:
You just need to rerun tfdoc

modules/bigquery-dataset/main.tf
```diff
@@ -244,19 +244,19 @@ resource "google_bigquery_table" "default" {
...
   }
 }
 
 resource "google_bigquery_table" "views" {
-  depends_on          = [google_bigquery_table.default]
   for_each            = var.views
   project             = var.project_id
   dataset_id          = google_bigquery_dataset.default.dataset_id
-  table_id            = each.key
+  table_id            = google_bigquery_table.default[each.key].table_id
```
wiktorn, line 259: Here should be ` google_bigquery_table.views[each.key].table_id`

ludoo: table_id should point to the table, what am I missing Wiktor?

devuonocar: IMO the test should be updated setting the table _Countries_ instead of _population_
- here: https://github.com/devuonocar/cloud-foundation-fabric/blob/8fc510f0334f1c05a7137a1a025a2f603b5c76ea/tests/modules/bigquery_dataset/examples/views.yaml#L28C13-L28C13
- here: https://github.com/devuonocar/cloud-foundation-fabric/blob/8fc510f0334f1c05a7137a1a025a2f603b5c76ea/tests/modules/bigquery_dataset/examples/views.yaml#L30

wiktorn: So my understanding of the intent was, that we are providing here `google_bigquery_table.default[each.key].table_id` to specify, that this view is using this table. But we do not specify which tables we use, other than query of the view itself.

Also, view can't have the same `table_id` as the table. They have to be unique.

devuonocar: Taking again a look to the code I think you're right. I'm going to update it

modules/bigquery-dataset/main.tf
```diff
@@ -256,7 +256,7 @@ resource "google_bigquery_table" "views" {
   for_each            = var.views
   project             = var.project_id
   dataset_id          = google_bigquery_dataset.default.dataset_id
-  table_id            = google_bigquery_table.default[each.key].table_id
+  table_id            = google_bigquery_table.views[each.key].table_id
```
wiktorn: This is circular dependency.

modules/bigquery-dataset/main.tf
wiktorn: I think there is no good solution to avoid `depends_on` here. The only solution I can imagine, is to reference the resources while creating a query, which in this case is impossible. 

So I think in this case, we have to live with `depends_on` all tables.

ludoo: Thanks for being thorough Wiktor! :)

devuonocar: Thanks both

ludoo (approved):

wiktorn (approved):

PR #1732, opened 2023-10-04
Added FQDN Network Policy feature on GKE Cluster

luigi-bitonti:
Added the possibily to enable FQDN Network Policy feature on standard GKE Cluster
---
**Checklist**


modules/gke-cluster-standard/README.md
```diff
@@ -55,7 +55,7 @@ module "cluster-1" {
 # tftest modules=1 resources=1 inventory=basic.yaml
 ```
 
-### Enable Dataplane V2
+### Enable Dataplane
```
juliocc, line 58: Why remove the "V2" here?

modules/gke-cluster-standard/main.tf
```diff
@@ -31,6 +31,7 @@ resource "google_container_cluster" "cluster" {
   enable_intranode_visibility = var.enable_features.intranode_visibility
   enable_l4_ilb_subsetting    = var.enable_features.l4_ilb_subsetting
   enable_shielded_nodes       = var.enable_features.shielded_nodes
+  enable_fqdn_network_policy  = var.enable_features.dataplane_v2 && var.enable_features.fqdn_network_policy ? true : false
```
juliocc, line 34: I'd simply pass the value without checking if if dataplane_v2 is set. Let the API fail if what the user specifies is inconsistent.

Alternatively add a validation directly in the variable.

juliocc (changes requested):

modules/gke-cluster-standard/README.md
luigi-bitonti: tfdoc.py removed it

juliocc: Are you sure it was tfdoc?

luigi-bitonti: Sorry, I didn't know what happened, I relaunched tfdoc.py and it worked properly.

danielmarzini:
Hey @luigi-bitonti thanks for the PR, can you also add the change on the autopilot-module?

luigi-bitonti:
> Hey @luigi-bitonti thanks for the PR, can you also add the change on the autopilot-module?

luigi-bitonti:
> Hey @luigi-bitonti thanks for the PR, can you also add the change on the autopilot-module?

Hi Daniel, I didn't do it because GCP doc (https://cloud.google.com/kubernetes-engine/docs/how-to/fqdn-network-policies#enable_fqdn_network_policy_in_a_new_cluster) said that "You can't enable this feature when you create a new Autopilot cluster", but only on an existing one.
Do you want the feature anyway?

danielmarzini:
> > Hey @luigi-bitonti thanks for the PR, can you also add the change on the autopilot-module?
> 
> Hi Daniel, I didn't do it because GCP doc (https://cloud.google.com/kubernetes-engine/docs/how-to/fqdn-network-policies#enable_fqdn_network_policy_in_a_new_cluster) said that "You can't enable this feature when you create a new Autopilot cluster", but only on an existing one. Do you want the feature anyway?

It's a good catch, I didn't see that we this one. So actually, during the preview, we can use it on existing clusters only, that would drive us to this in two steps. I think, but @juliocc keep me honest, I'd go ahead with Standard only and open an issue the integrate this as soon as it reaches the GA stage. @juliocc wdyt?

juliocc:
I'm okay not adding the flag to autopilot as long as we open an issue to keep track of it.

modules/gke-cluster-standard/main.tf
luigi-bitonti: Done :)

juliocc (approved):

PR #1735, opened 2023-10-05
Make deletion protection consistent across all modules

juliocc:
This PR exposes the `deletion_protection` attributed in GKE clusters introduced in version 5 of the provider.

---
**Checklist**


modules/gke-cluster-autopilot/variables.tf
```diff
@@ -34,6 +34,12 @@ variable "backup_configs" {
   nullable = false
 }
 
+variable "deletion_protection" {
+  description = "Whether or not to allow Terraform to destroy the cluster. Unless this field is set to false in Terraform state, a terraform destroy or terraform apply that would delete the cluster will fail."
+  default     = true
```
ludoo, line 39: we should have it as false by default Julio

juliocc: the provider defaults to true

ludoo: well the provider is not always sane :) anyway, we discussed it and I understasnd your point, I would just do the same change on GCS

ludoo (approved):

modules/gke-cluster-autopilot/variables.tf
ludoo: GCS already has it, I stand corrected; can we bubble up this variable in GKE blueprints too? anf FAST

juliocc: As discussed, I'll update `cloudsql-instance` which is behaving in the opposite way and expose it in blueprints to allow easy destruction.

ludoo (approved):

juliocc:
@lcaggio this mostly affects the data blueprints. Can you take a look and make sure I'm not breaking anything?

PR #1746, opened 2023-10-11
Module autopilot bug fixes

luigi-bitonti:
Removed unused variables and forced others to be true because addons _horizontal-pod-autoscaling_ and _http-load-balancing_ are required to be enabled for Autopilot clusters.
---
**Checklist**


modules/gke-cluster-autopilot/variables.tf
```diff
@@ -50,21 +50,14 @@ variable "description" {
 variable "enable_addons" {
   description = "Addons enabled in the cluster (true means enabled)."
   type = object({
-    cloudrun                   = optional(bool, false)
```
ludoo, line 53: Why are you removing more attributes? Please provide a rationale for doing that in the PR description, with links to the doc where it says those are not supported.

juliocc:
@olliefr this is bordering our GKE knowledge. Can you take a look?

modules/gke-cluster-autopilot/variables.tf
luigi-bitonti: image
- Network policy parameter conflicts with Autopilot cluster (https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy#enabling_network_policy_enforcement doc says it's enable by default)

image
- DNS Cache conflicts with Autopilot cluster (https://cloud.google.com/kubernetes-engine/docs/how-to/nodelocal-dns-cache doc says it's enabled by default)

image

- _horizontal-pod-autoscaling_ and _http-load-balancing_ are required to be enabled, so I enforced it in the code, because if users choose _false_ the apply fails.

juliocc: Awesome, can you drop a comment in the code explaining why each of those options is hardcoded for autopilot? What you just said in your comment would be perfect

juliocc (approved):

juliocc (approved):

PR #1754, opened 2023-10-12
Fix typo in GKE nodepool taints

### LINKED ISSUE #1749: gke-nodepool taint loop has faulty declaration
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/gke-nodepool/main.tf#L222

Should be probably `taint.value.effect`

ludoo:
Fixes #1749

### DIFF
--- a/modules/gke-nodepool/main.tf
+++ b/modules/gke-nodepool/main.tf
@@ -47,13 +47,12 @@ locals {
       "https://www.googleapis.com/auth/userinfo.email"
     ]
   )
-  taints_windows = (
-    local.image.is_win
-    ? [{
-      key = "node.kubernetes.io/os", value = "windows", effect = "NO_EXECUTE"
-    }]
-    : []
-  )
+  taints = merge(var.taints, !local.image.is_win ? {} : {
+    "node.kubernetes.io/os" = {
+      value  = "windows"
+      effect = "NO_EXECUTE"
+    }
+  })
 }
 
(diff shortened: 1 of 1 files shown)


juliocc (approved):

modules/gke-nodepool/main.tf
```diff
@@ -219,7 +219,7 @@ resource "google_container_node_pool" "nodepool" {
       content {
```
wiktorn, line 219: `iterator = taint` missing here?

wiktorn (approved):

modules/gke-nodepool/main.tf
juliocc: No, you get the [block name as the default iterator
](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks#:~:text=If%20omitted%2C%20the%20name%20of%20the%20variable%20defaults%20to%20the%20label%20of%20the%20dynamic%20block)

luigi-bitonti: If you don't specify the alias for iterator, it will have the name of the "dynamic" block: https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks 

"The iterator argument (optional) sets the name of a temporary variable that represents the current element of the complex value. If omitted, the name of the variable defaults to the label of the dynamic block."

juliocc (approved):

Issue #1740, opened 2023-10-06
Compute vm Module doesnot support dual stack (ipv4_ipv6) network interface


Avinash-jayakumar:
Compute vm Module doesnot support dual stack (ipv4_ipv6) network interface. This option is available in resource blocks but not in module .

[stack_type](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#stack_type) - (Optional) The stack type for this network interface to identify whether the IPv6 feature is enabled or not. Values are IPV4_IPV6 or IPV4_ONLY. If not specified, IPV4_ONLY will be used.

Avinash-jayakumar:
@ludoo @juliocc can you guys create a new release for this if possible ???

PR #1756, opened 2023-10-12
Exposed stack_type variable in compute_vm module

luigi-bitonti:
---
**Checklist**


ludoo:
Fixes #1740

modules/compute-vm/variables.tf
```diff
@@ -242,6 +242,7 @@ variable "network_interfaces" {
   type = list(object({
     nat        = optional(bool, false)
     network    = string
+    stack_type = optional(string, "IPV4_ONLY")
```
ludoo, line 245: typically we put non-optional attributes first, optional second

ludoo: I would also default the new attribute to `null`

modules/compute-vm/main.tf
```diff
@@ -235,6 +235,7 @@ resource "google_compute_instance" "default" {
     iterator = config
     content {
       network    = config.value.network
+      stack_type = config.value.stack_type
```
ludoo, line 238: network and subnetwork first :)

ludoo: fixed it myself

ludoo (approved):
I fixed variable ordering and aligned the instance template resource to the instance

PR #1757, opened 2023-10-12
Add autoclass to GCS

jeroenmonteban:
Add the option to create a GCS bucket with `autoclass` enabled. [Autoclass](https://cloud.google.com/storage/docs/autoclass) automatically transitions objects in a bucket to appropriate storage class based on each object's access pattern. 

---
**Checklist**


modules/gcs/main.tf
```diff
@@ -32,6 +32,9 @@ resource "google_storage_bucket" "bucket" {
   versioning {
     enabled = var.versioning
   }
+  autoclass {
```
ludoo, line 35: We typically use dynamic blocks even for booleans, so we avoid definining the block when false. It's not hugely important, but a convention we followed so far and it would have avoided having to modify test results. It's fine to leave it like this, we might refactor it later.

ludoo (approved):
Thanks for this, dropped a note but I will resolve it once/if tests pass.

Issue #1764, opened 2023-10-16
net-lb-int - add support for dual stack (IPv4 and IPv6) and for multiple FE IPs (forwarding rules)


LucaPrete:
The current internal network passthrough load balancer doesn't support:

* dual stack: being able to expose both IPv4 and IPv6 IP addresses
* exposing in general multiple frontend IPs (forwarding rules)

PR #1765, opened 2023-10-16
[#1764] net-lb-int: add support for dual stack and multiple forwarding rules

LucaPrete:
The PR adds support for:
* dual stack: being able to configure both IPv4 and IPv6 addresses on the same ILB
* use of multiple forwarding rules (indeed, both IPv4 and IPv6), thus being able to expose multiple frontend IPs

---
**Checklist**

### DIFF
--- a/modules/net-lb-int/README.md
+++ b/modules/net-lb-int/README.md
@@ -119,11 +119,15 @@ module "ilb" {
   project_id    = var.project_id
   region        = "europe-west1"
   name          = "ilb-test"
-  protocol      = "L3_DEFAULT"
   service_label = "ilb-test"
   vpc_config = {
     network    = var.vpc.self_link
     subnetwork = var.subnet.self_link
+  }
+  forwarding_rules_config = {
+    default = {
+      ip_protocol = "L3_DEFAULT"
+    }
   }
   group_configs = {
     my-group = {
@@ -192,7 +196,11 @@ module "ilb" {
     network    = var.vpc.self_link
     subnetwork = var.subnet.self_link
   }
-  ports = [80]
+  forwarding_rules_config = {
+    default = {
+      ports = [80]
+    }
+  }
   backends = [
     for z, mod in module.instance-group : {
       group          = mod.group.self_link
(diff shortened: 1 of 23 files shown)


modules/net-lb-int/outputs.tf
```diff
@@ -29,19 +29,28 @@ output "backend_service_self_link" {
...
-output "forwarding_rule" {
-  description = "Forwarding rule resource."
-  value       = google_compute_forwarding_rule.default
-}
-
-output "forwarding_rule_address" {
+output "forwarding_rule_addresses" {
   description = "Forwarding rule address."
-  value       = google_compute_forwarding_rule.default.ip_address
+  value = [
```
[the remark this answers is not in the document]
ludoo: agree this should be maps, like we do in many other modules for the same usecase

modules/net-lb-int/README.md
```diff
@@ -119,12 +119,16 @@ module "ilb" {
   project_id    = var.project_id
   region        = "europe-west1"
   name          = "ilb-test"
-  protocol      = "L3_DEFAULT"
   service_label = "ilb-test"
   vpc_config = {
     network    = var.vpc.self_link
     subnetwork = var.subnet.self_link
   }
+  forwarding_rules_config = {
```
ludoo, line 127: we should use the default in the variable default in examples, and show a single example with multiple rules defined

modules/net-lb-int/main.tf
```diff
@@ -25,24 +25,26 @@ locals {
...
+resource "google_compute_forwarding_rule" "forwarding_rules" {
+  for_each    = var.forwarding_rules_config
   provider    = google-beta
   project     = var.project_id
   region      = var.region
-  name        = var.name
-  description = var.description
-  ip_address  = var.address
-  ip_protocol = var.protocol
+  name        = "${var.name}-${each.key}"
```
ludoo, line 33: I would do

```hcl
name = (
  each.key == "" ? var.name : "${var.name}-${each.key}"
)
```

The common use case that probably covers 99.9% of usage (as we never had support for multiple forwarding rules till now) is to use a single rule, and in that case the name should match names used for all other resources like backend services, health check, etc.

ludoo (changes requested):
very nice find Luca, dropped a couple minor comments

modules/net-lb-int/README.md
ludoo: in this example, like 127-131 would be removed (we can default the protocol in the variable type too)

modules/net-lb-int/outputs.tf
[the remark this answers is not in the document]
LucaPrete: Yes, let's do this! I'll make it happen for the other fwr outputs as well

modules/net-lb-int/README.md
LucaPrete: we do set the default for the variable to "TCP". This is a specific example where we show how to set `L3_DEFAULT`. That's why I specified the protocol.
I'll add one more example with multiple protocols.

LucaPrete: sorry I closed this by mistake... let me know what you think ;)

LucaPrete:
Thanks folks! I'm not super convinced of leaving the key of the terraform state set to "" if we don't specify forwarding rules. We still have the choice to set it somehow in that case to "default". At the same time it might be confusing for users, as it wouldn't correspond to the default dictionary key. Let me know what you think. Thank you

wiktorn (approved):

modules/net-lb-int/outputs.tf
```diff
@@ -29,19 +29,28 @@ output "backend_service_self_link" {
   value       = google_compute_region_backend_service.default.self_link
 }
 
-output "forwarding_rule" {
-  description = "Forwarding rule resource."
-  value       = google_compute_forwarding_rule.default
+output "forwarding_rule_addresses" {
+  description = "Forwarding rule address."
+  value = {
+    for fwr_name, fwr in google_compute_forwarding_rule.forwarding_rules
```
ludoo, line 35: can we use `for k, v in` here and below?

ludoo (approved):

PR #1771, opened 2023-10-17
Fix resource manager tag bindings in compute-vm module

### LINKED ISSUE #1763: Compute VM, tag_bindings don't work
The implementation of setting secure tags on a VM instance in module compute_vm in tags.tf is not right.

There two ways:
A: Use of "google_tags_location_tag_binding" instead of "google_tags_tag_binding". Requires also a project_number variable.
B: Delete tags.tf and add this to main.tf 
params {
    resource_manager_tags = var.tag_bindings
  }

I am not sure what of this two is the right way.

ludoo:
Fixes #1763.

Context of the change is in the linked issue.


modules/compute-vm/README.md
```diff
@@ -677,6 +678,32 @@ module "instance" {
...
 ```
+
+### Resource Manager Tags
+
+Resource manager tags (or "secure tags") bindings are supported with the following limitations:
+
+- a single `tag_bindings` variable is used for both the instance and the boot disk
+- tag bindings are not created for attached disks
+- tag bindings will not be created for the boot disk if the `use_independent_disk` flag is true
+- tag bindings are ignore for instance templates
```
wiktorn, line 689: "are ignore" -> "are ignored"

wiktorn (approved):
Looks good!

modules/compute-vm/README.md
ludoo: 🤦  thanks for catching this :)

PR #1774, opened 2023-10-18
Added ProtectedApplication feature to GKE Backup

luigi-bitonti:
With this PR it was introduced the possibility to create fine-grained GKE backup.

---
**Checklist**


modules/gke-cluster-standard/main.tf
```diff
@@ -419,13 +419,24 @@ resource "google_gke_backup_backup_plan" "backup_plan" {
       }
     }
 
-    all_namespaces = lookup(each.value, "namespaces", null) != null ? null : true
+    all_namespaces = each.value.all_namespaces ? true : null
```
luigi-bitonti, line 422: This because onyl one of _all_namespaces_, _selected_namespaces_ and _selected_applications_ can be defined.

modules/gke-cluster-standard/variables.tf
```diff
@@ -19,13 +19,18 @@ variable "backup_configs" {
...
+        name      = string
+      })))
       encryption_key                    = optional(string)
       include_secrets                   = optional(bool, true)
       include_volume_data               = optional(bool, true)
       namespaces                        = optional(list(string))
       region                            = string
       schedule                          = string
-      retention_policy_days             = optional(string)
+      retention_policy_days             = optional(string, 1)
```
luigi-bitonti, line 33: This because with _schedule_ defined you must also set Backup Retain Days.

tests/modules/gke_cluster_standard/examples/backup.yaml
```diff
@@ -19,15 +19,18 @@ values:
 
   module.cluster-1.google_gke_backup_backup_plan.backup_plan["backup-1"]:
     backup_config:
-    - all_namespaces: true
+    - all_namespaces: null
       encryption_key: []
       include_secrets: true
       include_volume_data: true
-      selected_applications: []
+      selected_applications:
```
luigi-bitonti, line 26: This to test the new feature

juliocc (changes requested):

modules/gke-cluster-standard/variables.tf
luigi-bitonti: So can we make _retention_policy_days_  not optional?

juliocc: I don't think so. It looks like schedule is actually optional:

image

ludoo: while you're at it, can you pute required attributes before optional ones?

luigi-bitonti: OK @juliocc , so let's make _schedule_ optional too?

juliocc: sgtm!

modules/gke-cluster-standard/variables.tf
```diff
@@ -19,13 +19,17 @@ variable "backup_configs" {
   type = object({
     enable_backup_agent = optional(bool, false)
     backup_plans = optional(map(object({
+      applications = optional(list(object({
+        namespace = string
+        name      = string
+      })))
```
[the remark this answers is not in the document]
luigi-bitonti: Done!

modules/gke-cluster-standard/main.tf
```diff
@@ -419,13 +429,27 @@ resource "google_gke_backup_backup_plan" "backup_plan" {
...
+      content {
+        dynamic "namespaced_names" {
+          for_each = local.bkp_namespaces_object_map[each.key]
+          iterator = ns_obj_couple
+          content {
+            namespace = ns_obj_couple.value["namespace"]
+            name      = ns_obj_couple.value["object_name"]
+          }
+        }
+      }
```
juliocc, line 450: you're making this more complicated than necessary with the locals. The following should be enough

```hcl
        dynamic "namespaced_names" {
          for_each = flatten([for k, vs in each.value.applications : [
            for v in vs : { namespace = k, name = v }
          ]])
          content {
            namespace = namespaced_names.value.namespace
            name      = namespaced_names.value.name
          }
        }
```

luigi-bitonti: You're right, just pushed your suggest. I thought we need to avoid too many logic in resource block. Thank you (:

juliocc: Usually, yes. But this one was simple enough

juliocc (approved):

PR #1789, opened 2023-10-22
Add 30s sleep to make sure that services are ready for next steps

wiktorn:
Workaround for:
```
If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
```
---
**Checklist**

### DIFF
--- a/tests/examples_e2e/setup_module/main.tf
+++ b/tests/examples_e2e/setup_module/main.tf
@@ -48,19 +48,24 @@ resource "google_project_service" "project_service" {
   disable_dependent_services = true
 }
 
+resource "time_sleep" "wait_30_seconds" {
+  depends_on      = [google_project_service.project_service]
+  create_duration = "30s"
+}
+
 resource "google_storage_bucket" "bucket" {
   location      = var.region
   name          = "${local.prefix}-bucket"
   project       = google_project.project.project_id
   force_destroy = true
-  depends_on    = [google_project_service.project_service]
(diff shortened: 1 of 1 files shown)


juliocc (commented):
You don't need a timeout for this.

Are you seeing any failures?

wiktorn:
I do not see failures when running locally, but it happens when run by scheduled workflows [here](https://github.com/fabric-team/fast-e2e/actions/runs/6602351789) and [here](https://github.com/fabric-team/fast-e2e/actions/runs/6602351789/job/17934304194).

The other approach that comes to my mind, is to split the list of the services into 2-3 buckets, chain them, that second needs to be done after first, and third after second. And divide services into buckets:
* bucket 1 - services used in setup project (like compute)
* bucket 2 - few services to activate, instead of timeout
* bucket 3 - all other services required by tests

And make a dependency of resources in `setup_project` on bucket_2 of services. And the number / time to activate those services will ensure, that activation of bucket 1 services propagated too all required places.

It is surprising that it doesn't happen when workflow is run by hand or scheduled run is rerun after failure.

juliocc:
TBH i hate it. It's a code smell. 

This kind of thing usually works with modules. What's different here?

wiktorn:
Lucky enough, today's workflow did not fail. Let's give it a few more tries and see how often this occurs.

wiktorn:
Looks like few recent runs did not fail. Closing this.

PR #1806, opened 2023-10-24
Updating network dashboard: fixing Cloud SQL problem, fixing 1 metric…

aurelienlegrand:
Changes:
- fixing Cloud SQL IP address issue
- fixing 1 metric issue in the dashboard
- pausing for monitoring quota issues after 170 monitoring API queries
- Fixing behavior as described in the README: if monitored folders and projects are empty, every project under the discovery root node will be monitored.


blueprints/cloud-operations/network-dashboard/deploy-cloud-function/main.tf
```diff
@@ -47,7 +47,7 @@ module "pubsub" {
   project_id    = module.project.project_id
   name          = var.name
   regions       = [var.region]
-  subscriptions = { "${var.name}-default" = null }
+  subscriptions = {}
 }
```
ludoo, line 51: why this change?

blueprints/cloud-operations/network-dashboard/src/plugins/monitoring.py
```diff
@@ -103,4 +104,10 @@ def timeseries_requests(project_id, root, timeseries, descriptors):
     tot_num = sum(len(b) for b in ts_buckets)
     LOGGER.info(f'sending {req_num} remaining: {tot_num}')
     yield HTTPRequest(url, HEADERS, json.dumps(data))
+    api_calls += 1
+    # Default quota is 180 request per minute per user
+    if (api_calls > 170):
```
ludoo, line 109: you're not really checing that you have 170 calls in a minute, you're checking you made 170 in an arbitrary amount of time :)

blueprints/cloud-operations/network-dashboard/deploy-cloud-function/main.tf
juliocc: I think the subscription was not being used, and the `=null` part doesn't work since I updated the module

aurelienlegrand: What Julio said, I think the new version of the module works better with {} (uses the default subscription)

blueprints/cloud-operations/network-dashboard/src/plugins/monitoring.py
aurelienlegrand: I think this is a good enough approximation as API calls usually are pretty quick (multiple per second), I don't think it makes sense to have a more complex system?

blueprints/cloud-operations/network-dashboard/deploy-cloud-function/main.tf
ludoo: Aha Fabric is becoming too large to keep in a single head. Thanks for the clarification. 😀

blueprints/cloud-operations/network-dashboard/src/plugins/monitoring.py
ludoo: I think we either do it properly, or just catch the quota error and sleep then without counting.

ludoo: Something like this:

```python
t = time.time()
num_calls = 0
while ts_buckets:
  # loop code here
  num_calls += 1
  if num_calls == 170:
    td = time.time() - t
    if td < 60:
      time.sleep(60 - td)
    num_calls, td = 0, time.time()
```

ludoo: doing it with a queue of timestamps that you push/pop would be more precise, as you would continuously check the current interval, but this is still a decent approximation maybe

juliocc: What about mounting [this adapter](https://github.com/JWCook/requests-ratelimiter#adapter) to the [HTTP session object](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/a3290f2204342647f99d6d443d93fdfc438c2ce2/blueprints/cloud-operations/network-dashboard/src/main.py#L32C8-L32C25)?

ludoo: I would not bring in extra libraries, which then need to be kept up to date, understood, etc.

juliocc: I'm lazy and try not to reinvent the wheel ;)

ludoo: https://wiki.c2.com/?LazinessImpatienceHubris

juliocc: Just a suggestion, it's not my code so you're free to do what you do :)

ludoo: Well, I also think it's a general approach to how we do things :) I try to limit library use as much as possible to std lib, I even avoid the SDK so as to limit complexity and cognitive surface as much as possible.

aurelienlegrand: I think the time interval solution proposed by Ludo is good enough, I will update the PR :)

aurelienlegrand: After testing this optimization, we sleep for about 30s instead of 60s so we saved some execution time for large inventories!

ludoo: thanks for confirming, that was the plan :)

ludoo (approved):

PR #1810, opened 2023-10-24
FAST: Add access transparency logs to the default sinks

sruffilli:
Adds access transparency logs to the default sinks
---
**Checklist**


ludoo (commented):
WDYT about bundling them up with the other audit logs, since it will be easy to filter from the log viewer? Pros of this they are easy to turn off, but then they are small so it's not like it would save anything vs leaving them on... I personally prefer less clutter in variables file, YMMV ofc

sruffilli:
> WDYT about bundling them up with the other audit logs
Makes sense indeed

ludoo (approved):

Issue #1794, opened 2023-10-23
Factories fail when empty YaML files are passed


juliocc:
Factories fail when empty YaML files are passed. 
Some factories mitigate the problem by using `try(yamldecode())`, which however masks potential parse errors.

_original message below:_

getting the following error when applying 02-networking stage with empty YAML files for firewall rules
example files:
2-networking-b-vpn/data/hierarchical-ingress-rules.yaml
2-networking-b-vpn/data/firewall-rules/dev/rules.yaml

cc @aaarel

sruffilli:
WIP tests log:

Works if empty
`2-networking-b-vpn/data/hierarchical-ingress-rules.yaml` 

Fails if empty
`2-networking-b-vpn/data/firewall-rules/dev/rules.yaml`

ludoo:
Yep, problem is `yamldecode` not liking empty/comment only YAML files. We had `try` around the `yamldecode` in some places and not others, then Terraform was supposedly fixed and we removed it, and then it was added back in some places.

Discussed it the other day with Julio, and as long as we're consistent and users know what to expect, either solution works each with its own drawbacks

- `try` makes comment-only files work but hides errors
- no `try` exposes errors but then it should be a surprise that a comment-only file breaks

sruffilli:
Came here with the same comments. 

I'd be tempted to remove `try` everywhere, at which point users have two options:

- Create a valid YaML empty map `{}` (which can be a comment documenting the behavior at the top of each sample file)
- Remove the factory configuration altogether.

_Explicit is better than implicit_

Wdyt?

ludoo:
Agree, I would also drop a comment in our example files, stating that the file should either be valid or be removed altogether. So users know what to expect.

juliocc:
After a couple of tests, a empty file can be "fixed" with `---`
```
> yamldecode("")
╷
│ Error: Error in function call
│
│   on <console-input> line 1:
│   (source code not available)
│
│ Call to function "yamldecode" failed: on line 1, column 1: missing start of document.
╵


> yamldecode("#")
╷
│ Error: Error in function call
│
│   on <console-input> line 1:
│   (source code not available)
│
│ Call to function "yamldecode" failed: on line 1, column 1: missing start of document.
╵


> yamldecode("---")
null
> yamldecode("#\n---")
null
```

ludoo:
That's genius. So we just append `\n---` to any file before passing to yamldecode?

[...]

PR #1812, opened 2023-10-25
Stop wrapping yamldecode with try()

sruffilli:
Per #1794, updating all factories to avoid wrapping `yamldecode` in `try`, to have terraform explicitly fail in case of a malformed yaml file.

Since empty yaml files result in `yamldecode` to fail, this PR also updates all yaml files used in factories, adding the comment below to advice the user on how to create an empty document.

```yaml
---
# Terraform will be unable to decode this file if it does not contain valid YAML
# You can retain `---` (start of the document) to indicate an empty document.
```

**Checklist**


juliocc (approved):

modules/dns-response-policy/main.tf
```diff
@@ -15,7 +15,9 @@
  */
 
 locals {
-  _factory_rules = try(yamldecode(file(var.rules_file)), {})
+  _factory_data  = var.rules_file != null ? file(var.rules_file) : "{}"
```
sruffilli, line 18: @ludoo @juliocc PTAL.

sruffilli: Background: the ternary operator complained for inconsistent type only for some specific yaml files (I have isolated a test case if you're curious). 
To avoid piling up workarounds I'm leveraging a temporary string variable where I store the content of the file (or the representation of an empty yml), and yamldecode() that string after.

wiktorn: if this works, then probably `yamldecode(var.rules_file != null ? ... : ...)` should work too? (and I guess, should be default approach where ternary operator is accompanied by `yamldecode`)

ludoo: been through the same hell, would be very curious to see the test case together

PR #1813, opened 2023-10-26
empty gpu sharing config fix

ewojtach:
Hot fix for empty GPU sharing config 
---
**Checklist**


modules/gke-nodepool/main.tf
```diff
@@ -168,7 +168,8 @@ resource "google_container_node_pool" "nodepool" {
         gpu_partition_size = var.node_config.guest_accelerator.gpu_driver == null ? null : var.node_config.guest_accelerator.gpu_driver.partition_size
 
         dynamic "gpu_sharing_config" {
-          for_each = var.node_config.guest_accelerator.gpu_driver != null ? [""] : []
+          #for_each = var.node_config.guest_accelerator.gpu_driver != null ? [""] : []
+          for_each = var.node_config.guest_accelerator.gpu_driver == null ? [] : (var.node_config.guest_accelerator.gpu_driver.max_shared_clients_per_gpu != null ? [""] : [])
```
wiktorn, line 172: This gets pretty long, you can make it shorter for example like that:
```hcl
          for_each = lookup(
            lookup(var.node_config.guest_accelerator, "gpu_driver", {}),
            "max_shared_clients_per_gpu"
          ) != null ? [""] : []
```

The other thing is, given the code here - maybe `max_shared_clients_per_gpu` should not be optional, so once `gpu_driver` is non-null, `max_shared_clients_per_gpu` will be always set, so the additional conditions are not needed?

wiktorn (approved):

PR #1815, opened 2023-10-26
Fix logic for default source range in firewall ingress rules

ludoo:
This fixes i bug I observed today in the real world, where an unwanted source range set to 0/0 is added when

- `sources` is set to tags or service accounts
- no `source_ranges` are specified

The fix has been tested live with a customer and we have confirmed it removes the unwanted source range, with the resulting rule being identical to one defined int he console with the same input parameters.


wiktorn (approved):
Scary bug!

ludoo:
> Scary bug!

ahah yes :) thanks for the reviews!

PR #1816, opened 2023-10-26
import job support for kms module

cmalpe:
Added support import job support for the "kms" module using the official terraform resource "google_kms_key_ring_import_job". Link: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/kms_key_ring_import_job. 

I have successfully tested the "kms" module's new functionality by creating the Cloud KMS Import job resource in my test GCP environment. I was not sure if I needed to write any new python tests since this change was minor and only limited to the "kms" module.

### DIFF
--- a/modules/kms/main.tf
+++ b/modules/kms/main.tf
@@ -55,6 +55,7 @@ resource "google_kms_crypto_key" "default" {
 }
 
 resource "google_kms_key_ring_import_job" "default" {
+  count            = var.import_job_create ? 1 : 0
   key_ring         = local.keyring.id
   import_job_id    = var.import_job.id
   import_method    = var.import_job.import_method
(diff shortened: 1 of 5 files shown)


modules/kms/main.tf
```diff
@@ -53,3 +53,11 @@ resource "google_kms_crypto_key" "default" {
     }
   }
 }
+
+resource "google_kms_key_ring_import_job" "default" {
+  count            = var.import_job_create ? 1 : 0
```
juliocc, line 58: Looks like you can get away with a single variable. The condition here can be `var.import_job != null ? 1 : 0`

juliocc (changes requested):

modules/kms/main.tf
cmalpe: Hi @juliocc I don't know why I went with an additional variable :/ I will add this change.

cmalpe:
closing this PR and raising a new clean one

PR #1768, opened 2023-10-17
Lcaggio/df cx

lcaggio:
Create a [Dialgoflow CX agent](https://cloud.google.com/dialogflow/cx/docs/concept/agent) with private connectivity to a Cloud Run instance acting as webhook.

---

I applicable, I acknowledge that I have:
- [X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X] Ran `terraform fmt` on all modified files
- [X] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [X] Made sure all relevant tests pass

PR #1819, opened 2023-10-27
Serverless networking program

juliodiez:
Blueprint to deploy two microservices running in Cloud Run and communicating with each other.

The content of this blueprint corresponds to the chapter 'Microservices architectures - Developing Microservices applications' of the Serverless Networking Guide (https://services.google.com/fh/files/misc/serverless_networking_guide.pdf).

---
**Checklist**
- [ x] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [ x] Ran `terraform fmt` on all modified files
- [ x] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [ x] Made sure all relevant tests pass


blueprints/serverless/cloud-run-microservices/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+  })
+  default = null
+}
+
+variable "prj_main_id" {
+  description = "Main Project ID."
+  type        = string
+}
+
+variable "prj_svc1_create" {
```
wiktorn, line 57: can we call this (and related stuff) `service_project` instead of `prj_svc1`?

blueprints/serverless/cloud-run-microservices/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+    main = {
+      subnet_main       = "10.0.1.0/24"
+      subnet_proxy      = "10.10.0.0/24"
+      subnet_vpc_access = "10.10.10.0/28"
+      subnet_vpc_direct = "10.8.0.0/26"
+      psc_addr          = "10.0.0.100"
+    }
+  }
+}
+
```
wiktorn, line 42: Can we also add `prefix` variable and use it when creating projects?

blueprints/serverless/cloud-run-microservices/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+variable "prj_main_create" {
+  description = "Parameters for the creation of the main project."
+  type = object({
+    billing_account_id = string
+    parent             = string
+  })
+  default = null
+}
+
+variable "prj_main_id" {
```
wiktorn, line 52: what do you think about using such structure instead:
```hcl

variable "main_project" {
  type = object({
    project_id = string
    project_create = optional(object({
      billing_account_id = string
      parent             = string
    }, null))
  })
}
```

blueprints/serverless/cloud-run-microservices/cloudrun.tf
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# tfdoc:file:description Cloud Run services.
+
+# Cloud Run service A
+resource "google_cloud_run_v2_service" "svc_a" {
```
wiktorn, line 20: What is the rationale here to use raw resources instead of `modules/cloud-run`?

blueprints/serverless/cloud-run-microservices/cloudrun.tf
```diff
@@ -0,0 +1,94 @@
...
+
+resource "google_cloud_run_v2_service_iam_policy" "svc_a_policy" {
+  project     = module.project_main.project_id
+  location    = var.region
+  name        = google_cloud_run_v2_service.svc_a.name
+  policy_data = data.google_iam_policy.noauth.policy_data
+}
+
+# Cloud Run service B
+module "cloud_run_svc_b" {
```
wiktorn, line 65: use dashes in module names instead of underscore

blueprints/serverless/cloud-run-microservices/alb.tf
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# tfdoc:file:description Internal Application Load Balancer resource.
+
+# Internal Application Load Balancer in main (host) project
+module "int-alb" {
+  source     = "../../../modules/net-lb-app-int"
+  count      = var.prj_svc1_id != null ? 1 : 0
```
wiktorn, line 22: I know this follows what is described in `README.md` but results in a bit of magic behavior

wiktorn (changes requested):
I think a lot of comments in #1768 also apply here, especially incorporating external code + docker image build.

blueprints/serverless/cloud-run-microservices/variables.tf
juliodiez: Done.

blueprints/serverless/cloud-run-microservices/variables.tf
juliodiez: Done.

blueprints/serverless/cloud-run-microservices/variables.tf
juliodiez: I like it, though the _legacy_ way is all around. I used a slightly different one, it's simpler and shorter to use, PTAL.

blueprints/serverless/cloud-run-microservices/cloudrun.tf
juliodiez: Beta features (Direct VPC Egress), the cloud run module uses resource "google_cloud_run_service" that doesn't support it.

blueprints/serverless/cloud-run-microservices/cloudrun.tf
juliodiez: Done.

blueprints/serverless/cloud-run-microservices/alb.tf
juliodiez: I'm following this pattern in these cloud run blueprints because it allows to create several related use cases inside a blueprint simply (it's not so _magic_), reusing 80-90% of code instead of creating multiple very similar blueprints.

blueprints/serverless/cloud-run-microservices/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+  default = {
+    subnet_main       = "10.0.1.0/24"
+    subnet_proxy      = "10.10.0.0/24"
+    subnet_vpc_access = "10.10.10.0/28"
+    subnet_vpc_direct = "10.8.0.0/26"
+    psc_addr          = "10.0.0.100"
+  }
+}
+
+variable "main_project" {
```
ludoo, line 35: can we reafctor this so we have a single top-level `project_configs` variable with optional members for `main` and `service`?

blueprints/serverless/cloud-run-microservices/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+  description = "Service project."
+  type = object({
+    billing_account_id = optional(string)
+    parent             = optional(string)
+    project_id         = optional(string)
+  })
+  default = {}
+}
+
+variable "svc_a_image" {
```
ludoo, line 69: can we merge these two variables into a single `image_configs` variable?

blueprints/serverless/cloud-run-microservices/main.tf
```diff
@@ -0,0 +1,63 @@
...
+
+# tfdoc:file:description Project resources.
+
+locals {
+  cloud_run_domain = "run.app."
+  svc_a_name       = "svc-a"
+  svc_b_name       = "svc-b"
+}
+
+# Main (or host) project
```
ludoo, line 25: comment is unnecessary

blueprints/serverless/cloud-run-microservices/cloudrun.tf
ludoo: Agree with Wiktor, can we add support for this at the module level?

blueprints/serverless/cloud-run-microservices/cloudrun.tf
```diff
@@ -0,0 +1,91 @@
...
+      content { # Use Direct VPC Egress
+        network_interfaces {
+          subnetwork = module.vpc-main.subnets["${var.region}/subnet-vpc-direct"].name
+        }
+      }
+    }
+  }
+}
+
+data "google_iam_policy" "noauth" {
```
ludoo, line 47: why do you need a data source? this will make this blueprint untestable

blueprints/serverless/cloud-run-microservices/alb.tf
wiktorn: Totally agree, that there is no need for separate blueprints to cover very similar cases. What I don't like about it, that based on whether service project id is provided or not, so many things do change. But I guess there is no other simple way, and I consider simplicity to be a top-priority here.

blueprints/serverless/cloud-run-microservices/outputs.tf
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+output "custom_domain" {
+  description = "Custom domain for the Application Load Balancer."
+  value       = var.service_project.project_id != null ? "http://${var.custom_domain}" : "none"
+}
+
+output "default_URL_svc_a" {
```
wiktorn, line 22: We can also group output variables `default_URL_svc_a` and `default_URL_svc_b` under one map

blueprints/serverless/cloud-run-microservices/variables.tf
juliodiez: Done

blueprints/serverless/cloud-run-microservices/variables.tf
juliodiez: Done

blueprints/serverless/cloud-run-microservices/cloudrun.tf
juliodiez: I was using IAM policy (and following the provider example), changed to use IAM binding.

blueprints/serverless/cloud-run-microservices/outputs.tf
juliodiez: Done

blueprints/serverless/cloud-run-microservices/cloudrun.tf
wiktorn: We had the discussion some time ago about this with @ludoo and the conclusion was, that as the resources are the same from GCP perspective, we don't need a separate module (that is, two cloud-run modules in Fabric), just one that uses new API.

So I think that time has come to do that.

I'm just not sure whether to keep the same name, especially if we manage to keep the same interface. Because then - it will just work, but if you happen to have instances in the state, it will be a mess. Usually backward incompatible changes in Fabric announce themselves not only in `CHANGELOG.md` but also as errors before apply.

ludoo: Good point Wiktor, what if we created a new module, and then deprecated the old one in the next release?

ludoo: Not what we discussed, but maybe it's time we became a little bit more conservative at times...

juliodiez: This is what I was afraid of, releasing and deprecating may be not that smooth... So I wanted to move forward with this blueprint using the resource to not be blocked. Updating it once the new module is available would be straightforward.
It's your call, folks.

blueprints/serverless/cloud-run-microservices/README.md
```diff
@@ -0,0 +1,229 @@
...
+}
+prefix = "[prefix]"
+project_configs = {
+  main = {
+    project_id = "[project_id]"
+  }
+}
+```
+
+Note that final project ids will be of the form "[prefix]-[main-project-id]".
```
wiktorn, line 107:
```suggestion
Note that final project ids will be of the form "[prefix]-[project-id]".
```

wiktorn (approved):
Looks good, more comments to README.md than code itself :-)

blueprints/serverless/cloud-run-corporate/README.md
```diff
@@ -4,20 +4,21 @@
...
 
 ## Prerequisites
 
-Depending on the use case, you will need one or more projects with [billing enabled](https://cloud.google.com/billing/docs/how-to/modify-project) and a user with the “Project owner” [IAM](https://cloud.google.com/iam) role on those projects. You can use existing projects or let the blueprint creates them for you but in that case you will need to add extra information for each project. E.g.:
+Depending on the use case, you will need one or more projects with [billing enabled](https://cloud.google.com/billing/docs/how-to/modify-project) and a user with the “Project owner” [IAM](https://cloud.google.com/iam) role on those projects. You can use existing projects or let the blueprint create them for you but in that case you will need to add extra information for each project. E.g.:
```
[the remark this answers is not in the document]
juliodiez: I think this is legacy text taken from some older blueprint. It is true, even Editor is still too much. There would be multiple roles needed depending on the use case, resources in use, and if projects exist or not. But I don't think we need to give so much guidance on the IAM topic to deploy a blueprint, the idea of a blueprint is to play with it, not production-ready code. I prefer to remove it.

blueprints/serverless/cloud-run-corporate/README.md
```diff
@@ -27,7 +28,7 @@ prj_main_create = {
 }
 ```
 
-Below it is explained how to set this information.
+How to set this information is explained below.
```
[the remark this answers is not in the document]
juliodiez: Not exactly, it refers to the general process, not just projects. Slightly changed.

blueprints/serverless/cloud-run-microservices/README.md
```diff
@@ -0,0 +1,229 @@
...
+terraform apply
+```
+
+It will take a few minutes. When complete, you should see an output stating the command completed successfully, a list of the created resources, and some output variables with information to access your services.
+
+__Congratulations!__ You have successfully deployed the use case you chose based on the variables configuration.
+
+### Use case 1: Service to service communication in the same project
+
+This use case deploys two Cloud Run services in the same project. Service B is protected as an _internal only_ service and communication between Cloud Run services, even in the same project, is not considered internal by default. To communicate them, one option is to use PSC/PGA (the second option is shown in the following use case). We will use a PSC endpoint and a Serverless VPC connector to reach Service B. A DNS record for the PSC endpoint is created.
```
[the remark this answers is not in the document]
juliodiez: No. PGA/PSC is mostly the same method, so it is the first option. The second option is using an ILB, that's shown in use case 2. Text slightly changed.

blueprints/serverless/cloud-run-microservices/README.md
```diff
@@ -0,0 +1,229 @@
...
+}
+```
+
+Note that final project ids will be of the form "[prefix]-[main-project-id]".
+
+The service B default URL is created and shown as a terraform output variable. It will be similar to the one shown in the picture above. Get into service A and try to reach service B URL as shown below:
+
+<p align="center"> <img src="images/use-case-1-test.png" width="600"> </p>
+
+You can see service A is resolving service B to an internal IP, 10.0.0.100, the PSC endpoint. Public access is restricted, if you try to e.g. `curl` from your laptop you will get an error.
```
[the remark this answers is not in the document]
juliodiez: Well, the PSC address is really defined in variables.tf, I could add a reference to that.

blueprints/serverless/cloud-run-microservices/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+This use case deploys two Cloud Run services in the same project. Service B is protected as an _internal only_ service and communication between Cloud Run services, even in the same project, is not considered internal by default. To communicate them, one option is to use PSC/PGA (the second option is shown in the following use case). We will use a PSC endpoint and a Serverless VPC connector to reach Service B. A DNS record for the PSC endpoint is created.
+
+<p align="center"> <img src="images/use-case-1.png" width="600"> </p>
+
+Service A uses an application with a GUI to test connectivity. You can find its source code in:
+
+https://github.com/willypalacin/vpc-network-tester/tree/main
+
+Build an image and push it to Artifact Registry, setting the corresponding Terraform variable to its URL. Add the main project ID in `terraform.tfvars`. E.g.:
```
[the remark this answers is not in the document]
juliodiez: I think it's better to redirect users to quick guides. Added link to official one.

PR #1821, opened 2023-10-27
[net-address] enable ipv6

LucaPrete:
Allows users to reserve IPv6 addresses.

FYI I submitted the initial PR without modifying the BP tests (yet), so we can first make sure that the modification to global addresses is done the way we want. Please, let me know if looks of to you.

---
**Checklist**

### DIFF
--- a/modules/net-address/variables.tf
+++ b/modules/net-address/variables.tf
@@ -17,31 +17,39 @@
 variable "external_addresses" {
   description = "Map of external addresses, keyed by name."
   type = map(object({
-    region      = string
-    description = optional(string, "Terraform managed.")
-    labels      = optional(map(string), {})
-    name        = optional(string)
+    region             = string
+    description        = optional(string, "Terraform managed.")
+    ip_version         = optional(string, "IPV4")
+    ipv6_endpoint_type = optional(string)
+    labels             = optional(map(string), {})
+    name               = optional(string)
   }))
   default = {}
 }
 
 variable "global_addresses" {
   description = "List of global addresses to create."
-  type        = list(string)
-  default     = []
+  type = map(object({
+    description = optional(string, "Terraform managed.")
+    ip_version  = optional(string, "IPV4")
+    name        = optional(string)
+  }))
+  default = {}
 }
 
 variable "internal_addresses" {
   description = "Map of internal addresses to create, keyed by name."
   type = map(object({
-    region      = string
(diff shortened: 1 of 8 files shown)


modules/net-address/variables.tf
```diff
@@ -17,31 +17,39 @@
 variable "external_addresses" {
   description = "Map of external addresses, keyed by name."
   type = map(object({
-    region      = string
-    description = optional(string, "Terraform managed.")
-    labels      = optional(map(string), {})
-    name        = optional(string)
+    region             = string
+    description        = optional(string, "Terraform managed.")
+    ip_version         = optional(string, "IPV4")
```
ludoo, line 22: luca, would it make sense to have, instead of `ip_version` and `ipv6_endpoint_type`, something like

```hcl
 description        = optional(string, "Terraform managed.")
 ipv6 = optional(object({
  endpoint_type = optional(string)
}))
```

Then having `each.value.ipv6 == null` means it's IPv4.

LucaPrete: What if it becomes ipv6_endpoint_type only (or something similar...)?

LucaPrete: Something to consider is also that you don't specify the type for global addresses. Keeping the two params would let us use the same arguments to activate ipv6 for all ip addresses types

ludoo: > What if it becomes ipv6_endpoint_type only (or something similar...)?

It would be semantically confusing IMHO, what you want is a way to set IPv6, and the endpoint is optional.

> Something to consider is also that you don't specify the type for global addresses. Keeping the two params would let us use the same argument to activate ipv6 for all ip addresses types

You can as well use an empty object for the other types...

One more comment is that since the endpoint type has only two values, it could very well be a bool: `ipv6.netlb_endpoint = true`.

LucaPrete: Agree on the difference between what's mandatory and what's optional.
I would prefer to use the strings instead of the bool if you don't mind for the type.

Would that be ok?

ludoo: Sure, but we need to add a validation rule then.

LucaPrete: Yep. Makes sense! Thanks Ludo.

ludoo (approved):

PR #1822, opened 2023-10-28
Billing budget factory

ludoo:
This implements support for a billing budgets factory in the billing account module. A subsequent PR will refactor factories across modules

- align all factory variables to the same format used here
- move the project factory to the project module (for projects)
- move the service account factory to the service account module and add support for the full interface
- update the FAST project factory to leverage project, service account, and billing budget factories
- remove the project factory from blueprints and replace the blueprint project factory subfolders with a README file listing all modules that support a factory


modules/billing-account/README.md
```diff
@@ -212,21 +213,63 @@ module "billing-account" {
...
+  budget_notification_channels = {
+    billing-default = {
+      project_id = "tf-playground-simple"
+      type       = "email"
+      labels = {
+        email_address = "gcp-billing-admins@example.com"
+      }
+    }
+  }
+}
```
wiktorn, line 233: missing:
```
factory_configs = {
  budgets_folder = "data/billing-budgets"
}
```
?

ludoo: there's a default in the variable...

ludoo: meaning, the module tries the folder regardless of configuration; a filesystem lookup is a trivial penalty compared to all the rest that happens in a terraform plan/apply, IMHO :)

ludoo: image

ludoo: the idea is that all factory-supporting modules will share the same variable, where `data_path` can be set to the same folder, and then one additional attribute points to the location of the factory files inside it

wiktorn: :facepalm: I guess this might be worth pointing out somewhere in `README.md` that by default, the module will look for factory files in specific paths, or make `data_path` null by default and require setting it to enable this functionality.

ludoo: let's point out the default in the README, the split variable might also be confusing, but using a single variable would then make the type string (the folder path), and some factories might have more than one source of data / support more than one resource type

I might be falling in the premature optimization tunnel though :)

modules/billing-account/factory.tf
```diff
@@ -0,0 +1,103 @@
...
+        credit_types_treatment = (
+          try(v.filter.credit_types_treatment, null) == null
+          ? null
+          : merge(
+            { exclude_all = null, include_specified = null },
+            v.filter.credit_types_treatment
+          )
+        )
+        label = (
+          try(v.filter.label, null) == null ? null : v.filter.label
```
wiktorn, line 48: can't we have:
```hcl
try(v.filter.label, null)
```
as below?

modules/billing-account/README.md
wiktorn: It kind of feels like a bit like magic and explaining to users:
> don't create `data/` folder in your root as Fabric will peek there and pull stuff from there

Though I agree, that probability that someone hits the same naming pattern and DO NOT want to use this feature is like <1%.

ludoo: let me revert to a simpler thing, we will have time to make it more complex :)

modules/billing-account/factory.tf
ludoo: 🤦

ludoo: done

modules/billing-account/README.md
```diff
@@ -212,21 +213,63 @@ module "billing-account" {
 # tftest modules=1 resources=2 inventory=budget-monitoring-channel.yaml
 ```
 
+#### Budget factory
+
+This module also exposes a factory for billing budgets, that works in a similar way to factories in other modules: a specific folder is searched for YAML files, which contain one budget description per file. The file name is used to generate the key of the resulting map of budgets, which is merged with the one coming from the `budgets` variable. The YAML files support the same type of the `budgets` variable.
```
LucaPrete, line 218: very minor. Can we rephrase this? It's really difficult to understand imho

modules/billing-account/factory.tf
```diff
@@ -0,0 +1,98 @@
...
+ * limitations under the License.
+ */
+
+locals {
+  _factory_data = {
+    for f in fileset("${local._factory_path}", "**/*.yaml") :
+    trimsuffix(f, ".yaml") => yamldecode(file("${local._factory_path}/${f}"))
+  }
+  _factory_path = var.factory_config.budgets_data_path
+  factory_budgets = {
```
LucaPrete, line 23: it's quite difficult to read. I'd consider splitting this in multiple locals

ludoo: what is difficult to read Luca?

modules/billing-account/README.md
ludoo: try to read it slowly Luca :)

PR #1823, opened 2023-10-28
Add end-to-end tests for project module

wiktorn:
## `modules/project` examples
Modifying almost all examples in project module allowing to be run as end-to-end tests. The only one left is the one requiring org.

Interesting finding - module fails if no billing account is attached, unless you provide `auto_create_network = true`. This is probably due to the fact, that provider removes the `default` network after project is created and this requires `compute` API, and this is only allowed if billing is attached.

```
Error: Error enabling the Compute Engine API required to delete the default network: failed to enable services: failed on request preconditions: googleapi: Error 400: Billing account for project '***' is not found. Billing must be enabled for activation of service(s) 'compute.googleapis.com,compute.googleapis.com,compute.googleapis.com' to proceed
```

I decided to add billing account to all examples

Added one big example using `project_create = false` to test (hopefully) all cases with data resource in use.

## Bug fixes for `modules/project`
* log sink creation failed as it didn't wait to activate `logging` service

## E2E framework updates
Modifying E2E test framework to provide unique prefix on each test run. This is very useful here, and will be useful in any other areas where global uniqueness is required.

This results in dependency on providing `TF_VAR_prefix` variable hence making the `*.tfvars` approach even less comfortable and I lean towards removing it. The problem is, that now the `TF_VAR_prefix` is needed both in Terraform code and in Python code.

Added new required resources to test harness:
* KMS key id
* group_email

Added support for `files=` clause in `# tftest` line to end-to-end tests, to make sure that we are testing the same thing whether running standard `testes/examples` and `test/examples_e2e`

[Sample E2E tests run](https://github.com/fabric-team/fast-e2e/actions/runs/6682356647)

---
**Checklist**


ludoo:
Will take me a while to go through this, plz be patient. :)

tests/examples/test_plan.py
```diff
@@ -23,6 +23,15 @@
...
+  if line is not None:
+    requested_files = line.split(',')
+    for f in requested_files:
+      destination = test_path / example.files[f].path
+      destination.parent.mkdir(parents=True, exist_ok=True)
+      destination.write_text(example.files[f].content)
+
+
 def test_example(plan_validator, tmp_path, example):
   if match := COUNT_TEST_RE.search(example.code):
```
ludoo, line 36: nit: can we use named groups in `COUNT_TEST_RE`? it makes it easier to follow the match logic. Feel free to resolve ofc, don't hold the PR for this.

ludoo (approved):
very nice, thanks Wiktor

tests/examples/test_plan.py
wiktorn: Added named groups for now. In future, I think it warrants proper parsing function that returns NamedTuple

PR #1825, opened 2023-10-30
Handling SQL IP address issue

aurelienlegrand:
I have a customer who is facing issue with SQL instances without IP addresses, trying to handle this unusual case.

### DIFF
--- a/blueprints/cloud-operations/network-dashboard/src/plugins/discover-cai.py
+++ b/blueprints/cloud-operations/network-dashboard/src/plugins/discover-cai.py
@@ -212,17 +212,14 @@ def _handle_routes(resource, data):
 
 def _handle_sql_instances(resource, data):
   'Handles cloud sql instance type resource data.'
-  ip_addressses = []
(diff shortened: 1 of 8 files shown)


blueprints/cloud-operations/network-dashboard/src/plugins/discover-cai.py
```diff
@@ -212,14 +212,17 @@ def _handle_routes(resource, data):
 
 def _handle_sql_instances(resource, data):
   'Handles cloud sql instance type resource data.'
+  ip_addressses = []
+  if data.get('ipAddresses'):
+    ip_addressses = [
+        i['ipAddress']
+        for i in data.get('ipAddresses')
```
wiktorn, line 219: You can do:
```python
for i in data.get('ipAddresses', [])
```
And keep the code where it was. (without if)

aurelienlegrand:
Thanks for your feedback, your solution is cleaner!
I updated the PR.

PR #1826, opened 2023-10-30
Add public_access_prevention field to GCS module

devuonocar:
Added `public_access_prevention` field to GCS module.

---
**Checklist**


modules/gcs/variables.tf
```diff
@@ -221,6 +221,12 @@ variable "project_id" {
   type        = string
 }
 
+variable "public_access_prevention" {
+  description = "Prevents public access to a bucket. Acceptable values are inherited or enforced. If inherited, the bucket uses public access prevention, only if the bucket is subject to the public access prevention organization policy constraint."
+  type        = string
+  default     = "inherited"
```
wiktorn, line 227: I guess we can have null here as default and use provider default for that.

wiktorn (approved):
Thank you for this contribution. Just a minor remark.

modules/gcs/variables.tf
devuonocar: Done! 😃

PR #1833, opened 2023-11-01
Net VPC Peering: added stack_type field

cmalpe:
Added the stack_type field for the "google_compute_network_peering" terraform resource. Link to documentation: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network_peering#stack_type

### DIFF
--- a/modules/net-vpc-peering/variables.tf
+++ b/modules/net-vpc-peering/variables.tf
@@ -63,5 +63,9 @@ variable "routes_config" {
 variable "stack_type" {
   description = "IP version(s) of traffic and routes that are allowed to be imported or exported between peer networks. Possible values: IPV4_ONLY, IPV4_IPV6."
   type        = string
-  default     = "IPV4_ONLY"
+  default     = null
+  validation {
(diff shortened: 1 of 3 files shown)


modules/net-vpc-peering/variables.tf
```diff
@@ -59,3 +59,9 @@ variable "routes_config" {
   nullable = false
   default  = {}
 }
+
+variable "stack_type" {
```
ludoo, line 63: Can you add a validation rule to make sure only allowed values can be specified, and default to `null`?

cmalpe: Hi @ludoo I will add the validation for the possible values. But for the null part, it says in the documentation that the default value for stack_type is "IPV4_ONLY". Should I still keep it the default as null?

ludoo: In general, when the provider has a default value we pass `null` so we defer to the provider by default.

ludoo (approved):

PR #1836, opened 2023-11-01
Avoid map-related casting errors in project factory

ludoo:
Using a map in the `for` loop that reads and decodes the project YAML data files can trigger unpredictable errors, as Terraform tries to enforce the same type for every map value. This cannot work in our case since most of the project attributes are optional, or accept a variable number of values (e.g. variable number of project services).

The solution implemented here is to replace the map with two lists: one for the YAML file names, and one for the actual raw data decoded from each YAML file. They are then recombined as map keys and values when the final project data structure is assembled for each element.

This PR also simplifies the variable used for the data path by removing the option to pass a structure with project data. This has never been needed so far and only complicates the module.

An additional test is added to the README to replicate an actual use case that is fixed by this change.


wiktorn (approved):
Looks good. 

I did some experimentation here just to better understand Terraform internals. It looks like TF is verifying the type within the map only in trenary operator. So the minimal fix for added example can look like that:
```diff
diff --git a/blueprints/factories/project-factory/factory.tf b/blueprints/factories/project-factory/factory.tf
index e0351a0a..53017266 100644
--- a/blueprints/factories/project-factory/factory.tf
+++ b/blueprints/factories/project-factory/factory.tf
@@ -16,10 +16,8 @@
 
 locals {
   _data = (
-    var.factory_data.data != null
-    ? var.factory_data.data
-    : {
-      for f in fileset("${local._data_path}", "**/*.yaml") :
+    {
+      for f in fileset(local._data_path, "**/*.yaml") :
       trimsuffix(f, ".yaml") => yamldecode(file("${local._data_path}/${f}"))
     }
   )
```

That is, just refactoring the variable and getting rid of `factory.data` is sufficient. :exploding_head:

ludoo:
> Looks good.
> 
> I did some experimentation here just to better understand Terraform internals. It looks like TF is verifying the type within the map only in trenary operator. So the minimal fix for added example can look like that:
> 
> ```diff
> diff --git a/blueprints/factories/project-factory/factory.tf b/blueprints/factories/project-factory/factory.tf
> index e0351a0a..53017266 100644
> --- a/blueprints/factories/project-factory/factory.tf
> +++ b/blueprints/factories/project-factory/factory.tf
> @@ -16,10 +16,8 @@
>  
>  locals {
>    _data = (
> -    var.factory_data.data != null
> -    ? var.factory_data.data
> -    : {
> -      for f in fileset("${local._data_path}", "**/*.yaml") :
> +    {
> +      for f in fileset(local._data_path, "**/*.yaml") :
>        trimsuffix(f, ".yaml") => yamldecode(file("${local._data_path}/${f}"))
>      }
>    )
> ```
> 
> That is, just refactoring the variable and getting rid of `factory.data` is sufficient. 🤯

Ach, I had set automerge. Can you send a patch to implement yours, which is much more readable?

PR #1844, opened 2023-11-07
Allow disabling IAM for sink identity in resource manager modules

apichick:
Added back sink iam flag as module users might not have access to the sink destination and the role might need to be granted somewhere else

---
**Checklist**


modules/folder/variables.tf
```diff
@@ -117,6 +117,7 @@ variable "logging_sinks" {
     disabled             = optional(bool, false)
     exclusions           = optional(map(string), {})
     filter               = string
+    iam                  = optional(bool, true)
```
ludoo, line 120: maybe rename to `create_iam_binding`?

ludoo (approved):

modules/folder/variables.tf
apichick: shall I then change it in all the modules (organization, folder and project)?

ludoo: ah, project already has it, ignore my comment then :)

PR #1845, opened 2023-11-07
Extend `cluster_autoscaling` fields in gke-cluster-standard

anthonyhaussman:
## [feat(gke-cluster-standard): Add feature to setup cluster_autoscaling](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1845/commits/9eca7a68e21c3779e1700a900953348ec42fdd06)

 - Set the `autoscaling_profile` if `OPTIMIZE_UTILIZATION` is needed
 - Add the possibility to set `disk_size` and `disk_type`
 - Add configuration of `management` and `shielded_instance_config`

Each of these options is optional and set to their default value if not set.

## [feat(gke-cluster-standard): Add GPUs setup feature for `cluster_autoscaler`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1845/commits/de7b7f3bb320395404ef76e23191a6ec668d1d1f)

Add the possibility to setup `gpu_resources` into node auto-provisioning.

---
**Checklist**


modules/gke-cluster-standard/variables.tf
```diff
@@ -38,11 +38,22 @@ variable "backup_configs" {
 variable "cluster_autoscaling" {
   description = "Enable and configure limits for Node Auto-Provisioning with Cluster Autoscaler."
   type = object({
+    autoscaling_profile = optional(string)
     auto_provisioning_defaults = optional(object({
       boot_disk_kms_key = optional(string)
+      disk_size         = optional(number)
+      disk_type         = optional(string)
```
juliocc, line 45: add validations to `autoscaling_profile` and `disk_type` to ensure only valid values are specified.

modules/gke-cluster-standard/variables.tf
```diff
@@ -52,6 +63,11 @@ variable "cluster_autoscaling" {
       min = number
       max = number
     }))
+    gpu_resources = optional(list(object({
+      resource_type = string
+      min           = number
+      max           = number
```
juliocc, line 69: This is inconsistent with the two othe limits we already implement (cpu/mem). You probably want to skip the `resource_type` key and just hard-code it in the block if the users provides a value for `gpu_resources`

juliocc (changes requested):

modules/gke-cluster-standard/variables.tf
anthonyhaussman: Done in e2c030d8f7521f4f478a0058f7e9f40f69ca7a67 :slightly_smiling_face:

modules/gke-cluster-standard/variables.tf
anthonyhaussman: Done in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/commit/e2c030d8f7521f4f478a0058f7e9f40f69ca7a67 🙂

modules/gke-cluster-standard/variables.tf
anthonyhaussman: Well for GPU, there can be several `resourceType` such as `nvidia-tesla-k80` or  `nvidia-tesla-v100`.
There's a long list of GPU types so I think it's better to let a user choose which one to use.

Ref:
 - https://cloud.google.com/kubernetes-engine/docs/how-to/node-auto-provisioning#gpu_limits

juliocc: I see. We can leave like this for now but I think eventually we'll want to use a single field for all limits.

juliocc (approved):

juliocc:
@anthonyhaussman ;can you fix the failing tests?

anthonyhaussman:
> @anthonyhaussman ;can you fix the failing tests?

@juliocc It should be good now :pray:

anthonyhaussman:
@juliocc I had some trouble finding the problem and reproducing it.
But I have found the problem and pushed the fix in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/commit/e74e0caced62edcb6c351e3f79eb4cb9f300d7bb. 

Sorry for the trouble. :pray:

PR #1846, opened 2023-11-08
Add support for IAM to vpc sc module

ludoo:


sruffilli (approved):

modules/vpc-sc/README.md
```diff
@@ -49,6 +51,23 @@ module "test" {
 # tftest modules=1 resources=1 inventory=scoped-access-policy.yaml
 ```
 
+#### Access policy IAM
+
+The usual IAM interface is also implemented here, and can be used with service accounts or user principals:
```
lcaggio, line 56: Groups are also supported here.

lcaggio (approved):

modules/vpc-sc/README.md
ludoo: no they're not, you can add a group but it does not work, as stated in the docs

PR #1860, opened 2023-11-14
Organization module end-to-end tests

wiktorn:
* added tag `serial` to mark tests to be run serially
* always run tests using `loadgroup` distribution to make use of `serial` tag
* added end-to-end tests for organization, not adding to custom constraints as the name has to be unique
* fixed granting custom roles created in the same module call

[E2E tests run](https://github.com/GoogleCloudPlatform/fast-e2e/actions/runs/6865343619)

---
**Checklist**

### DIFF
--- a/modules/organization/iam.tf
+++ b/modules/organization/iam.tf
@@ -46,6 +46,7 @@ resource "google_organization_iam_binding" "authoritative" {
   org_id   = local.organization_id_numeric
   role     = each.key
   members  = each.value
+  depends_on = [google_organization_iam_custom_role.roles]
 }
 
 resource "google_organization_iam_binding" "bindings" {
@@ -61,6 +62,7 @@ resource "google_organization_iam_binding" "bindings" {
       description = each.value.condition.description
     }
   }
+  depends_on = [google_organization_iam_custom_role.roles]
 }
 
 resource "google_organization_iam_member" "bindings" {
@@ -76,4 +78,5 @@ resource "google_organization_iam_member" "bindings" {
       description = each.value.condition.description
     }
   }
+  depends_on = [google_organization_iam_custom_role.roles]
 }
--- a/modules/organization/iam.tf
+++ b/modules/organization/iam.tf
@@ -42,10 +42,10 @@ resource "google_organization_iam_custom_role" "roles" {
 }
 
 resource "google_organization_iam_binding" "authoritative" {
-  for_each = local.iam
-  org_id   = local.organization_id_numeric
-  role     = each.key
-  members  = each.value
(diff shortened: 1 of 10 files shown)


modules/organization/iam.tf
```diff
@@ -42,10 +42,11 @@ resource "google_organization_iam_custom_role" "roles" {
...
 resource "google_organization_iam_binding" "authoritative" {
-  for_each = local.iam
-  org_id   = local.organization_id_numeric
-  role     = each.key
-  members  = each.value
+  for_each   = local.iam
+  org_id     = local.organization_id_numeric
+  role       = each.key
+  members    = each.value
+  depends_on = [google_organization_iam_custom_role.roles]
```
ludoo, line 49: I'm not entirely sure about this. You need the role admin role to be able to create a custom role, so this dependency might raise issues in specific cases. I remember looking at this for FAST, and deciding the custom roles should depend in bindings, but not viceversa. Remember we're at org level, so unless you have role admin role, you cannot create custom roles.

ludoo (changes requested):
I would remove the added dependency, logically it makes little sense at org level and I remember it triggering issues with FAST when we tried in the past.

modules/organization/iam.tf
wiktorn: This dependency is "not null", only if you create custom roles. Custom role doesn't have any explicit dependencies, so it won't loop back here. In usual case, this will be empty, so I assume this should not raise any issues.

Otherwise this example:
```hcl
module "org" {
  source          = "./fabric/modules/organization"
  organization_id = var.organization_id
  custom_roles = {
    "myRole" = [
      "compute.instances.list",
    ]
  }
  iam = {
    (module.org.custom_role_id.myRole) = ["group:${var.group_email}"]
  }
}
```
Should be considered as "not supported use", because this may fail.

I can give a FAST end-to-end test on this branch to check if it breaks anything, WDYT?

ludoo: yep, the part I'm worried about is not assigning a custom role, but creating one at org level which requires having the role admin role assigned

ludoo: if both fail with/without, priority goes to the FAST use case ofc (so no dependency)

wiktorn: rolled back this change. I realized that the example provides explicit dependency on the role so it should work (and it is, as far as I tested).

For FAST - we have prerequisite that when running `0-bootstrap` the identity running bootstrap already has organization role admin.

ludoo (approved):
Thanks Wiktor!

PR #1868, opened 2023-11-17
Fix/dlpagent

ddaluka:
When DLP API service is enabled for a project, we also need its service agent to be available for further IAM roles assignment. 

dlp.googleapis.com API supports the identity creation using gcloud beta services identity create --service dlp.googleapis.com

---
**Checklist**


wiktorn (changes requested):
If you need to have DLP service agent force-created, add to [service-agents.yaml](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/project/service-agents.yaml#L170):
```
jit: true
```

For DLP

ddaluka:
> If you need to have DLP service agent force-created, add to [service-agents.yaml](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/project/service-agents.yaml#L170):
> 
> ```
> jit: true
> ```
> 
> For DLP. Also, bare in mind, that services force-created may miss their role, in this case it will be `roles/dlp.serviceAgent`, so this is not a perfect solution.

Thanks, I have removed the explicit identity creation block and have set jit as true . I guess , we can assign the service agent role if required once agent is available.

wiktorn:
> Thanks, I have removed the explicit identity creation block and have set jit as true . I guess , we can assign the service agent role if required once agent is available.

Yes, we have this documented [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/project#service-identities-requiring-manual-iam-grants)

Can you add DLP there too? Thanks.

wiktorn:
@ddaluka There was still one blueprint tests that were failing due to this change, will you be able to fix that, or shall I help you with that?

wiktorn (approved):

ddaluka:
> @ddaluka There was still one blueprint tests that were failing due to this change, will you be able to fix that, or shall I help you with that?

Thanks. I have fixed the issue

wiktorn:
Thank you for fixing those :heart:

PR #1874, opened 2023-11-21
Added PSC support to CloudSQL Module

luigi-bitonti:
In this PR a variable to enable PSC on CloudSQL Instance was introduced. The design of the variables that manage the network has been improved, by introducing _network_config_ variable.
---
**Checklist**


modules/cloudsql-instance/variables.tf
```diff
@@ -172,6 +172,7 @@ variable "name" {
 variable "network" {
   description = "VPC self link where the instances will be deployed. Private Service Networking must be enabled and configured in this VPC."
   type        = string
+  default     = ""
```
wiktorn, line 175: I don't we need need such a default. The other thing is - if I understand correctly, instance can use either PSA (then we need `network` + optional `allocated_ip_range`) or PSC (then we need `allowed_consumer_projects` and maybe `psc_enabled`). or the instance does not have private IP at all. I think we can improve the interface here by grouping all of those under one variable like `network_config`.

Something similar to what we have [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/compute-vm/variables.tf#L315)  with either hourly, daily or weekly schedule.

wiktorn (changes requested):
Thank you for this. I see some space for improvement in the interface for this module. Are you willing to take this? I can help with the design .

modules/cloudsql-instance/variables.tf
```diff
@@ -195,6 +196,15 @@ variable "project_id" {
   type        = string
 }
 
+variable "psc_config" {
+  description = "Whether Private Service Connect (PSC) connectivity is enabled for this instance and list of consumer projects that are allow-listed for PSC connections to this instance."
+  type = object({
+    allowed_consumer_projects = optional(list(string), [])
+    psc_enabled               = optional(bool, true)
```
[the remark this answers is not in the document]
luigi-bitonti: https://cloud.google.com/sql/docs/mysql/configure-private-service-connect#create-cloud-sql-instance-psc-enabled
Following this doc, _consumer projects_ I think that an empty value means that no project has access.
I think it's better to leave the variable as it is because you can decide, for some reason, to remove a project and in the future add a new one, without disabing the PSC.

modules/cloudsql-instance/variables.tf
juliocc: I'ts not clear from the docs but my understanding is that you still need the network (i.e. PSA) and you can optionally enable PSC on top of it:
> At least ipv4_enabled must be enabled or a private_network must be configured

If that's correct,n the this variable should remain mandatory

@luigi-bitonti can you confirm?

juliocc (changes requested):
The PR that introduced PSC also added two additional (undocumented) fields:
- [dns_name](https://github.com/GoogleCloudPlatform/magic-modules/pull/8580/files#diff-7743a2a7560e6c866ca3fa681e1cc709950f50fed72ba4d6532f789acead0c89R900)
- [psc_service_attachment_link](https://github.com/GoogleCloudPlatform/magic-modules/pull/8580/files#diff-7743a2a7560e6c866ca3fa681e1cc709950f50fed72ba4d6532f789acead0c89R895) we should add this as an output

modules/cloudsql-instance/variables.tf
luigi-bitonti: If you enable PSC and private_network, you will receive an error. This is an open issue on the provider's source code that talks about it: : https://github.com/hashicorp/terraform-provider-google/issues/16257

luigi-bitonti:
> The PR that introduced PSC also added two additional (undocumented) fields:
> 
> * [dns_name](https://github.com/GoogleCloudPlatform/magic-modules/pull/8580/files#diff-7743a2a7560e6c866ca3fa681e1cc709950f50fed72ba4d6532f789acead0c89R900) not sure what this is but sounds relevant to PSC
> * [psc_service_attachment_link](https://github.com/GoogleCloudPlatform/magic-modules/pull/8580/files#diff-7743a2a7560e6c866ca3fa681e1cc709950f50fed72ba4d6532f789acead0c89R895) we should add this as an output

Thanks, outputs added to the module. _dns_name_ is correlated with PSC.

modules/cloudsql-instance/variables.tf
luigi-bitonti: Hi @wiktorn, as you suggest I tried to introduce a new variable, network_config. Let me know if you think it's ok.

modules/cloudsql-instance/main.tf
```diff
@@ -80,6 +80,14 @@ resource "google_sql_database_instance" "primary" {
           value = network.value
         }
       }
+      dynamic "psc_config" {
```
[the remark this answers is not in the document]
luigi-bitonti: Done!

modules/cloudsql-instance/outputs.tf
```diff
@@ -34,6 +34,11 @@ output "connection_names" {
   }
 }
 
+output "dns_name" {
```
wiktorn, line 37: I think we should output dns names for all instances too, like in `id`, `ids`; `ip`, `ips` etc.

modules/cloudsql-instance/variables.tf
```diff
@@ -169,9 +148,26 @@ variable "name" {
...
+      primary = optional(string)
+      replica = optional(string)
+    }))
+    psc_config = optional(object({
+      allowed_consumer_projects = optional(list(string), [])
+      psc_enabled               = optional(bool, true)
+    }))
+  })
+  validation {
+    condition     = (var.network_config.private_network != null ? 1 : 0) + (var.network_config.psc_config != null ? 1 : 0) == 1
```
wiktorn, line 168: This prevents having instance with no private network connectivity (which is rare, but still valid use case) - i.e. only public IP address. Also, this should be reflected in description.

modules/cloudsql-instance/variables.tf
wiktorn: Thank you for this

modules/cloudsql-instance/variables.tf
luigi-bitonti: You're right, I fixed this.

luigi-bitonti: You're righ, I fixed this.

modules/cloudsql-instance/outputs.tf
luigi-bitonti: Done (: Also for psc_service_attachment_link

modules/cloudsql-instance/variables.tf
```diff
@@ -169,9 +148,26 @@ variable "name" {
   type        = string
 }
 
-variable "network" {
-  description = "VPC self link where the instances will be deployed. Private Service Networking must be enabled and configured in this VPC."
-  type        = string
+variable "network_config" {
```
wiktorn, line 151: One more idea here: can we maybe deviate a bit from provider naming (I really don't like `enable_ipv4`) here and have something like that:
```hcl
object({
  authorized_networks = optional(map(string))
  ...
  connectivity = object({
    public_ip  = optional(bool, false)
    psa_config = optional(object({
      private_network     = string
      allocated_ip_ranges = optional(object({
        primary = optional(string)
        replica = optional(string)
      })
    })
    psc_allowed_consumer_projects = optional(list(string))
  })
  ...
})
```

With semantics for `psc_config` that if it is null, then PSC is disabled, if not null, then enabled.

Two non-optional fields above will also make apparent, that user needs to make a choice here and simplify validation a bit.

@juliocc @luigi-bitonti WDYT?

juliocc: I like this. `ipv4_enabled` is really confusing indeed.

wiktorn: And it should be `public_ipv4`  instead of `public_ip` in anticipation of IPv6 support.

luigi-bitonti: Thank you for the suggestion, I agree with you. I'll made the changes ASAP.

luigi-bitonti: Just pushed.

modules/cloudsql-instance/README.md
```diff
@@ -46,9 +52,15 @@ module "db" {
 
 ```hcl
 module "db" {
-  source           = "./fabric/modules/cloudsql-instance"
-  project_id       = var.project_id
-  network          = var.vpc.self_link
+  source     = "./fabric/modules/cloudsql-instance"
```
wiktorn, line 55: Last thing we miss here, is an example with PSC instead of PSA. Please add if you have bandwidth, or we can add this as a follow-up.

wiktorn (approved):

modules/cloudsql-instance/README.md
luigi-bitonti: Done.

juliocc (approved):

PR #1875, opened 2023-11-21
feat(gke-cluster-standard): Add optional `boot_disk_kms_key` into node_config

anthonyhaussman:
Add optional setup to add `boot_disk_kms_key` into the `node_config`.

Without it, configuring a standard GKE cluster with the policy `constraints/gcp.restrictNonCmekServices` returns an error:
```
│ Error: googleapi: Error 400: Failed precondition: Constraint `constraints/gcp.restrictNonCmekServices` violated for `projects/XXXXXXXX` attempting to create a resource without specifying a KMS CryptoKey.
│ Details:
│ [
│   {
│     "@type": "type.googleapis.com/google.rpc.RequestInfo",
│     "requestId": "0xde962b1494496ff1"
│   }
│ ]
│ , failedPrecondition
```

---
**Checklist**


modules/gke-cluster-standard/main.tf
```diff
@@ -128,7 +129,7 @@ resource "google_container_cluster" "cluster" {
       dynamic "auto_provisioning_defaults" {
         for_each = var.cluster_autoscaling.auto_provisioning_defaults != null ? [""] : []
         content {
-          boot_disk_kms_key = var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key
+          boot_disk_kms_key = var.boot_disk_kms_key != null ? var.boot_disk_kms_key : var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key
```
ludoo, line 132: why don't you use the existing field in auto provisioning defaults?

anthonyhaussman: As `boot_disk_kms_key` is a "root" variable, it makes sense to set the default KMS to the "root" variable if it's available.
I have changed the logic in my last commit to using in `auto_provisioning_defaults` first the variable ` var.cluster_autoscaling.auto_provisioning_defaults.boot_disk_kms_key` if it exists and fallback to `var.boot_disk_kms_key` if it exists.

What do you think?

ludoo: I don't think we need two variables for the same thing. :) And in general we always try to keep our variable spaces small. I am closing this, but thanks anyway.

anthonyhaussman:
@ludoo,
Can I know why did you close the PR? :confused: 
Is that possible to have an explanation for this? This is a blocker on our side to push strict Org policies.

modules/gke-cluster-standard/main.tf
anthonyhaussman: But without the setting into the `node_config` configuration, the constraint policy `constraints/gcp.restrictNonCmekServices` blocks everything and the module can't be used at all.

Do we have a solution?

ludoo:
Again: why don't you use the existing variable for the same thing?

ludoo:
BTW you're free to fork and do your own changes, of course.

ludoo:
Had a chat with @wiktorn and while this is indeed a problem, the proposed solution does not fit into our design patterns (and I did an awful job at explaining it, sorry for that): adding a top-level variable for a single value is a "waste of variable space" which we always try to avoid.

The right approach is to add a top-level variable for all the current `node_config` attributes (service account, tags, etc.) and add the KMS key there. Then using that in autoscale config with a condition much like in your PR, probably.

So we can now either

- update this PR to do the above, if you still feel like it, and also check that the autopilot module has the same changes
- do it ourselves

In the meantime, we'll turn this into an issue for tracking. Thanks for your patience, and for sticking with it. :)

anthonyhaussman:
> Had a chat with @wiktorn and while this is indeed a problem, the proposed solution does not fit into our design patterns (and I did an awful job at explaining it, sorry for that): adding a top-level variable for a single value is a "waste of variable space" which we always try to avoid.
> 
> The right approach is to add a top-level variable for all the current `node_config` attributes (service account, tags, etc.) and add the KMS key there. Then using that in autoscale config with a condition much like in your PR, probably.
> 
> So we can now either
> 
> * update this PR to do the above, if you still feel like it, and also check that the autopilot module has the same changes
> * do it ourselves
> 
> In the meantime, we'll turn this into an issue for tracking. Thanks for your patience, and for sticking with it. :)

Gonna try to propose something in this PR. 
Thanks for your explanations. :pray:

ludoo:
Thanks for flagging this! PR is incoming, I had a free half hour. :)

anthonyhaussman:
> Thanks for flagging this! PR is incoming, I had a free half hour. :)

Thanks a lot for the reactivity :pray: 
It helps a lot! Many thanks! :pray:

ludoo:
Thank you for being patient despite my pushback! :)

PR #1883, opened 2023-11-23
F5 deployment blueprint

LucaPrete:
The blueprint shows a private IPv4/IPv6 F5 deployment.

---
**Checklist**


blueprints/third-party-solutions/f5-bigip/f5-bigip-ha-active-deployment/main.tf
```diff
@@ -0,0 +1,130 @@
...
+
+locals {
+  project_id = (
+    var.project_create
+    ? module.project.project_id
+    : var.project_id
+  )
+}
+
+data "google_client_openid_userinfo" "provider_identity" {}
```
ludoo, line 25: Luca, can you make this optional so we can run tests (which will have to be added here)?

I also think that this is fine for testing, but not what you want if you deploy the blueprint for actual usage. In that case IAM is taken care of at the folder or project level, and would use groups anyway.

ludoo (changes requested):
Please add a test example at the bottom of the README. :)

ludoo (approved):
chatted f2f, thanks Luca

PR #1898, opened 2023-12-03
Use unique names for logging buckets in examples

wiktorn:
Logging bucket name can be reused only after 7 days (when it is actually deleted). When different tests reuse the same name, the ones that are executed as 2nd and later will fail with message:
```
Error updating Logging Bucket Config [...]: googleapi: Error 400: Buckets 
must be in an ACTIVE state to be modified
```

As their actual state is:
```
lifecycleState: DELETE_REQUESTED
```

---
**Checklist**

### DIFF
--- a/modules/folder/README.md
+++ b/modules/folder/README.md
@@ -225,7 +225,7 @@ module "bucket" {
   source      = "./fabric/modules/logging-bucket"
   parent_type = "project"
   parent      = var.project_id
-  id          = "bucket"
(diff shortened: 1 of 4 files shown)


wiktorn:
[E2E tests run](https://github.com/GoogleCloudPlatform/fast-e2e/actions/runs/7076178573)

ludoo (approved):

PR #1904, opened 2023-12-06
gke-cluster-standard : Add possibility to enable image streaming feature at cluster level

noony:
Module modified : gke-cluster-standard

This PR add a variable to enable image streaming feature at cluster level.

following documentation of terraform object `` : [link](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#node_config_defaults)

---
**Checklist**


modules/gke-cluster-standard/variables.tf
```diff
@@ -135,6 +135,7 @@ variable "enable_features" {
     fqdn_network_policy  = optional(bool, false)
     gateway_api          = optional(bool, false)
     groups_for_rbac      = optional(string)
+    image_streaming      = optional(bool, false)
```
ludoo, line 138: I'm a bit torn on whether we should stick to the provider naming (`gcfs`), but this is probably the right way to do it.

PR #1905, opened 2023-12-07
gke-cluster-standard : Support upgrade_settings for node auto provisioner

noony:
This PR modifies gke-cluster-standard list of changes : 

## Support Upgrade settings for Node Auto Provisionner
We can now define a strategy for NAP nodes upgrade SURGE / BLUE_GREEN and related configurations for each modes.

google_container_cluster documentation [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#upgrade_settings)

## Set max_surge by default to 1 & max_unavailable 0 when NAP enabled
Currently node auto provisionner default for max_surge surge & max_unavailable are 0. This result in an invalid configuration for default settings in nap config.
Screenshot 2023-12-07 at 01 18 30

When `upgrade_settings` is not defined we declare by default a max_surge 1 and max_unavailable 0 now.

---
**Checklist**


modules/gke-cluster-standard/variables.tf
```diff
@@ -54,6 +54,19 @@ variable "cluster_autoscaling" {
         integrity_monitoring = optional(bool, true)
         secure_boot          = optional(bool, false)
       })
+      upgrade_settings = optional(object({
```
ludoo, line 57: We need to redesign this variable.

`strategy` only has two values with `"SURGE"` being the default, and our approach is to always turn those to booleans to clearly formulate the choice to module users, something like

```hcl
upgrade_settings = optional(object({
  blue_green_strategy = optional(bool, false)
```

But then the choice of strategy defines which other fields are possible, so what we typically do in those cases (health checks in load balancer modules for example) is to have multiple optional objects, and validation to ensure only one is selected, which would also redefine `blue_green_settings`. Something like

```hcl
upgrade_settings = optional(object({
  blue_green = optional(object({
    node_pool_soak_duration = optional(string)
    standard_rollout_policy = optional(object({
      batch_percentage    = optional(number)
      batch_node_count    = optional(number)
      batch_soak_duration = optional(string)
    }))
  })
  surge = optional(object({
    max = optional(number)
    unavailable = optional(number)
  })
})
# add validation rule to ensure only one is present if upgrade settings is defined
```

Then if a user set this in the module

```hcl
upgrade_settings = {}
```

you would internally need to assume surge as default.

modules/gke-cluster-standard/main.tf
```diff
@@ -149,6 +149,25 @@ resource "google_container_cluster" "cluster" {
               enable_secure_boot          = var.cluster_autoscaling.auto_provisioning_defaults.shielded_instance_config.secure_boot
             }
           }
+          upgrade_settings {
```
ludoo, line 152: This entire block should be dynamic and conditional on `upgrade_settings` being defined.

ludoo (changes requested):

modules/gke-cluster-standard/main.tf
juliocc: Also considering using iterator (or the default block name) to make the nested blocks easier to read

ludoo (approved):

PR #1907, opened 2023-12-07
Improve project factory example

ludoo:
Show how to manage service identities roles on the Shared VPC host project.

### DIFF
--- a/blueprints/factories/project-factory/README.md
+++ b/blueprints/factories/project-factory/README.md
@@ -57,7 +57,7 @@ module "project-factory" {
   # location where the yaml files are read from
   factory_data_path = "data"
 }
-# tftest modules=7 resources=26 files=prj-app-1,prj-app-2,prj-app-3 inventory=example.yaml
+# tftest modules=7 resources=31 files=prj-app-1,prj-app-2,prj-app-3 inventory=example.yaml
 ```
 
 ```yaml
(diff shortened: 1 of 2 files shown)


modules/project/README.md
```diff
@@ -296,13 +308,47 @@ module "service-project" {
...
-    host_project       = module.host-project.project_id
+    host_project = module.host-project.project_id
+    # reuse the list of services from the module's outputs
     service_iam_grants = module.service-project.services
   }
 }
 # tftest modules=2 resources=9 inventory=shared-vpc-auto-grants.yaml e2e
 ```
 
+In specific cases it might make sense to selectively grant the `compute.networkUser` role for service identities at the subnet level, and while that is best done via org policies it's also supported by this module.
```
wiktorn, line 319: It is worth adding:
"In this example, Compute service identity will be granted `compute.networkUser` in the `gce` subnet defined in `europe-west1` region."

modules/project/README.md
```diff
@@ -228,9 +228,21 @@ This table lists all affected services and roles that you need to grant to servi
...
-You can enable Shared VPC Host at the project level and manage project service association independently.
+Project service association for VPC host projects can be
+
+- autoritatively managed in the host project by enabling Shared VPC and specifying the set of service projects, or
+- additively managed in service projects by by enabling Shared VPC in the host project and then "attaching" each service project independently
+
+IAM bindings in the host project for API service identities can be managed from service projects in two different ways:
+
+- via the `service_identity_iam` attribute, by specifying the set of roles and service agents
+- via the `service_iam_grants` attribute that leverages a [fixed list of roles for each service](./sharedvpc-agent-iam.yaml), by specifying a list of services
```
wiktorn, line 241: To make this list exhaustive, can we also add `service_identity_subnet_iam` here?

PR #1899, opened 2023-12-03
Read-only service accounts for automation and CI/CD

ludoo:
WIP, design document in this PR has the rationale and explanation of changes.

This PR also

- adds a custom roles factory to the organization and project modules
- refactors the factory variables in organization, folder and project modules so they use our latest standard
- moves FAST stage 0 custom roles to YAML files passed to the organization module's factory
- removes the (now stale) `custom_role_names` variable from FAST stage 0
- refactors the FAST CI/CD workflows

Test errors will be expected until this has been completed.

PR #1912, opened 2023-12-09
Custom role factories for organization and project modules

ludoo:
Backported from #1899 to reduce the amount of changes in that PR.


modules/organization/iam.tf
```diff
@@ -32,13 +52,27 @@ locals {
...
 
+# we use a different key for custom roles to allow referring to the role alias
+# in Terraform, while still being able to define unique role names
+
+check "custom_roles" {
+  assert {
+    condition = (
+      length(local.custom_roles) == length({
+        for k, v in local.custom_roles : v.name => null
+      })
```
juliocc, line 63: nit: I'd be explicit and use `distinct` instead of putting the names in a map: `distinct([for _, v in local.custom_roles: v.name])`

juliocc (approved):
Nice! And thanks for updating the resman modules to use `factories_config`

Left on small comment but feel free to ignore.

modules/organization/iam.tf
ludoo: well, I want to fail as the key clash might imply different values, and generate issues when the role is differently implemented

PR #1913, opened 2023-12-09
net-lb-app-ext-e2e

andybubu:
---
**Checklist**

I applicable, I acknowledge that I have:
- [x] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [x] Ran `terraform fmt` on all modified files
- [x] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [x] Made sure all relevant tests pass

PR #1914, opened 2023-12-10
Allow per-module terraform fixtures

juliocc:
This PR introduces "terraform fixtures" which allow defining common terraform resources (or modules) that are deployed together with a given example. This allows:
- Describing all the required infrastructure for a given example
- Reusing common infrastructure used by more than one example (e.g. instance groups in the load balancing modules)
- Not polluting examples with repetitive boilerplate


modules/net-lb-app-ext/README.md
```diff
@@ -45,13 +45,13 @@ module "glb-0" {
...
-        { backend = "projects/myprj/zones/europe-west8-b/instanceGroups/myig-b" },
-        { backend = "projects/myprj/zones/europe-west8-c/instanceGroups/myig-c" },
+        { backend = module.ig-zone-a.group },
+        { backend = module.ig-zone-b.group },
       ]
     }
   }
 }
-# tftest modules=1 resources=5
+# tftest modules=3 resources=9 fixtures=ig
```
wiktorn, line 54: Let's add e2e here, so we can check that it works for both cases

wiktorn (commented):
I really love this one! One comment to consider - in README.md, maybe instead of putting fixtures at the end, let's keep them after examples and before TFDOC, and name the section something like `Definitions of the modules used in the examples` to make it easy discoverable.

ludoo:
Julio, as we discussed I would much prefer to centralize fixtures instead of having a lot of them scattered around modules, which will become

- one more cognitive hurdle for any module changes
- a maintenance nightmare eventually as the provider, resources, and our understanding and use of fixtures evolve

So while this is nice, I am starting to strongly feel that we're focusing too much on the tooling and too little on the actual code. :) So unless we have a way of centralziing fixtures not at the module level, but at the repo level, this is a -1 from me. :)

BTW I also completely share your PoV in a different PR that with e2e, examples are starting to become cryptic. You and I both know **the primary use of examples is to show module usage** not to run e2e tests, and if usability starts becoming impaired we will need to start removing tooling. Can we find a way of reminding this to all our collaborators?

tests/examples/test_plan.py
```diff
@@ -18,19 +18,34 @@
...
     for f in requested_files:
       destination = test_path / example.files[f].path
       destination.parent.mkdir(parents=True, exist_ok=True)
       destination.write_text(example.files[f].content)
 
+  if fixtures is not None:
+    requested_fixtures = fixtures.split(',')
+    for f in requested_fixtures:
+      if f.startswith('fixtures/'):
+        # fixture is specified referencing a global fixture in (tests/fixtures/)
```
wiktorn, line 40: @ludoo Here is a way to define centralized fixtures.

ludoo: Can we default to that, and use tfdoc like Julio and I were discussing to inject the link to the fixtures used (or an expandable box) in the README?

wiktorn:
> BTW I also completely share your PoV in a different PR that with e2e, examples are starting to become cryptic. You and I both know **the primary use of examples is to show module usage** not to run e2e tests, and if usability starts becoming impaired we will need to start removing tooling. Can we find a way of reminding this to all our collaborators?
> 

I totally agree on the purpose of the examples, though I like to extend this definition, not to only show the module usage per se, but how it is used in the context of other modules.

Indeed, maybe now we are leaning too much over developer comfort over usability of examples. I think that by following a naming convention we can achieve both goals:

- Use defined module names in fixtures, that derive from module names in the repository. I.e define `module "net-address"` for `modules/net-address`. If we need more instances of modules, use `-a`, `-b`, etc. suffixes.

The result is, that it is easy to comprehend, which module output goes where on the other's module inputs.

The other thing we can do is to generate links in `README.md` to fixtures used, so it is easy to glue full example. WDYT?

ludoo:
> The other thing we can do is to generate links in `README.md` to fixtures used, so it is easy to glue full example. WDYT?

Yep, I mentioned it above and it seems a great idea and really easy to do via tfdoc. Links or even better expandables, I liked the one you and Julio showed a few days ago.

juliocc:
First of all this is an initial draft to start the discussion before we invest more time in all the e2e. Nothing here is set in stone and we can change (or discard) things as needed.

> Julio, as we discussed I would much prefer to centralize fixtures instead of having a lot of them scattered around modules, which will become

I rushed this PR a bit because of #1913 and didn't fully document the two options I implemented: 
- Inlining the fixture directly in the README file using the `tftest-fixture` directive. This method keeps the code close the example but also becomes a bit unmaintainable IMO
- Dropping a terraform file fixtures in `tests/fixtures` and then reference it from examples. This method is a bit cleaner requires extra tooling to make fixtures easy to discover.

I started with the first method and then brainstorming with @ludoo we came up with the second idea. Both have pros and cons and I'm still not sure which one is better yet.

> the primary use of examples is to show module usage

Agreed 100%. That's why I opened this PR so we can start discussing. My intention was to ping you both on Monday but you both were faster than me ;)

> The other thing we can do is to generate links in README.md to fixtures used, so it is easy to glue full example. WDYT?

@ludoo and I discussed something similar. Let's first decide if we want fixtures at all and then we can figure out how to make them easy to discover.

modules/net-lb-app-ext/README.md
juliocc: done. how can we test it?

wiktorn:
@juliocc I allowed myself to move the test to global file, and change naming of the module.

Other notable changes:
* while preparing test directory, copy symlinks instead of copying of the contents (should reduce the IO during the tests, as we will not copy repo multiple times)
* create a fixtures folder with *symlinks* to root directory and `variables.tf`, to make the development of fixtures easier. My assumption is that here, it won't be a big problem for Windows-based developers

I'd love to get this merged soon, so others can base their work on fixtures.

tests/examples/test_plan.py
juliocc: done. Updated tfdoc to list the used fixtures after the outputs section

wiktorn (approved):

PR #1915, opened 2023-12-10
Added the possibility to create self signed certificates for applicat…

apichick:
Added the possibility to create self signed certificates for applicat…

---
**Checklist**


modules/net-lb-app-ext/self-signed-certs.tf
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+resource "tls_private_key" "private_keys" {
+  for_each  = toset(var.ssl_certificates.self_signed_configs)
+  algorithm = "RSA"
+  rsa_bits  = 4096
```
wiktorn, line 20: We only support [RSA-2048 or ECDSA P-256](https://cloud.google.com/load-balancing/docs/ssl-certificates/self-managed-certs#create-key-and-cert) encryption algorithms.

juliocc (commented):
I'm on the fence with this PR. To me, creating a certificate feels more like an example than something to be handled by the LB modules.

modules/net-lb-app-ext/self-signed-certs.tf
apichick: ah true, let me change that

apichick: done

ludoo:
> I'm on the fence with this PR. To me, creating a certificate feels more like an example than something to be handled by the LB modules.

I also don't like having this in the LB modules. But +1 on showing how to do it in one example.

ludoo:
Miren, I'm closing this let's come up with an example instead. :)

PR #1917, opened 2023-12-11
Added the possibility to configure maintenance window and deny maintenance period in Cloud SQL module module

francesco-pavan-huware:
Added the possibility to configure the following Cloud SQL functionalities in the Cloud SQL module:
- maintenance window
- deny maintenance period

---
**Checklist**


modules/cloudsql-instance/variables.tf
```diff
@@ -90,6 +90,16 @@ variable "deletion_protection_enabled" {
   nullable    = false
 }
 
+variable "deny_maintenance_period" {
```
ludoo, line 93: Can we collapse this and `maintenance_window` into a single `maintenance_config` variable? We always try really hard to keep our variable space small.

francesco-pavan-huware: Done! Thank you

modules/cloudsql-instance/variables.tf
```diff
@@ -143,6 +143,26 @@ variable "labels" {
...
+      hour         = number
+      update_track = optional(string, "stable")
+    }), null)
+    deny_maintenance_period = optional(object({
+      start_date = string
+      end_date   = string
+      start_time = optional(string, "00:00:00")
+    }), null)
+  })
+  default = {
```
ludoo, line 160: you can use use `{}` as a default, then optionals will kick in and set both top-level attributes to `null`

modules/cloudsql-instance/main.tf
```diff
@@ -131,6 +140,15 @@ resource "google_sql_database_instance" "primary" {
         query_plans_per_minute  = var.insights_config.query_plans_per_minute
       }
     }
+
+    dynamic "maintenance_window" {
+      for_each = var.maintenance_config.maintenance_window != null ? [1] : []
```
ludoo, line 145: nit: we use `[""]` instead of `[1]`, don't ask me why though :)

modules/cloudsql-instance/variables.tf
```diff
@@ -143,6 +143,26 @@ variable "labels" {
   default     = null
 }
 
+variable "maintenance_config" {
+  description = "Set maintenance window configuration and maintenance deny period (up to 90 days). Date format: 'yyyy-mm-dd'."
+  type = object({
+    maintenance_window = optional(object({
+      day          = number
+      hour         = number
+      update_track = optional(string, "stable")
```
ludoo, line 152: tqo things here
- if this is a "string as an enum", meaning it can have multiple fixed values, we need a validation condition that ensures only possible values (or `null`) are passed in
- if the provider uses `stable` as default, we leave the module's default to `null` and let the provider infer it

ludoo (changes requested):
Very nice, thanks a lot for being super quick!

Dropped a couple minor comments, then this should be good to go.

ludoo (approved):

modules/cloudsql-instance/variables.tf
```diff
@@ -143,6 +143,36 @@ variable "labels" {
   default     = null
 }
 
+variable "maintenance_config" {
+  description = "Set maintenance window configuration and maintenance deny period (up to 90 days). Date format: 'yyyy-mm-dd'."
+  type = object({
+    maintenance_window = optional(object({
+      day          = number
+      hour         = number
+      update_track = optional(string, null)
```
ludoo, line 152: we still need validation on this for canary/stable/null :)

francesco-pavan-huware: You're right, I missed it sorry. Now it should be fine!

ludoo (approved):
Thanks a lot for being patient with our review comments! :)

PR #1920, opened 2023-12-12
Bump tf version

ludoo:
Check blocks we recently started using require 1.5.0+

### DIFF
--- a/default-versions.tf
+++ b/default-versions.tf
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 terraform {
-  required_version = ">= 1.4.4"
+  required_version = ">= 1.5.0"
   required_providers {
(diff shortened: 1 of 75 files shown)


wiktorn (approved):

wiktorn:
One note, for some reason, our tests are running with `TF_VERSION: 1.5.1`, maybe it's worth aligning both?

ludoo:
> One note, for some reason, [our tests](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/.github/workflows/tests.yml#L28) are running with `TF_VERSION: 1.5.1`, maybe it's worth aligning both?

ach, missed this, can you bump provider again? :)

PR #1932, opened 2023-12-18
Simplify organization tags.tf locals

juliocc:
I was about to add tags to the project module and I noticed some on of the locals in modules/organization/tags.tf could be simplified leveraging the guarantees of optional types.


ludoo (approved):
❤️

PR #1960, opened 2024-01-05
Add PNA support to Service Directory module

stribioli:
Endpoints in Service Directory can be *associated* with a VPC. In this case, they can be used by supported Google Cloud products to send requests directly to resources inside a VPC. This feature is called [Private Network Access](https://cloud.google.com/service-directory/docs/private-network-access-overview).

The `google_service_directory_endpoint` resource supports this configuration with a new argument `network`.
Unfortunately, this argument has an unusual format: it is similar to a standard VPC ID, but instead of the project ID, it expects the project number.

---
**Checklist**

### DIFF
--- a/modules/service-directory/README.md
+++ b/modules/service-directory/README.md
@@ -90,20 +90,59 @@ module "dns-sd" {
 }
 # tftest modules=2 resources=5 inventory=dns.yaml
 ```
-<!-- BEGIN TFDOC -->
 
+### Services with endpoints using private network access
+
+[Private network access](https://cloud.google.com/service-directory/docs/private-network-access-overview) enables supported Google Cloud products to send HTTP requests to resources inside a VPC.
+
+```hcl
+locals {
+  project_number = "123456789012"
+}
+
+module "service-directory" {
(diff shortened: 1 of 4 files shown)


juliocc (approved):

modules/service-directory/README.md
```diff
@@ -90,20 +90,59 @@ module "dns-sd" {
 }
 # tftest modules=2 resources=5 inventory=dns.yaml
 ```
-<!-- BEGIN TFDOC -->
 
+### Services with endpoints using private network access
```
ludoo, line 94: Private Network Access

ludoo (approved):
Thanks!

modules/service-directory/README.md
stribioli: Official branding for _private network access_ is lowercase, unlike (for example) _Private Service Connect_.

ludoo: I would still make it uppercase to flag it's a product feature.

stribioli: Ok!

PR #1969, opened 2024-01-10
Integrate checklist data in FAST

ludoo:
- [x] stage 0
- [x] stage 1
- [x] stage 0 tests
- [x] stage 1 tests
- [x] remove the sample data files from the stage folder


fast/stages/0-bootstrap/checklist.tf
```diff
@@ -0,0 +1,98 @@
...
+  _cl_org_iam = local._cl_org_iam_raw == null ? null : (
+    local._cl_org_iam_raw.version != "0.1.0"
+    ||
+    local._cl_org_iam_raw.organization.id != tostring(var.organization.id)
+    ? null
+    : local._cl_org_iam_raw.iam_bindings
+  )
+  # do a first pass on IAM bindings to identify groups and normalize
+  _cl_org_iam_bindings = local._cl_org_iam == null ? {} : {
+    for b in local._cl_org_iam :
```
wiktorn, line 56:
```suggestion
    for b in local._cl_org_iam :
    # checklist IAM permissions contain either placeholder values as defined in keys of local._cl_groups, or IAM principal
```

fast/stages/0-bootstrap/checklist.tf
```diff
@@ -0,0 +1,98 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  # group mapping from checklist to ours
+  _cl_groups = {
+    BILLING_ADMINS = local.groups.gcp-billing-admins
```
wiktorn, line 20:
```suggestion
    BILLING_ADMINS = local.groups_iam.gcp-billing-admins 
```
Using groups_iam here simplifies `local.checklist.group_iam` creation - you won't need to check if entry is a group

fast/stages/0-bootstrap/checklist.tf
```diff
@@ -0,0 +1,98 @@
...
+        for r in v.additive : [
+          {
+            key    = v.is_group ? "${r}-group:${k}" : "${r}-${k}"
+            member = v.is_group ? "group:${k}" : k
+            role   = r
+          }
+        ]
+      ]
+    ]))
+    location = (
```
wiktorn, line 92:
```suggestion
    logging_location = (
```
Or `location = { logging: (`, to make it more explicit

fast/stages/0-bootstrap/checklist.tf
```diff
@@ -0,0 +1,98 @@
...
+      roles    = b.role
+      is_group = lookup(local._cl_groups, b.group_id, null) != null
+    }
+  }
+  # compile the final data structure we will consume from various places
+  checklist = {
+    billing_account = (
+      local._cl_data == null ? null : local._cl_data.billing_account
+    )
+    group_iam = {
```
wiktorn, line 73: Is there a reason, that we keep `group_iam` separate from `iam` here? IIUC only groups mentioned in `local._cl_groups` will be here. The checklist file may contain the principals mentioning other groups, but we will put them in `iam`. And in the end, those will get merged anyway when we will parse them in organization module

fast/stages/0-bootstrap/main.tf
```diff
@@ -28,6 +28,12 @@ locals {
     for k, v in local.groups :
     k => "group:${v}"
   }
+  locations = local.checklist.location == null ? var.locations : {
+    bq      = var.locations.bq
+    gcs     = var.locations.gcs
+    logging = local.checklist.location
```
wiktorn, line 34: In places we use now `locations.logging` we had earlier `var.locations.logging`. We are now only passing value from checklist. Missing `coalesce()`?

fast/stages/0-bootstrap/organization.tf
```diff
@@ -58,23 +58,52 @@ locals {
...
+  iam = {
+    for k, v in local._iam : k => distinct(concat(
+      v,
+      try(local.checklist.iam[k].authoritative, [])
+    ))
+  }
+  iam_bindings_additive = merge(
+    local._iam_bindings_additive,
+    {
+      for k, v in try(local.checklist.iam_bindings, []) :
```
wiktorn, line 95:
```suggestion
      for k, v in try(local.checklist.iam_bindings, {}) :
```

fast/stages/0-bootstrap/organization.tf
```diff
@@ -58,23 +58,52 @@ locals {
...
+      v.key => v if lookup(local._iam_bindings_additive, v.key, null) == null
```
wiktorn, line 96: This only adds permission, if there is no same key in variables. Given the construction of the keys, this should be exactly the same permission, but why not `disctinct(concat(...))` them here too just in case?

fast/stages/0-bootstrap/data/org-iam.tfvars.json
```diff
@@ -0,0 +1,112 @@
...
+        "roles/owner",
+        "roles/cloudasset.owner",
+        "roles/compute.osAdminLogin",
+        "roles/compute.osLoginExternalUser",
+        "roles/resourcemanager.tagAdmin",
+        "roles/compute.xpnAdmin"
+      ],
+      "resource": {
+        "type": "ORGANIZATION",
+        "id": "656131167402"
```
wiktorn, line 29: I noticed that this org id is not matching the org id on the top. But it is the top id that we verify that is in agreement (and then - I think it is where we do apply the IAM). Is this mistake in data or in logic?

wiktorn (commented):
My attention span was exhausted just as I reached the resman.

Happy to see these changes, hope that comments are helpful

fast/stages/0-bootstrap/checklist.tf
```diff
@@ -0,0 +1,98 @@
...
+    # MONITORING_ADMINS
+    NETWORK_ADMINS  = local.groups.gcp-network-admins
+    ORG_ADMINS      = local.groups.gcp-organization-admins
+    SECURITY_ADMINS = local.groups.gcp-security-admins
+  }
+  # parse raw data from JSON files if they exist
+  _cl_data_raw = (
+    var.factories_config.checklist_data == null
+    ? null
+    : yamldecode(file(pathexpand(var.factories_config.checklist_data)))
```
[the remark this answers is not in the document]
ludoo: we're doing it because I'm dumb and I context switch too much :)

fast/stages/0-bootstrap/checklist.tf
```diff
@@ -0,0 +1,98 @@
...
+  )
+  _cl_org_iam_raw = (
+    var.factories_config.checklist_org_iam == null
+    ? null
+    : yamldecode(file(pathexpand(var.factories_config.checklist_org_iam)))
+  )
+  # check that version and organization id are fine
+  _cl_data = local._cl_data_raw == null ? null : (
```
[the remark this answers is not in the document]
ludoo: this is the only known version so far, plan is to change the format so the json files can be loaded as tfvars for a single top-level variable

fast/stages/0-bootstrap/checklist.tf
```diff
@@ -0,0 +1,98 @@
...
+    local._cl_data_raw.version != "0.1.0"
+    ||
+    local._cl_data_raw.organization.id != tostring(var.organization.id)
+    ? null
```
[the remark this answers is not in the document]
ludoo: good point, this preliminary version was mainly meant to assess feasibility, would you mind sending a patch to this PR with the assertion check?

fast/stages/0-bootstrap/main.tf
ludoo: why would you need coalesce? if the checklist location is null we use the var, if it's not null we use it

ludoo: unless I did my usual hurried PR, and we are using locations.logging instead of local.locations.logging :)

fast/stages/0-bootstrap/data/org-iam.tfvars.json
ludoo: this is my org id for tests, once we move these files to test assets the org id will change :)

fast/stages/0-bootstrap/checklist.tf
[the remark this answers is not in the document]
wiktorn: sure

fast/stages/0-bootstrap/main.tf
wiktorn: We use `local.locations.logging` in other places now. I'd expect that with no checklist, I'd had `var.locations.logging` there.  I think this is the best place to select the one (either `var.locations.logging` or `local.checklist.location`)

fast/stages/0-bootstrap/checklist.tf
ludoo: there will be no other location coming in :)

fast/stages/0-bootstrap/checklist.tf
ludoo: nope, this will mess up the keys for additive bindings...

fast/stages/0-bootstrap/checklist.tf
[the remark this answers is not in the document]
ludoo: I will ignore this as you're writing the check based on our chat

fast/stages/0-bootstrap/checklist.tf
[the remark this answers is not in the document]
ludoo: done

fast/stages/0-bootstrap/main.tf
ludoo: I'm not understanding what you're saying Wiktor: we do the choice of which locations to use here, and we use this local everywhere so what is it you are not liking? I'm resolving this, feel free to reopen and make it clearer

fast/stages/0-bootstrap/organization.tf
ludoo: done

fast/stages/0-bootstrap/main.tf
wiktorn: I mean, if you do now:
`$ grep 'var.locations.logging' *
`
This variable is never used. 

I imagine that here it should state:
```suggestion
    logging = coalesce(local.checklist.location, var.locations.logging)
```

Otherwise we change the behavior if checklist is not provided in logging export configuration.

ludoo: I am still not following :) If checklist is not null logging will always be provided

ludoo: done :)

PR #1978, opened 2024-01-15
Fix identity_type

viliampucik:
Google provider supports empty string for `identity_type`. Even if `IDENTITY_TYPE_UNSPECIFIED` is used as the value, Terraform state stores empty string as the value instead. Hence it make sense to support empty string as a value in the module.

Also `ANY_USER_ACCOUNT` is the correct value for ingress `identity_type`.

---
**Checklist**

### DIFF
--- a/modules/vpc-sc/variables.tf
+++ b/modules/vpc-sc/variables.tf
@@ -93,7 +93,7 @@ variable "egress_policies" {
       v.from.identity_type == null || contains([
         "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
         "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT", ""
-      ], v.from.identity_type)
+      ], coalesce(v.from.identity_type, "-"))
     ])
     error_message = "Invalid `from.identity_type` value in egress policy."
   }
@@ -161,7 +161,7 @@ variable "ingress_policies" {
       v.from.identity_type == null || contains([
         "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
         "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT", ""
-      ], v.from.identity_type)
+      ], coalesce(v.from.identity_type, "-"))
     ])
     error_message = "Invalid `from.identity_type` value in ingress policy."
   }
--- a/modules/vpc-sc/variables.tf
+++ b/modules/vpc-sc/variables.tf
@@ -92,8 +92,8 @@ variable "egress_policies" {
       for k, v in var.egress_policies :
       v.from.identity_type == null || contains([
         "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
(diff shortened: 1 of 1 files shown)


ludoo:
Can you provide a use case? We initially defaulted to empty and the effect was not what users intended, unless an actual value was passed. Unless behaviour changed, I don't think we should default to something that has no actual use.

viliampucik:
@ludoo The issue is not with the default value. The issue is that when you use `IDENTITY_TYPE_UNSPECIFIED` value for `identity_type` in your Terraform code, then after `terraform apply`, Terraform state will store empty string for `identity_type` instead. The new run of `terraform plan/apply` will show a difference between your Terraform code and state, so you will end up with infinite diff. It is better to allow and use empty string instead of `IDENTITY_TYPE_UNSPECIFIED`.

modules/vpc-sc/variables.tf
```diff
@@ -160,8 +160,8 @@ variable "ingress_policies" {
       for k, v in var.ingress_policies :
       v.from.identity_type == null || contains([
         "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
-        "ANY_USER", "ANY_SERVICE_ACCOUNT"
-      ], coalesce(v.from.identity_type, "-"))
+        "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT", ""
+      ], v.from.identity_type)
```
ludoo, line 164: this is what the API has in this enum
image

can you align on those values?

ludoo (commented):
You're removing values which are legitime according to the API docs

https://cloud.google.com/access-context-manager/docs/reference/rest/v1/accessPolicies.servicePerimeters#IdentityType

modules/vpc-sc/variables.tf
jamesdalf: If you populate values in the identities list, you must use "IDENTITY_TYPE_UNSPECIFIED" or "".

When we use "IDENTITY_TYPE_UNSPECIFIED" as the API expects, we encounter the bug. Terraform stores "IDENTITY_TYPE_UNSPECIFIED" as "" as @viliampucik explained. 

This is using provider version 5.11 with terraform 1.6.6. 

The only way for us to avoid an infinite difference was to use "" instead of "IDENTITY_TYPE_UNSPECIFIED" so that our terraform code matches the state. 

Altough the API reference doesn't have "" as a valid value, the provider code shows otherwise: https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/accesscontextmanager/resource_access_context_manager_service_perimeter.go

We tested publishing a module and only changed the validation and it fixes it. 

We either need a fix in the provider or in the validation.

jamesdalf: VPC-SC code is very complicated. After also working with it for some time now, we understand! 

Yes, just adding in support for an empty string will support our use case. 

Thanks for the support!

steenblik: Shall we also raise an issue in the provider repo? Seems like the infinite diff when specifying `IDENTITY_TYPE_UNSPECIFIED` might be a provider bug.

ludoo (approved):

PR #1980, opened 2024-01-16
Proper validation of empty string value in identity_type

viliampucik:
This is continuation of https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1978.

Because the logical operators (`||`, `&&`) in Terraform [do not short-circuit](https://developer.hashicorp.com/terraform/language/expressions/operators#logical-operators), and [coalesce()](https://developer.hashicorp.com/terraform/language/functions/coalesce) never returns non-empty string, empty string validation needs to be done differently.

The provided code change properly validates if `identity_type` is null or empty string or any of the allowed keywords.

---
**Checklist**

### DIFF
--- a/modules/vpc-sc/variables.tf
+++ b/modules/vpc-sc/variables.tf
@@ -90,10 +90,10 @@ variable "egress_policies" {
   validation {
     condition = alltrue([
       for k, v in var.egress_policies :
-      v.from.identity_type == null || contains([
+      v.from.identity_type == null ? true : contains([
         "IDENTITY_TYPE_UNSPECIFIED", "ANY_IDENTITY",
         "ANY_USER_ACCOUNT", "ANY_SERVICE_ACCOUNT", ""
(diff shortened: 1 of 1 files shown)


ludoo (approved):
Thanks for this.

Nit: I did not check but from memory `contains([null, "", ...])` might also work.

viliampucik:
`contains()` would fail if its last argument would be null:

```
contains([null, "", "a", "b"], null)
╷
│
│   on <console-input> line 1:
│   (source code not available)
│
│ Invalid value for "value" parameter: argument must not be null.

ludoo:
> `contains()` would fail if its last argument would be null:
> 
> ```
> contains([null, "", "a", "b"], null)
> ╷
> │ Error: Invalid function argument
> │
> │   on <console-input> line 1:
> │   (source code not available)
> │
> │ Invalid value for "value" parameter: argument must not be null.
> ```

Thanks for correcting me, you're absolutely right :)

PR #1983, opened 2024-01-16
Fixed value in compute-vm module

luigi-bitonti:
- "disk" value requires disk name and not the disk source path: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk_resource_policy_attachment#example-usage---disk-resource-policy-attachment-basic

- "existing" is not a supported source type.

---
**Checklist**


modules/compute-vm/resource-policies.tf
```diff
@@ -142,7 +142,7 @@ resource "google_compute_disk_resource_policy_attachment" "attached" {
   )
   disk = (
     each.value.source_type == "attach"
-    ? each.value.source
+    ? element(split("/", each.value.source), length(split("/", each.value.source)) - 1)
```
ludoo, line 145: maybe `reverse(split("/", each.value.source))[0]`?

luigi-bitonti: Yeah, you've right, it's cleaner. Thanks :)

wiktorn: I want to double check on that, in README.md we have:
> - `source_type = "attach"` can be used for both instances and templates to attach an existing disk, set source to the name (for zonal disks) or self link (for regional disks) of the existing disk to attach; no disk will be created

I think this needs fixing README.md too (and removing regional examples)

The other approach (preferred) is to actually implement attachments for regional disks using `google_compute_region_disk_resource_policy_attachment` and keeping the current semantics of the interface.

@ludoo WDYT?

modules/compute-vm/variables.tf
```diff
@@ -35,7 +35,7 @@ variable "attached_disk_defaults" {
 }
 
 variable "attached_disks" {
-  description = "Additional disks, if options is null defaults will be used in its place. Source type is one of 'image' (zonal disks in vms and template), 'snapshot' (vm), 'existing', and null."
+  description = "Additional disks, if options is null defaults will be used in its place. Source type is one of 'image' (zonal disks in vms and template), 'snapshot' (vm), 'attach', and null."
```
ludoo, line 38: attach is used to flag a disk needs to be created and managed by the module

image

modules/compute-vm/resource-policies.tf
ludoo: attach is used in the module to flag disk creation and management, let's have a discussion because I don't think this PR is doing what you think :)

ludoo:
Closing it for now, please open an architecture design doc for every change involving this module, as it's core to a lot of other stuff we do in Fabric

modules/compute-vm/variables.tf
luigi-bitonti: Yes, but you cannot use "existing" as value, but only one between 'image', 'snapshot', 'attach', 'null'.

ludoo:
Then can you open an issue and provide repro instructions? Then if you propose a change via a PR you link it to the issues, but it gives us the full context on the bug.

PR #1987, opened 2024-01-18
Specify `docker_repository` field for google_cloudfunctions2_function

kumadee:
When docker repository is not specified, by default docker repository `projects/PROJECT_ID/locations/REGION/repositories/gcf-artifacts` is used.
In such a case, terraform plan always generates a difference for `docker_repository` field as the module passes null value but the tfstate file has the above specified default value.
```bash
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.foo.google_cloudfunctions2_function.function will be updated in-place
  ~ resource "google_cloudfunctions2_function" "function" {
        id               = "projects/my-project-e044dd03/locations/europe-west1/functions/foo"
        name             = "foo"
        # (10 unchanged attributes hidden)

      ~ build_config {
          - docker_repository     = "projects/my-project-e044dd03/locations/europe-west1/repositories/gcf-artifacts" -> null
            # (4 unchanged attributes hidden)

            # (1 unchanged block hidden)
        }

        # (2 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```
This fix allows one to prevent unnecessary infrastructure change by setting the default repository id as well as any user created repository.

---
**Checklist**


ludoo (approved):
The permadiff should be fixed in the provider, as optional fields which are not specified should not generate a change once the resource is created.

Adding the variable to be able to use a non-default repository does of course make sense, thanks for this PR.

kumadee:
> The permadiff should be fixed in the provider, as optional fields which are not specified should not generate a change once the resource is created.
> 
> 
> 
> Adding the variable to be able to use a non-default repository does of course make sense, thanks for this PR.

Cool. I can raise the PR to fix the permadiff in the provider next week.

ludoo:
> Cool. I can raise the PR to fix the permadiff in the provider next week.

this would be great, thanks a lot! :)

PR #2002, opened 2024-01-23
Fixes and additional support for ssl_mode for CloudSQL module

spica29:
The CloudSQL module had some minor issues, which are covered in this PR. These are the following:
1. "postgres_client_certificates" variable was renamed to "**client_certificates**" to make it more generic
2. **maintenance_config** validation was failing when validating var.maintenance_config.maintenance_window if the var.maintenance_config is set to null
3. support for **ssl_mode** with validation

---
**Checklist**


modules/cloudsql-instance/main.tf
```diff
@@ -234,8 +234,8 @@ resource "google_sql_user" "users" {
   type     = each.value.type
 }
 
-resource "google_sql_ssl_cert" "postgres_client_certificates" {
-  for_each    = var.postgres_client_certificates != null ? toset(var.postgres_client_certificates) : toset([])
+resource "google_sql_ssl_cert" "client_certificates" {
```
juliocc, line 237: can you add a moved block for this? Otherwise you'll force recreation for existing users

modules/cloudsql-instance/variables.tf
```diff
@@ -204,7 +204,17 @@ variable "network_config" {
   }
 }
 
-variable "postgres_client_certificates" {
+variable "ssl_mode" {
+  description = "(Optional) Specify how SSL connection should be enforced in DB connections. This field provides more SSL enforcment options compared to require_ssl. To change this field, also set the correspoding value in require_ssl. For PostgreSQL instances, the value pairs are listed in the API reference doc for ssl_mode field. For MySQL instances, use the same value pairs as the PostgreSQL instances. For SQL Server instances, set it to ALLOW_UNENCRYPTED_AND_ENCRYPTED when require_ssl=false and ENCRYPTED_ONLY otherwise."
```
juliocc, line 208: Can your trim down this description a little bit? Maybe just one or two sentences and link to the provider field for more details.

Also, remove the "(Optional)" which is already implied by `default=null`

juliocc (dismissed):
thanks for this!

Left a couple of comments below.

modules/cloudsql-instance/variables.tf
```diff
@@ -204,10 +210,15 @@ variable "network_config" {
   }
 }
 
-variable "postgres_client_certificates" {
-  description = "Map of cert keys connect to the application(s) using public IP."
-  type        = list(string)
+variable "ssl_mode" {
```
ludoo, line 213: Can we bundle this into the existing `network_config` variable? Or (maybe better), create a new top-level `ssl` variable and move certificates and SSL configuration there?

spica29: I didnt like how `network_config` contained a lot of variables and validation would be complicated. I can move it into another ssl variable, as suggested, where I will put require_ssl, ssl_mode and certs, maybe that is a better approach

ludoo: Yep, I also agree this is the better option. If you could do it that would be great!

modules/cloudsql-instance/variables.tf
```diff
@@ -245,6 +239,22 @@ variable "root_password" {
...
 }
 
+variable "ssl" {
+  description = "Setting to enable SSL, set config and certificates."
+  type = object({
+    require_ssl = optional(bool, false)
+    # More details @ https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#ssl_mode
+    ssl_mode            = optional(string)
+    client_certificates = optional(list(string))
+
```
ludoo, line 249: nit: can you sort attributes alphabetically, and remove the extra newline?

modules/cloudsql-instance/variables.tf
```diff
@@ -245,6 +239,22 @@ variable "root_password" {
   default     = null
 }
 
+variable "ssl" {
+  description = "Setting to enable SSL, set config and certificates."
+  type = object({
+    require_ssl = optional(bool, false)
```
ludoo, line 245: I would leave default to `null` as the provider then implements its own default (or the API does).

modules/cloudsql-instance/variables.tf
```diff
@@ -245,6 +239,22 @@ variable "root_password" {
...
+    # More details @ https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#ssl_mode
+    ssl_mode            = optional(string)
+    client_certificates = optional(list(string))
+
+  })
+  default = null
```
ludoo, line 251:
```
default = {}
nullable = false
```

then you can remove the `try`s from the validation rules.

modules/cloudsql-instance/main.tf
```diff
@@ -71,7 +71,8 @@ resource "google_sql_database_instance" "primary" {
       ipv4_enabled       = var.network_config.connectivity.public_ipv4
       private_network    = try(var.network_config.connectivity.psa_config.private_network, null)
       allocated_ip_range = try(var.network_config.connectivity.psa_config.allocated_ip_ranges.primary, null)
-      require_ssl        = var.network_config.require_ssl
+      require_ssl        = try(var.ssl.require_ssl, null)
```
ludoo, line 74: see comment in variables on setting default to `{}` and preventing null values for `var.ssl`, `try` can also be removed here

ludoo (changes requested):
Thanks, a couple small nits then it's good to go. :)

modules/cloudsql-instance/variables.tf
ludoo: can you add `nullable = false`?

spica29: it should be there, please check the new version

ludoo: thanks :)

ludoo (approved):

ludoo:
Thanks for your patience!

PR #1991, opened 2024-01-18
Fix named ranges behaviour if cidr_tpl_file variable not provided.

miromichalicka:
Issue: If factories_config.cidr_tpl_file variable wasn't provided, the code failed. 
```
│ Error: Invalid function argument
│ 
│   on .terraform/modules/internal_firewall/modules/net-vpc-firewall/main.tf line 50, in locals:
│   50:     can(var.factories_config.cidr_tpl_file) ? yamldecode(file(var.factories_config.cidr_tpl_file)) : {},
│     ├────────────────
│     │ while calling file(path)
│     │ var.factories_config.cidr_tpl_file is null
│ 
│ Invalid value for "path" parameter: argument must not be null.
```

Solution: Validate if the variable isn't null (variable's default value) and only then proceed to load the file. 

---
**Checklist**

I applicable, I acknowledge that I have:
- [x] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [x] Ran `terraform fmt` on all modified files
- [x] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [x] Made sure all relevant tests pass

PR #2005, opened 2024-01-24
Fix named ranges behaviour if cidr_tpl_file variable not provided.

miromichalicka:
Issue: If factories_config.cidr_tpl_file variable wasn't provided, the code failed. 
```
│ 
│   on .terraform/modules/internal_firewall/modules/net-vpc-firewall/main.tf line 50, in locals:
│   50:     can(var.factories_config.cidr_tpl_file) ? yamldecode(file(var.factories_config.cidr_tpl_file)) : {},
│     ├────────────────
│     │ while calling file(path)
│     │ var.factories_config.cidr_tpl_file is null
│ 
│ Invalid value for "path" parameter: argument must not be null.
```

Solution: Validate if the variable isn't null (variable's default value) and only then proceed to load the file. 

---
**Checklist**


modules/net-vpc-firewall/main.tf
```diff
@@ -47,7 +47,7 @@ locals {
     if contains(["EGRESS", "INGRESS"], r.direction)
   }
   _named_ranges = merge(
-    can(var.factories_config.cidr_tpl_file) ? yamldecode(file(var.factories_config.cidr_tpl_file)) : {},
+    can(var.factories_config.cidr_tpl_file) ? var.factories_config.cidr_tpl_file != null ? yamldecode(file(var.factories_config.cidr_tpl_file)) : {} : {},
```
juliocc, line 50: can you merge the conditions with `&&` instead of adding another ternary?

modules/net-vpc-firewall/README.md
```diff
@@ -249,6 +249,30 @@ healthchecks:
...
+module "firewall" {
+  source     = "./fabric/modules/net-vpc-firewall"
+  project_id = var.project_id
+  network    = var.vpc.name
+  factories_config = {
+    rules_folder = "configs/firewall/rules"
+  }
+  default_rules_config = { disabled = true }
+  named_ranges = {
+    healthchecks : [
```
juliocc, line 265: use `=` instead of `:` for maps

juliocc (changes requested):

modules/net-vpc-firewall/main.tf
miromichalicka: I tried that in #1991 and also ```var.factories_config != null && var.factories_config.cidr_tpl_file != null ? yamldecode(file(var.factories_config.cidr_tpl_file)) : {}``` just now, but in both cases the tests failed. Or do you have something else in mind?

modules/net-vpc-firewall/README.md
miromichalicka: fixed, thanks

modules/net-vpc-firewall/main.tf
miromichalicka: ```The logical operators in Terraform do not short-circuit, meaning var.foo || var.foo.bar will produce an error message if var.foo is null because both var.foo and var.foo.bar are evaluated.``` I think the same applies for `&&` as well.

juliocc: You right. I always forget logical operators don't short circuit.

juliocc (approved):

PR #2013, opened 2024-01-25
Add Tag Template module

lcaggio:
Create Data Catalog - Tag Template module. The module will handle the creation of tag templates and the corresponding factory.


modules/data-catalog-tag-template/main.tf
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+locals {
+  _factory_tag_template = {
+    for f in fileset(var.factory_config_path, "**/*.yaml") :
+    trimsuffix(f, ".yaml") => yamldecode(file("${var.factory_config_path}/${f}"))
+  }
+
+  _tag_template_default = {
```
lcaggio, line 23: is there a better way to handle defaults?

wiktorn: I prefer:
```hcl
  is_required = try(fields.value["is_required"], false)
```
To:
```hcl
  is_required = try(fields.value["is_required"], local._tag_template_default.fields.is_required)
```

As you can see the default value exactly where it is used.

lcaggio: I see, thanks. Updated.

modules/data-catalog-tag-template/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "factory_config_path" {
+  description = "Path to data files and folders that enable factory functionality."
+  type        = string
```
ludoo, line 19: Can you align to this naming and type?

https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/organization/variables.tf#L31

modules/data-catalog-tag-template/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+}
+
+variable "tag_templates" {
+  description = "Tag templates definitions in the form {TAG_TEMPLATE_ID => TEMPLATE_DEFINITION}."
+  type = map(object({
+    display_name = optional(string)
+    force_delete = optional(bool, false)
+    region       = string
+    fields = list(object({
```
ludoo, line 70: can fields be changed/modified after creation? if yes, this should be a map keyed maybe on field_id

modules/data-catalog-tag-template/variables.tf
lcaggio: done

modules/data-catalog-tag-template/variables.tf
lcaggio: updated

ludoo (approved):

PR #1998, opened 2024-01-23
fix bigquery_options dynamic block typo from biquery to bigquery

fawzihmouda:
---
**Checklist**

I applicable, I acknowledge that I have:
- [ ] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [ ] Ran `terraform fmt` on all modified files
- [ ] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [ ] Made sure all relevant tests pass

PR #2015, opened 2024-01-28
Fix typo in logging sinks implementation

ludoo:
This implements the same changes in #1998 which never had the CLA approved.

The partitioned tables boolean is also not optional anymore, so a default value is provided and the block is always defined for BigQuery type destinations.


wiktorn:
I allowed myself to add `bigquery_options` to inventory.

I'm also considering extending `Log sinks` testcase for billing-account to include all types of sinks.

ludoo:
@wiktorn I would merge without waiting for more test-related work, as this fixes an existing bug. Test-related changes can easily go in a later PR.

wiktorn (approved):

PR #2030, opened 2024-01-30
logging for default ingress rules in FAST (CSPR-related)

### LINKED ISSUE #2024: Firewall Logging Default Ingress Deny
**Describe the bug**
When a connection is blocked by the default inbound firewall rule, no logging is in place. 

**Rationale**
The default ingress deny firewall rule does not log blocked packets. [...]

The security best practice is to create a low priority catch-all deny firewall rule for ingress [...] with logging enabled. This allows one to audit what traffic is blocked and allows tools like Security Command Center Event Threat Detection to monitor for threats.

**Recommendation**
1) Create a low priority (65534) deny ingress firewall rule with source 0.0.0.0/0 and enable logging, to log blocked incoming connections.

**Expected Behavior**
Create by default a firewall including logging as described under **Recommendation**.

juliocc:
Fixes #2024

### DIFF
--- a/tests/fast/stages/s2_networking_a_peering/stage.yaml
+++ b/tests/fast/stages/s2_networking_a_peering/stage.yaml
@@ -14,4 +14,4 @@
 
 counts:
   modules: 29
-  resources: 149
+  resources: 151


modules/net-vpc-firewall/main.tf
```diff
@@ -27,7 +27,7 @@ locals {
         for name, rule in ruleset : {
           name                 = name
           deny                 = try(rule.deny, false)
-          rules                = try(rule.rules, [{ protocol = "all" }])
+          rules                = try(rule.rules, [{ protocol = "all", ports = null }])
```
ludoo, line 30: is this needed? I remember an ugly permadiff if we set this but that might not happen anymore... I would test with an actual apply :)

ludoo (approved):
dropped a comment on a potential permadiff, but my assumption needs to be verified :)

modules/net-vpc-firewall/main.tf
juliocc: This is used only if you don't have a `rules` block in the firewall rule. In any other case we use whatever the user specifies. I didn't get any permadiffs when applying.

PR #2032, opened 2024-01-31
Selectively enable logging in FAST and firewall policy module rules

### LINKED ISSUE #2023: Firewall Logging
**Describe the bug**
Firewall logging is not enabled. Firewall logs can help to detect threat actors and to investigate security events. Therefore, it is recommended to have it enabled by default in productive VPCs

**Rational**
Firewall rules or policies are critical for providing network security by defining rules for traffic incoming and outgoing from your network.  Therefore, it is critical to audit and verify that deployed firewall rules or policies are working as intended by auditing what connections are denied and allowed. Firewall logs can be used as a source of network forensics in the case of an incident. Security Command Center Event Threat Detection can use firewall logs to detect threats. Firewall logs are also very helpful in debugging firewall rule and policy issues.

**Recommendation**
It is recommended to implement firewall rule and policy logging especially for: 
1) sensitive networks so that it can be verified that your firewall rules are functioning as expected.
2) the deny ingress and egress rule so that you can audit traffic attempting to enter and leave the network but blocked by the firewall.

Consideration: Firewall rule logs have standard Cloud Logging costs associated with them. Firewall logs can be enabled in production, but not in non-production if cost is more of a concern than network forensics. Firewall logs in non-production can be enabled when needed to debug firewall issues.

**Expected behavior**
Firewall rules should have logging enabled by default. There can be the option to actively opt-in to disable the logs to save costs. This option should have a comment that it is only recommended in non-prod environments.

ludoo:
Fixes #2023, the rationale for this PR is in that issue.

### DIFF
--- a/tests/modules/net_firewall_policy/examples/hierarchical.yaml
+++ b/tests/modules/net_firewall_policy/examples/hierarchical.yaml
@@ -74,7 +74,7 @@ values:
     description: null
     direction: INGRESS
     disabled: false
-    enable_logging: null
+    enable_logging: True
     match:
     - dest_address_groups: null
       dest_fqdns: null
@@ -98,7 +98,7 @@ values:
     description: null
     direction: INGRESS
     disabled: false
-    enable_logging: null
+    enable_logging: True
     match:
     - dest_address_groups: null
       dest_fqdns: null
--- a/fast/stages/2-networking-a-peering/data/hierarchical-ingress-rules.yaml
+++ b/fast/stages/2-networking-a-peering/data/hierarchical-ingress-rules.yaml
@@ -22,6 +22,7 @@ allow-healthchecks:
 
 allow-ssh-from-iap:
   description: Enable SSH from IAP
+  enable_logging: true
   priority: 1002
   match:
     source_ranges:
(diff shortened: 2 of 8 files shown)


tests/modules/net_firewall_policy/examples/hierarchical.yaml
```diff
@@ -74,7 +74,7 @@ values:
     description: null
     direction: INGRESS
     disabled: false
-    enable_logging: null
+    enable_logging: True
```
juliocc, line 77: my OCD is triggered by having both `True` and `true` in the same PR

juliocc (approved):
left one small nit but LGTM!

tests/modules/net_firewall_policy/examples/hierarchical.yaml
ludoo: well, `true` for Terraform, `True` in yaml for inventories :) But maybe I did not get your comment...

juliocc: ignore me ;)

PR #2053, opened 2024-02-06
Added destroy_scheduled_duration variable

luigi-bitonti:
With this PR the possibility of setting the Scheduled for destruction state of a KMS key was introduced.
As suggested [here](https://cloud.google.com/kms/docs/key-states#variable_duration_of_the_scheduled_for_destruction_state), a default duration of 30d has been set.
---
**Checklist**


modules/kms/variables.tf
```diff
@@ -78,6 +78,7 @@ variable "keyring_create" {
 variable "keys" {
   description = "Key names and base attributes. Set attributes to null if not needed."
   type = map(object({
+    destroy_scheduled_duration    = optional(string, "2592000s")
```
juliocc, line 81: can we leave the default as null?

juliocc (changes requested):

juliocc (approved):

PR #2064, opened 2024-02-10
Extend FAST to support different principal types

ludoo:
This is a blocker we have been discussing for a while, that prevents usage of domain-less orgs, or complicates designs where operational groups come from a different Cloud Identity.

The rationale and approach for this PR is documented in [this ADR](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/ludo/iam-changes/fast/docs/0-domainless-iam.md).


ludoo:
Force merging as we might have entered a review loop with Julio doing some changes here. Reviewed offline together.

PR #2071, opened 2024-02-13
Bswenka/psc glb and armor 2 producers

bswenka:
---
**Checklist**


blueprints/networking/psc-glb-and-armor/modules/producer/main.tf
wiktorn, line 1: Can we use here [modules/net-lb-app-ext](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/net-lb-app-ext#private-service-connect-neg-creation) instead of creating that on our own here?

Also, not sure if we still need ILB here, or can we just use Serverless NEG.

blueprints/networking/psc-glb-and-armor/modules/producer/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+
+variable "producer_project_id" {
+  description = "The producer project, in which the LB, PSC Service Attachment and Cloud Run service should be created."
+  type        = string
+}
+
+variable "project_create" {
+  description = "Create project instead of using an existing one."
+  type        = bool
+  default     = false
```
wiktorn, line 25: Using `module/project` with `project_create` false will result in failed  tests. You need a way to set it to `true` from the test in README.md file

blueprints/networking/psc-glb-and-armor/modules/producer/main.tf
wiktorn, line 1: I'm reluctant on creating the module within blueprint. If we can use here existing module, I think we can refer then there, to explain the reader how to map different paths to different backends?

bswenka: Thanks for your feedback. I think the module in this case is intrinsic to the solution we are trying to show here, but open to feedback or other ideas?

blueprints/networking/psc-glb-and-armor/modules/producer/variables.tf
bswenka: Thank you, this has been fixed.

blueprints/networking/psc-glb-and-armor/modules/producer/main.tf
bswenka: We can take a look at this. The recent changes were building on the previous example and are working, but happy to look at refactoring if we can simplify the architecture.

juliocc (approved):

PR #2076, opened 2024-02-14
Use Fabric modules in blueprints/networking/psc-glb-and-armor

wiktorn:
Use Fabric modules in blueprints/networking/psc-glb-and-armor

Few minor improvements along the way:
* remove unnecessary subnets (additional NAT for PSC, VMS)
* remove NOOP compute engine instance
* added url_rewrite for service b, so the service actually works as expected
* added allow-list for consumer projects that can connect to the producers PSC service
* improved `curl` to it uses `terraform output` for find the IP address
* added possibility to meaningfully create projects (using prefixes, parent etc.)

This that are different , though not sure if that's OK:
* added network in consumer project, so PSC attachment doesn't complain on missing `default` network, not sure if that was expected

---
**Checklist**

### DIFF
--- a/blueprints/networking/psc-glb-and-armor/consumer.tf
+++ b/blueprints/networking/psc-glb-and-armor/consumer.tf
@@ -19,8 +19,8 @@ module "consumer-project" {
   name           = var.consumer_project_id
   project_create = var.project_create_config != null
 
-  billing_account = try(var.project_create_config.billing_account)
-  parent          = try(var.project_create_config.parent)
+  billing_account = try(var.project_create_config.billing_account, null)
+  parent          = try(var.project_create_config.parent, null)
   prefix          = var.prefix
   services = [
     "iam.googleapis.com",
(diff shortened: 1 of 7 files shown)


ludoo (approved):
❤️

bswenka:
I appreciate the changes / enhancements, and am a fan of simplification, but these changes broke this blueprint and the proposed ways of testing this with the serverless backend httpbin, this blueprint no longer works as expected, or seems to work at all. 

Can we please revert these changes and be more thoughtful about these updates? The previous code worked as expected and there were no errors.

blueprints/networking/psc-glb-and-armor/consumer.tf
```diff
@@ -14,125 +14,124 @@
...
-  psc_target_service    = module.producer_b_project.psc_ilb_service_attachment.self_link
-
-  network    = "default"
-  subnetwork = "default"
+module "producer-a" {
+  source                = "./modules/producer"
+  consumer_project_id   = module.consumer-project.project_id
+  prefix                = var.prefix
+  producer_project_id   = var.producer_a_project_id
+  project_create_config = var.project_create_config
```
wiktorn, line 36: Missing setting of `region`, so producer resources are crated in different region and it breaks the blueprint.

Issue #1972, opened 2024-01-10
Fail to change network endpoint group definition used by a backendService


josekasna:
**Describe the bug**
Making changes to network endpoint groups created via the `neg_configs` block from module [net-lb-app-ext](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/net-lb-app-ext) fails with `is already being used by`

This is, for example, trying to change a cloud functions name of the pre-existing neg already created via the module.

**Environment**
```
/home/runner/work/_temp/409388a3-c76a-4473-930a-af80260728a4/terraform-bin version
Terraform v1.6.6
on linux_amd64

```

```
output from `git rev-parse --short HEAD`
```

**To Reproduce**
Create network endpoint groups with `neg_configs` block. Change the name of the cloud function from a "functiona" to "FunctionA".

**Expected behavior**
The terraform module would: 

- Detach the network endpoint group from the back-ends they are being used by;
- Re-create the neg with the new name;
- Attach network endpoint group with the new function name to the back-ends;

**Result**
Terraform output and/or error messages
[logs_560.zip](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/files/13895605/logs_560.zip)

```
│ Error: Error when reading or editing RegionNetworkEndpointGroup: googleapi: Error 400: The network_endpoint_group resource 'projects/defiplatformdev/regions/australia-southeast1/networkEndpointGroups/glb-neg-39' is already being used by 'projects/defiplatformdev/global/backendServices/glb-backend-39', resourceInUseByAnotherResource
```

**Additional context**
Add any other context about the problem here

ludoo:
Can you paste your module configuration here so we reproduce your exact use case?

josekasna:
`
locals {
  gclb_create      = var.glb_custom_domain == null ? false : true
}

module "glb" {
  source              = "./modules/net-lb-app-ext"
  count               = local.gclb_create && var.enable_glb ? 1 : 0
  project_id          = var.project_id
  name                = "glb"
  address             = google_compute_global_address.default[0].address
  use_classic_version = false
  https_proxy_config = {
    ssl_policy = google_compute_ssl_policy.main-ssl-policy[0].name
  }

  backend_buckets_config = {
    gcs-0 = {
      bucket_name = "lbdefault-${var.project_id}"
    }
  }

  backend_service_configs = {
    for idx, cf in var.glb_cloud_functions :
[...]

PR #2080, opened 2024-02-15
WIP - Allow changes of the LB configurations

wiktorn:
Work in progress, to discuss approach

This is approach, which could solve #1972 

The idea is to force recreation of the resource when used resources are recreated. Resource is recreated, it's `id` should be known only after apply, hence even if it stays the same, the plan will include replacement of the `terraform_data` trigger resource and then - the resource itself.

The approach should be added to almost all resources within each load-balancer module.

---
**Checklist**

### DIFF
--- a/modules/net-lb-app-int/README.md
+++ b/modules/net-lb-app-int/README.md
@@ -641,12 +641,12 @@ module "ilb-l7" {
 
 | name | description | resources |
 |---|---|---|
-| [backend-service.tf](./backend-service.tf) | Backend service resources. | <code>google_compute_region_backend_service</code> |
+| [backend-service.tf](./backend-service.tf) | Backend service resources. | <code>google_compute_region_backend_service</code> · <code>terraform_data</code> |
 | [groups.tf](./groups.tf) | None | <code>google_compute_instance_group</code> |
 | [health-check.tf](./health-check.tf) | Health check resource. | <code>google_compute_health_check</code> |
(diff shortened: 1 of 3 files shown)


ludoo:
I would avoid adding a data source to all LB modules. Nice workaround, but data sources and lifecycle in modules are a no-no from my PoV.

juliocc:
> I would avoid adding a data source to all LB modules. Nice workaround, but data sources and lifecycle in modules are a no-no from my PoV.

I'm not sure how it works but it's not a data source

ludoo:
> > I would avoid adding a data source to all LB modules. Nice workaround, but data sources and lifecycle in modules are a no-no from my PoV.
> 
> I'm not sure how it works but it's not a data source

which teaches me not to check PRs on my mobile while having coffee at the bar downstairs :)

if the resource does not require an API call, then sure... if it breaks tests then I'd much prefer having testable modules with a known corner case, than the opposite

ludoo:
> The only drawback of this approach I can see, is that it will recreate resources, even if we are just adding new NEG or backend, which might be suboptimal in production use cases, though I see great value when working in "rapid prototyping" mode.

This is a deal breaker.

ludoo:
Can we close this? I don't think a change that force regenerates resources should be the default. It would be a bit like saying we switch to using lists in all variables where we use maps, as we don't much care if elements are recreated on unrelated changes. :)

wiktorn:
I did some further tests and totally agree, that this should not be a default. 

I think we can do a better job explaining how to handle this, and I think we do not have currently any good way to support rotation of self-managed TLS certificates, I'll create issues for that.

ludoo:
Thanks Wiktor, completely agree on the need to document this. Should we open an issue for it?

PR #2087, opened 2024-02-17
add analytics hub module

thinhha:
---
**Checklist**


modules/analytics-hub/variables.tf
```diff
@@ -0,0 +1,103 @@
...
+variable "icon" {
+  description = "Base64 encoded image representing the data exchange."
+  default     = null
+  type        = string
+}
+
+variable "listings" {
+  description = "Listings definitions in the form {LISTING_ID => LISTING_CONFIGS}. LISTING_ID must contain only Unicode letters, numbers (0-9), underscores (_). Should not use characters that require URL-escaping, or characters outside of ASCII, spaces."
+  type = map(object({
+    description      = optional(string)
```
ludoo, line 54: nit: can you put the non-optional fields first, then the optionals, then the object-typed ones; then sort each group alphabetically?

modules/analytics-hub/variables.tf
```diff
@@ -0,0 +1,103 @@
...
+variable "factories_config" {
+  description = "Paths to data files and folders that enable factory functionality."
+  type = object({
+    listings = optional(string)
+  })
+  nullable = false
+  default  = {}
+}
+
+variable "iam" {
```
ludoo, line 38: If you implement IAM, please implement the full interface. Simply copy `variables-iam.tf` and `iam.tf` from any of the modules implementing the IAM interface (e.g. folder or project), then replace the resources names and tweak their attributes.

modules/analytics-hub/iam.tf
```diff
@@ -0,0 +1,50 @@
...
+    for index, listing_iam in local.listing_iam :
+    "${listing_iam.listing_id}-${listing_iam.role}" => listing_iam
+  }
+  project          = google_bigquery_analytics_hub_data_exchange.data_exchange.project
+  location         = google_bigquery_analytics_hub_data_exchange.data_exchange.location
+  data_exchange_id = google_bigquery_analytics_hub_data_exchange.data_exchange.data_exchange_id
+  listing_id       = each.value.listing_id
+  role             = each.value.role
+  members          = each.value.members
+  depends_on       = [google_bigquery_analytics_hub_listing.listing]
```
ludoo, line 49: don't use explicit dependencies, just use the resource in the `listing_id` attribute to create an implicit dependency

ludoo (changes requested):

modules/analytics-hub/variables.tf
thinhha: done

modules/analytics-hub/iam.tf
thinhha: I wasn't sure if this would work at first but found a way to remove the depends_on block here.

ludoo (approved):
Can you also update the top-level repo and modules folder READMEs, so the module is listed there?

thinhha:
Thanks for your review @ludoo. I've raised a separate PR for README in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2094

PR #2100, opened 2024-02-20
Module Data Catalog Tag - Add support for types

lcaggio:
Module Data Catalog Tag - Add support for types
---
**Checklist**


modules/data-catalog-tag/main.tf
```diff
@@ -31,8 +31,13 @@ resource "google_data_catalog_tag" "engine" {
...
     content {
-      field_name   = fields.key
-      string_value = fields.value
+      field_name      = fields.key
+      display_name    = try(fields.value.display_name, null)
+      order           = try(fields.value.order, null)
+      string_value    = try(fields.value.string_value, null)
+      double_value    = try(fields.value.double_value, null)
+      enum_value      = try(fields.value.enum_value, null)
+      timestamp_value = try(fields.value.timestamp_value, null)
```
wiktorn, line 40: As `fields.value` is never null, can we just skip `try(...)` here? All other fields just just resolve to `null`. That would allow to capture, that `order` and `display_name` is not defined in the variable :-)

modules/data-catalog-tag/variables.tf
```diff
@@ -31,7 +31,12 @@ variable "tags" {
     column     = optional(string)
     location   = string
     template   = string
-    fields     = map(string)
+    fields = map(object({
+      double_value    = optional(number)
+      string_value    = optional(string)
+      timestamp_value = optional(string)
+      enum_value      = optional(string)
```
wiktorn, line 38:
```suggestion
      double_value    = optional(number)
      enum_value      = optional(string)
      string_value    = optional(string)
      timestamp_value = optional(string)
```
Sorted

modules/data-catalog-tag/README.md
```diff
@@ -5,6 +5,8 @@ This module allows managing [Data Catalog Tag](https://cloud.google.com/data-cat
 ## TODO
 
 - Add support for entries different than Bigquery resources.
+- Ass support to BOOL when [issue](https://github.com/hashicorp/terraform-provider-google/issues/16856) is fixed.
```
wiktorn, line 8:
```suggestion
- Add support to BOOL when [issue](https://github.com/hashicorp/terraform-provider-google/issues/16856) is fixed.
```

wiktorn (approved):

modules/data-catalog-tag/main.tf
lcaggio: Order should not be there, it was a mistake. We need to keep try() because in the yaml definition the user may put just one of them.

wiktorn: Ah understood, the approach to avoid `try(...)` in this situation is to go through data from yaml and merge each element with map which has all of the objects with null value

modules/data-catalog-tag/variables.tf
```diff
@@ -31,7 +31,12 @@ variable "tags" {
     column     = optional(string)
     location   = string
     template   = string
-    fields     = map(string)
+    fields = map(object({
+      double_value    = optional(number)
```
wiktorn, line 35: Do we add `display_name` here then?

modules/data-catalog-tag/README.md
```diff
@@ -109,15 +128,20 @@ parent: projects/project-data-product/datasets/exposure
 location: europe-west1
 template: projects/project-datagov/locations/europe-west1/tagTemplates/test
 fields:
-  owner_email: example@example.com
+  owner_email:
```
wiktorn, line 131: Can we have a test that also covers display name here?

modules/data-catalog-tag/README.md
```diff
@@ -20,7 +22,18 @@ module "data-catalog-tag" {
       location   = "europe-west1"
       template   = "projects/project-datagov/locations/europe-west1/tagTemplates/demo"
       fields = {
-        source = "DB-1"
+        source = {
```
wiktorn, line 25: Covering `dispaly_name` here would also be helpful, Can help with inventory generation for that test case, as it would be great to have it here.

modules/data-catalog-tag/variables.tf
lcaggio: Display name is an output of the resource. Removed in the main.tf, no need to add it here.

modules/data-catalog-tag/README.md
lcaggio: Display name is an output of the resource. Removed in the main.tf, no need to add it here.

modules/data-catalog-tag/README.md
lcaggio: Display name is an output of the resource. Removed in the main.tf, no need to add it here.

lcaggio (commented):
implementing your comments

wiktorn (approved):

PR #2112, opened 2024-02-27
Add support for billing budgets to project factory

ludoo:
This PR adds support for billing budgets to the project factory.

The approach is consistent with how budgets are designed, where there's no 1:1 relationship between a budget and a project, but budgets are meant to be defined for groups of projects identified by one or more common ancestors (organization or folder), and filtered via a list of specific projects if needed.

This design accommodates the 1:1 scenario of course, but is really meant to support budget definition for groups of projects belonging to a team or service, or even for a whole organization. The naive approach we followed in the previous project factory version which is also from time to time submitted via PRs, of having a forced 1:1 mapping between a project and a billing account, is too narrow and prevents using budgets as they are designed to be used.


modules/project-factory/README.md
```diff
@@ -26,7 +24,49 @@ In addition to the YAML-based project configurations, the factory accepts three
...
-Some examples on where to use each of the three sets are provided below.
+Some examples on where to use each of the three sets are [provided below](#example).
+
+## Additional resources
+
+### Service accounts
+
+Service accounts can be managed as part of each project's YAML configuration. This allows creation of default service accounts used for GCE instances, in firewall rules, or for application-level credentials without resorting to a separate Terraform configuration.
+
+Each service account is represented by one key and a set of optional key/value pairs in the `service_accounts` top-evel YAML map, like in this example:
```
wiktorn, line 35:
```suggestion
Each service account is represented by one key and a set of optional key/value pairs in the `service_accounts` top-level YAML map, like in this example:
```

wiktorn (approved):

PR #2115, opened 2024-02-28
Align resource names in FAST networking stages

ludoo:
This PR aligns resource names in FAST networking stages c/d/e, so that transitioning from one stage to a different one can be done with minimal pain. This is already possible for a/b stages, but not for the other ones.

- [x] c-nva
- [x] d-separate-envs
- [x] e-nva-bgp


fast/stages/2-networking-a-peering/data/subnets/dev/dev-dataplatform-ew1.yaml
```diff
@@ -1,5 +1,6 @@
 # skip boilerplate check
 
+name: dev-dataplatform
```
juliocc, line 3: are you intentionally removing the region suffix?

juliocc (approved):
LGTM. Left one small comment

fast/stages/2-networking-a-peering/data/subnets/dev/dev-dataplatform-ew1.yaml
ludoo: Yes. We have added support in the factory a while ago, and it makes the code a lot more readable. Compare for example the ncc stage, where subnet references were some of the longest and ugliest I've ever seen in FAT/Fabric. It also makes the code more portable.

PR #2122, opened 2024-02-29
Define service attachment interface for lb modules and implement in internal LBs

ludoo:
In keeping with the Fabric module design principles, PSC service attachment definition should be supported as an optional but integral part of our load balancer modules.

This PR defines the variable interface for a service attachment, and implements it in two key modules:

- `net-app-lb-int` with an interface supporting a single forwarding rule
- `net-lb-int` with the same interface but supporting multiple forwarding rules

This is an actual working example of the multi-forwarding rules interface for the passthrough modules, the simpler single forwarding rule interface is identical but with a single object in its type:

```hcl
module "simple-ilb-l4-producer" {
  source     = "../cloud-foundation-fabric/modules/net-lb-int"
  project_id = module.simple-project.project_id
  region     = "europe-west8"
  name       = "ilb-l4-test-producer"
  backends = [{
    group = module.simple-vm-nginx.group.id
  }]
  forwarding_rules_config = {
    port-80 = {
      ports = [80]
    }
    port-8080 = {
      ports = [8080]
    }
  }
  vpc_config = {
    network    = module.simple-vpc.id
    subnetwork = module.simple-vpc.subnet_ids["europe-west8/default"]
  }
  service_attachments = {
    port-80 = {
      nat_subnets          = [module.simple-vpc.subnets_psc["europe-west8/playground-psc-l4-80"].id]
    }
    port-8080 = {
      nat_subnets          = [module.simple-vpc.subnets_psc["europe-west8/playground-psc-l4-8080"].id]
      automatic_connection = true
    }
  }
}
```

This PR also contains a small fix to the `net-lb-int` that allows direct referencing the module-managed group by key, aligning it to the behaviour in more recent lb modules.

It also does some code hygiene by adding missing TOCs to READMEs, and renaming the forwarding rules resource in the passthrough modules to conform it to our usual standard.


modules/net-lb-app-int/README.md
```diff
@@ -507,6 +508,35 @@ module "ilb-l7" {
...
+      }]
+    }
+  }
+  vpc_config = {
+    network    = var.vpc.self_link
+    subnetwork = var.subnet.self_link
+  }
+  service_attachment = {
+    nat_subnets          = [var.subnet_psc_1.self_link]
+    automatic_connection = true
```
wiktorn, line 534:
```suggestion
    automatic_connection = true
    consumer_accept_lists = { (var.project_id) = 10 }  # map of `project_id` => `connection_limit`
```
I think its worth to call it out in the example.

wiktorn (approved):
Tested with `psc-glb-and-armor` blueprint

PR #2125, opened 2024-03-03
Add support for PSC network attachments and interfaces in modules

ludoo:
This PR implements support for PSC network attachments and interfaces. The design moves from Fabric principles, and

- implements support for network attachments in the `net-vpc` module, to allow network owners to directly define them
- implements support for network attachments in the `net-address` module, to allow the use case of creating attachments from Shared VPC service projects
- implements support for network attached interfaces in the `compute-vm` module

At a lower level, the design choice on where and how to implement support (apart from the obvious choice of the vm module), moves from a few simple considerations.

The `net-vpc` variable design uses a separate variable, as network attachments have a 1:1 relationship with subnets (single subnet per attachment), and subnets have a 1:n relationship with attachments (multiple attachments possible per subnet). 
Implementing support directly in the subnets variable would have made it overly verbose; using a separate variable for attachments keeps the variable types simpler, and follows the same design principle used for proxy only and psc subnets, which allows to clearly identify a family of resources in variables at a glance (as opposed to IAM which we consider a resource configuration).

The choice of implementing support for attachment creation in service projects in the `net-address` module moves from the fact that at a high level, a network attachment is not entirely dissimilar from reserving subnet addresses (*"If a network attachment accepts a connection from a Private Service Connect interface, Google Cloud allocates to the interface an internal IP address from a consumer subnet that's specified by the network attachment"* from the GCP docs). Implementing the same in the `net-vpc` module would have resulted in weird configurations, where the project passed to the module did not match the one where the resource needs to be created (host vs service), and subnets would still have to be specified by full URL without being internally referenced by the module who would not be managing them.

The `compute-vm` module support for attached interfaces has been implemented via a dedicated variable, as the type is radically different from regular network interfaces, and would have resulted in a very intricate variable with lots of corner cases to be addressed in validation. This choice also makes it obvious that at least one regular network interface is still required even when attached interfaces are used.

The module changes have been tested with actual apply commands in my own setup.

### DIFF
--- a/modules/compute-vm/README.md
+++ b/modules/compute-vm/README.md
@@ -25,6 +25,7 @@ In both modes, an optional service account can be created and assigned to either
     - [Internal and external IPs](#internal-and-external-ips)
     - [Using Alias IPs](#using-alias-ips)
     - [Using gVNIC](#using-gvnic)
+    - [PSC interfaces](#psc-interfaces)
   - [Metadata](#metadata)
   - [IAM](#iam)
   - [Spot VM](#spot-vm)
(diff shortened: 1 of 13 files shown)


juliocc (approved):
great stuff

PR #2151, opened 2024-03-12
Kms e2e tests

dibaskar-google:
---
**Checklist**

- [ X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)

### DIFF
--- a/modules/kms/README.md
+++ b/modules/kms/README.md
@@ -7,8 +7,8 @@ When using an existing keyring be mindful about applying IAM bindings, as all bi
 <!-- BEGIN TOC -->
 - [Protecting against destroy](#protecting-against-destroy)
 - [Examples](#examples)
-  - [Using an existing keyring](#using-an-existing-keyring)
   - [Keyring creation and crypto key rotation and IAM roles](#keyring-creation-and-crypto-key-rotation-and-iam-roles)
+  - [Using an existing keyring](#using-an-existing-keyring)
   - [Crypto key purpose](#crypto-key-purpose)
   - [Import job](#import-job)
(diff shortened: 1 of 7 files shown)


modules/kms/README.md
```diff
@@ -70,18 +54,34 @@ module "kms" {
...
+### Using an existing keyring
+
+```hcl
+module "kms" {
+  source     = "./fabric/modules/kms"
+  project_id = var.project_id
+  iam = {
+    "roles/cloudkms.admin" = ["group:${var.group_email}"]
+  }
+  keyring        = { location = var.region, name = "test-1" }
```
wiktorn, line 69: In the [setup module](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/tests/examples_e2e/setup_module/main.tf#L123) we are creating a `keyring` but not exporting its name / id as variable.

This test fails, as it assumes existence of "test-1" keyring which will not exists in test environment.

Instead, please modify setup_module to export keyring id as variable ([here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/tests/examples_e2e/setup_module/e2e_tests.tfvars.tftpl) and [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/tests/examples/variables.tf) and then use the variable here.

dibaskar-google: Thanks for the suggestion Wiktor. Exported the existing keyring.

wiktorn (approved):

PR #2175, opened 2024-03-26
feat(gke-cluster-standard): Set optionnal `default_node_pool` configuration

anthonyhaussman:
## Description

Add the possibility to set and initialize a `default_node_pool` to correctly size GKE Control-Planes in a restoration process.

## Explanation

As I understand the reason for disabling the creation of the `default_node_pool` is to take advantage of the `gke-nodepool` module, I want to give the option to add this node_pool with a specific size to size Control-Planes during a Disaster Recovery Plan using the GKE restoration features.

To add more details, in my test when I used the `gke-nodepool` to add a nodePool after the creation of the cluster and launch directly a restore process, this one failed due to the resize of Control-Plane by adding nodes.
This action can take a long time, around 45 minutes.

By setting the initial `default_node_pool` at the creation of the cluster, GKE Control-Planes do not need to resize and improve the restoration process time by a lot.

That's why I would like to add this option.

By default, it keeps the same behavior as before and does not create the `default_node_pool`.
Furthermore, a `validation` block has been added to force to set the `initial_node_count` if `remove_pool` has been disabled.

The option can be simply added by setting:
```
  default_node_pool = {
    remove_pool        = false
    initial_node_count = 12
  }
```

I hope to have been clear enough. :pray: 

---
**Checklist**


ludoo:
Thanks for the thorough explanation, it kind of does make sense to me. @juliocc WDYT?

modules/gke-cluster-standard/variables.tf
```diff
@@ -113,6 +113,23 @@ variable "cluster_autoscaling" {
...
+variable "default_node_pool" {
+  description = "Enable default node-pool."
+  type = object({
+    remove_pool        = optional(bool, true)
+    initial_node_count = optional(number, null)
+  })
+  default = {
+    remove_pool        = true
+    initial_node_count = null
+  }
```
juliocc, line 125: Remove outer default, remove null default (it's implied), make non-nullable.
```hcl
  type = object({
    remove_pool        = optional(bool, true)
    initial_node_count = optional(number)
  })
  default = {}
  nullable = false
```

modules/gke-cluster-standard/variables.tf
```diff
@@ -113,6 +113,23 @@ variable "cluster_autoscaling" {
   }
 }
 
+variable "default_node_pool" {
+  description = "Enable default node-pool."
+  type = object({
+    remove_pool        = optional(bool, true)
+    initial_node_count = optional(number, null)
+  })
+  default = {
```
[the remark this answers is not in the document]
ludoo: Condition is also not needed, `remove_pool` has higher precedence so we can internally ignore the node count if that is true.

juliocc (changes requested):

modules/gke-cluster-standard/variables.tf
anthonyhaussman: Fixed in 835c2f9543ac93505a330bc836a4a590893883ee :pray:

modules/gke-cluster-standard/variables.tf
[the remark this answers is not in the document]
anthonyhaussman: Fixed in 835c2f9543ac93505a330bc836a4a590893883ee :pray:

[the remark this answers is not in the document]
anthonyhaussman: Do you think that the validation block can be deleted ?

ludoo (approved):

modules/gke-cluster-standard/variables.tf
```diff
@@ -113,6 +113,21 @@ variable "cluster_autoscaling" {
...
+  description = "Enable default node-pool."
+  type = object({
+    remove_pool        = optional(bool, true)
+    initial_node_count = optional(number)
+  })
+  default  = {}
+  nullable = false
+
+  validation {
+    condition     = ((var.default_node_pool.remove_pool == true && var.default_node_pool.initial_node_count == null) || (var.default_node_pool.remove_pool == false && var.default_node_pool.initial_node_count != null))
```
juliocc, line 126: I think this is the same a this:
```hcl
    condition = (var.default_node_pool.remove_pool != (var.default_node_pool.initial_node_count == null))
```

juliocc (approved):

modules/gke-cluster-standard/variables.tf
anthonyhaussman: The condition does not seem to work on my test. :confused:

anthonyhaussman: But working with:
```
    condition = (var.default_node_pool.remove_pool != (var.default_node_pool.initial_node_count != null))
```

anthonyhaussman: Fixed in 56c81aea7a66bcbfcb0e8d66d03c9e36a0600ec3 :slightly_smiling_face:

PR #2182, opened 2024-03-28
Fix default nodepool defaults in gke standard module

ludoo:
This fixes the defaults in the default nodepool variable for the GKE standard module introduced in #2175. The default variable value should allow creating a cluster, and needs initial node count set to 1 when default nodepool is the be deleted.

It also changes the variable to adopt a more consistent name vs the rest of Fabric.


wiktorn:
LGTM, [E2E run](https://github.com/GoogleCloudPlatform/fast-e2e/actions/runs/8464840407)

PR #2190, opened 2024-04-02
VPC module - PSA configurable service producer

spica29:
---
**Checklist**


juliocc:
@spica29 the PR looks reasonable but I'm wondering what's the use case for this. I've never seen any other value besides "servicenetworking.googleapis.com".

spica29:
> @spica29 the PR looks reasonable but I'm wondering what's the use case for this. I've never seen any other value besides "servicenetworking.googleapis.com".

Hi, we are trying to set up a PSA for NetApp and it uses another service: "netapp.servicenetworking.goog". [Here](https://cloud.google.com/netapp/volumes/docs/get-started/configure-access/networking#configure_private_services_access) is the documentation. I believe that we often just see GCP as a service producer, but we can also have third parties. There is not a lot of documentation about it though.

modules/net-vpc/variables.tf
```diff
@@ -177,10 +177,11 @@ variable "project_id" {
 variable "psa_config" {
   description = "The Private Service Access configuration for Service Networking."
```
drebes, line 178: You can also remove "for Service Networking".

ludoo (approved):

PR #2202, opened 2024-04-04
added force_destroy to dns module

nika-pr:
force_destroy was missing from the DNS module, so I added it.

---
**Checklist**

- [X] Made sure all relevant tests pass:
   
_Outdated:
When I execute `pytest -k 'modules and dns:' tests/examples`,
I am getting `pytest: error: unrecognized arguments: --dist`.
I am using Python `3.10.12` and pytest `8.1.1` on WSL._


modules/dns/variables.tf
```diff
@@ -20,6 +20,12 @@ variable "description" {
   default     = "Terraform managed."
 }
 
+variable "force_destroy" {
+  description = "Set this to true to delete all records in the zone upon zone destruction."
+  type        = bool
+  default     = false
```
juliocc, line 26: Let's leave the default as null

juliocc (approved):
Thanks for the PR @nika-pr!

Left one small comment below

wiktorn:
> * When I execute `pytest -k 'modules and dns:' tests/examples`,
>   I am getting `pytest: error: unrecognized arguments: --dist`.
>   I am using Python `3.10.12` and pytest `8.1.1` on WSL.

Regarding this error. This could happen if you're missing `pytest-xdist` package installed, but it is in `tests/requirements.txt` for quite a long time.

Does installing `pytest-xdist` manually solve the issue for you?

juliocc:
> but it is in `tests/requirements.txt` for quite a long time

[It's not](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/tools/requirements.txt) ;)

AFAIK you don't need unless you're passing `-n` to pytest

wiktorn:
> > but it is in `tests/requirements.txt` for quite a long time
> 
> [It's not](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/tools/requirements.txt) ;)

[Is it](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blame/master/tests/requirements.txt)? ;-)


> AFAIK you don't need pytest-xdist unless you're passing `-n` to pytest

You needed, because I've added `pytest.ini` entry, which depends on having `pytest-xdist` installed always.

wiktorn:
> > You needed, because I've added `pytest.ini` entry, which depends on having `pytest-xdist` installed always.
> hmmmm do we really need it?
> edit: I don't much like forcing pytest behaviour on users...

We can remove that, but then users running e2e tests needs to add `--dist loadgroup` so tests that can't run in parallel, are executed one after another.

As running E2E is niche as of now, this can be removed, but i prefer stuff to just work by default.

nika-pr:
> Does installing `pytest-xdist` manually solve the issue for you?

Yes it did, alls DNS module tests passed :)

PR #2203, opened 2024-04-06
FAST security stage refactor

ludoo:
This PR implements changes discussed with Julio to the security stage in FAST so that VPC-SC

- supports a single perimeter
- optionally leverages module-level factories
- optionally leverages auto-discovery via CAI


lcaggio (approved):
Great addition and great work!

PR #2210, opened 2024-04-10
Add support for quotas to project module

### LINKED ISSUE #2193: Add support for new quota preference resource to project module and project factory
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_quotas_quota_preference

ludoo:
Fixes #2193


modules/project/README.md
```diff
@@ -797,6 +798,29 @@ includedPermissions:
...
+      quota_id        = "CPUS-per-project-region"
+      contact_email   = "user@example.com"
+      preferred_value = 80
+      dimensions = {
+        region = "europe-west8"
+      }
+    }
+  }
+}
+# tftest modules=1 resources=2
```
wiktorn, line 821:
```suggestion
# tftest modules=1 resources=2 e2e
```
Can we add straight away this as E2E? I'm not sure if the contact email has to come from the org, if yes, then `var.group_email` might be sufficient

ludoo: I would not use e2e as this needs a specific provider configuration we can't use as default.

wiktorn:
[E2E test run](https://github.com/GoogleCloudPlatform/fast-e2e/actions/runs/8633172190) to verify new testcase

ludoo:
Hold this while Wiktor deep dives on the quota project thingy. :)

wiktorn:
E2E tests succeeded (once `cloudquotas.googleapis.com` was enabled in IAC project by FAST stage) so I think we can remove the caveat regarding quota project.

PR #2211, opened 2024-04-10
Add project quotas factory

wiktorn:
Follow up on #2210 - add factory to manage quotas.

Use cases for this factory:
* application teams - to manage quotas within their projects (usually together with `project_create = false`), especially when using GPUs
* to be used in [modules/project-factory](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/project-factory) - to allow quota management during project creation

---
**Checklist**

### DIFF
--- a/modules/project/README.md
+++ b/modules/project/README.md
@@ -23,6 +23,7 @@ This module implements the creation and management of one GCP project including 
 - [Custom Roles](#custom-roles)
   - [Custom Roles Factory](#custom-roles-factory)
 - [Quotas](#quotas)
+- [Quotas factory](#quotas-factory)
 - [Outputs](#outputs)
   - [Managing project related configuration without creating it](#managing-project-related-configuration-without-creating-it)
 - [Files](#files)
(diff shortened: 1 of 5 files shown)


ludoo (commented):
Great stuff, but do you think one factory per project is a valid use case? For the type of usage I've seen, I would think of

- support for multiple projects in a single factory, only used to control quotas
- simple support for quotas in a project factory project-level YAML file, to control specific quotas at project creation

Do you have an actual use case for the factory in this pr?

wiktorn:
Updated the description above. My use-case is the - application teams. They do get a project from central team, but they are free to manage IAM and quotas locally. They have just a bunch of projects (few projects per environment), where they do need to manage the quotas and one IaC service account per project (limited to the project scope).

Multiple-project quota-only factory could be used here as well, but in this case, it would be used just for single project.

ludoo:
> Updated the description above. My use-case is the - application teams. They do get a project from central team, but they are free to manage IAM and quotas locally. They have just a bunch of projects (few projects per environment), where they do need to manage the quotas and one IaC service account per project (limited to the project scope).
> 
> Multiple-project quota-only factory could be used here as well, but in this case, it would be used just for single project.

Makes sense, thanks. What would you think of a slightly different factory, that used one file per region respecting the format you get with `gcloud compute regions describe`? It would allow dumping the yaml from gcloud, deleting a bunch of lines, and reusing it.

wiktorn:
> Makes sense, thanks. What would you think of a slightly different factory, that used one file per region respecting the format you get with `gcloud compute regions describe`? It would allow dumping the yaml from gcloud, deleting a bunch of lines, and reusing it.

Good idea. I still think it is necessary to keep current structure as well, for all the quotas that are not region based or have multiple dimensions, and user wants to specify them all.

This will allow some foot-shooting, when user will specify the same quota twice with different values for the same dimensions. The other approach is to "re-key" them by dimensions and quota_id's.

ludoo (approved):
Well, new format has dimensions and is what we need so Wiktor wins this discussion. :)

PR #2212, opened 2024-04-11
Add support for GCS soft-delete retention period

sruffilli:


modules/gcs/main.tf
```diff
@@ -108,11 +99,27 @@ resource "google_storage_bucket" "bucket" {
...
-    for_each = var.custom_placement_config == null ? [] : [""]
+  dynamic "retention_policy" {
+    for_each = var.retention_policy == null ? [] : [""]
+    content {
+      retention_period = var.retention_policy.retention_period
+      is_locked        = var.retention_policy.is_locked
+    }
+  }
 
+  dynamic "soft_delete_policy" {
```
sruffilli, line 110: This is the actual added block. 
I resorted all other dynamics - apologies for the diff-spam :)

ludoo (changes requested):
Variable is too long. Can we do `soft_delete_policy.retention_period` if we assume more controls will be available later for it, or `soft_delete_retention` if not?

sruffilli:
> Variable is too long. Can we do `soft_delete_policy.retention_period` if we assume more controls will be available later for it, or `soft_delete_retention` if not?

 Had the same gut feeling. I'll go with the latter.

ludoo (approved):

PR #2215, opened 2024-04-15
Add new org policies to FAST

juliocc:
**Enabled by default:**

- iam.serviceAccountKeyExposureResponse: disable any SA keys that are detected as exposed
- storage.secureHttpTransport: disable HTTP access to storage resources
- 

**Commented out:**
- constraints/cloudfunctions.restrictAllowedGenerations: Cloud Function versions


juliocc:
@aumohr FYI

ludoo (approved):

PR #2224, opened 2024-04-18
added missing option for exclusion scope

cmalpe:
Maintenance exclusion scope option is missing in the gke standard cluster terraform resource located in the main.tf file forcing the scope to always take up the default value. This PR adds the missing option.
---
**Checklist**

### DIFF
--- a/modules/gke-cluster-standard/main.tf
+++ b/modules/gke-cluster-standard/main.tf
@@ -338,6 +338,9 @@ resource "google_container_cluster" "cluster" {
         exclusion_name = exclusion.value.name
         start_time     = exclusion.value.start_time
         end_time       = exclusion.value.end_time
+        exclusion_options {
+          scope = exclusion.value.scope
+        }
       }
     }
   }


modules/gke-cluster-standard/main.tf
```diff
@@ -338,6 +338,9 @@ resource "google_container_cluster" "cluster" {
         exclusion_name = exclusion.value.name
         start_time     = exclusion.value.start_time
         end_time       = exclusion.value.end_time
+        exclusion_options {
```
juliocc, line 341: can you make this block dynamic based on wether `exclusion.value.scope` is null?

juliocc (changes requested):

juliocc (changes requested):
Also, can you apply the same change in `gke-cluster-autopilot`?

modules/gke-cluster-standard/main.tf
cmalpe: Hi @juliocc yes will work on this and push the change.

juliocc (approved):

PR #2235, opened 2024-04-24
Update FAST logging

juliocc:
- Simplify syntax of default logging filters
- Add policy denied (i.e. IAM deny) logs to default audit logs sink
- Capture audit transparency logs using log name instead of metadata type
- Add new org-level sink for IAM data access logs
- Enable IAM data access logs in the IAC project


ludoo:
I'm always wary of turning on data access logs at the org level. I'm not sure this should be the default especially if we start targeting AI jumpstarts and mini-LZs...

ludoo (approved):

juliocc:
> I'm always wary of turning on data access logs at the org level. I'm not sure this should be the default especially if we start targeting AI jumpstarts and mini-LZs...

Agreed, and it' not the default :)

Data access logs are only enabled in the automation project and only for IAM. The sink is org-level in case the user wants to activate somewhere else

ludoo (commented):
TL;DR don't do reviews before your morning coffee :) really nice PR

PR #2238, opened 2024-04-25
fix: allow disabling node autoprovisioning

kumadee:
- This fix allows a GKE Standard cluster to be configured with no auto-provisioned node pool, but allow setting autocluster profile for user-provisioned node pools like created via `gke-nodepool` module.

---
**Checklist**


juliocc:
Can't this be achieved by setting `cluster_autoscaling` to null?

kumadee:
> Can't this be achieved by setting `cluster_autoscaling` to null?

Yes, one can disable auto node provisioning by setting `cluster_autoscaling` to null, and that is what I currently do.
But the side effect is that I am unable to change the default value of `cluster_autoscaling.autoscaling_profile` to OPTIMIZE_UTILIZATION via terraform and have to change it externally.

juliocc (approved):

PR #2242, opened 2024-04-27
CloudSQL PSC Endpoints support

wiktorn:
PSC Endpoints for CloudSQL module. Simplifies creation of the PSC endpoint for instance.

Added generation of the inventories for YAML based tests.

Open tasks:
* [ ] verify why `psc_config` is not always set in the plan, some inventories have `psc_config: []`

---
**Checklist**


juliocc:
> decide what to do with network_config.connectivity.psc_allowed_consumer_projects (either remove or merge var.psc_consumers into var.network_config.connectivity)

I'm not so sure I like having the creation of the PSC endpoints in this module. Are we going to add PSC endpoint management to every product that supports it (e.g. GKE, Apigee, etc)?

Personally I'd prefer keeping this module just setting the allowed PSC consumer projects and leaving the creation of the endpoints to the consumer. Having said that, PSC creation is becoming common enough that I can see the use of a `net-psc-endpoint` module that creates the `google_compute_address` and the `google_compute_forwarding_rule`.

tests/collectors.py
```diff
@@ -87,8 +87,12 @@ def __init__(self, name, parent, module, inventory, tf_var_files,
...
 
   def runtest(self):
-    s = plan_validator(self.module, self.inventory, self.parent.path.parent,
-                       self.tf_var_files, self.extra_files)
+    summary = plan_validator(self.module, self.inventory, self.parent.path.parent,
+                             self.tf_var_files, self.extra_files)
+    print(f'\n{self.inventory}')
+    print(yaml.dump({'values': summary.values}))
+    print(yaml.dump({'counts': summary.counts}))
+    print(yaml.dump({'outputs': summary.outputs}))
```
juliocc, line 95: Can we remove these prints? You can use tools/plan_summary.py for the same purpose.

ludoo:
> I'm not so sure I like having the creation of the PSC endpoints in this module. Are we going to add PSC endpoint management to every product that supports it (e.g. GKE, Apigee, etc)?
> 
> Personally I'd prefer keeping this module just setting the allowed PSC consumer projects and leaving the creation of the endpoints to the consumer. Having said that, PSC creation is becoming common enough that I can see the use of a `net-psc-endpoint` module that creates the `google_compute_address` and the `google_compute_forwarding_rule`.

100% agree, endpoints are conceptually part of the `net-` family of modules, as they materialize and LB and consume an address. I would explore either integration into `net-address` or a new `net-xxxx` module for them.

tests/collectors.py
ludoo: Use of that tool is not immediate... can we replace those prints with a single one that spits out the command to use the plan summary tool? It would save me a few brain cycles every time :)

wiktorn: I added exact `python tools/plan_summary.py` command to run to regenerate inventory.

wiktorn:
> > Personally I'd prefer keeping this module just setting the allowed PSC consumer projects and leaving the creation of the endpoints to the consumer. Having said that, PSC creation is becoming common enough that I can see the use of a `net-psc-endpoint` module that creates the `google_compute_address` and the `google_compute_forwarding_rule`.
> 
> 100% agree, endpoints are conceptually part of the `net-` family of modules, as they materialize an LB and consume an address. I would explore either integration into `net-address` or a new `net-xxxx` module for them.
> 
> Plus, this would avoid integrating them everywhere as more more products start offering support, like Julio noted above.

I really liked the idea of adding that to `net-address`, which already has some PSC logic, so I followed this path.

This PR is now a bit split-brained, between E2E tests in `cloudsql-instance` and changes in `net-address`, but I hope that you don't mind that.

modules/net-address/README.md
```diff
@@ -122,6 +122,27 @@ module "addresses" {
 # tftest modules=1 resources=1 inventory=psc.yaml e2e
 ```
 
+To create PSC address targeting a service provide `service_attachment`.
```
ludoo, line 125: nit: *a* PSC address targeting a service provide*r*

modules/net-address/README.md
```diff
@@ -122,6 +122,27 @@ module "addresses" {
...
+module "addresses" {
+  source     = "./fabric/modules/net-address"
+  project_id = var.project_id
+  psc_addresses = {
+    cloudsql-one = {
+      address          = "10.0.16.32"
+      subnet_self_link = var.subnet.self_link
+      region           = var.region
+      service_attachment = {
+        name                        = "cloudsql-one"
```
ludoo, line 136: should we just reuse the address name for the service attachment name, and make `service_attachment` a string?

ludoo (approved):
two nits, great PR

modules/net-address/README.md
wiktorn: Making `service_attachment` a string is problematic, because we have a condition `if psc.service_attachment != null` when creating forwarding rules. 

This results in error:
```
The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will identify the instances of this resource.
```

Because when trying to use this, service_attachment is known only after apply. So I'd keep it as an object, even if we ditch separate name for it.
```
     + cloudsql-one = {
          + address            = "10.0.16.32"
          + description        = "Terraform managed."
          + name               = null
          + network            = null
          + region             = "europe-west4"
          + service_attachment = (known after apply)
          + subnet_self_link   = "https://www.googleapis.com/compute/v1/projects/***/regions/europe-west4/subnetworks/e2e-test-1"

```

ludoo: makes total sense, we do the same in several other modules by adding "useless" keys to work around the same issue, sorry for giving you extra work :)

PR #2262, opened 2024-05-09
Make Simple NVA route IAP traffic through NIC 0

juliocc:


modules/cloud-config-container/simple-nva/cloud-config.yaml
```diff
@@ -51,6 +51,9 @@ write_files:
 %{ endfor ~}
       iptables -t nat -A POSTROUTING -o ${interface.name} -j MASQUERADE
 %{ endif ~}
+%{ if interface.number == 0 ~}
+      ip route add 35.235.240.0/20 via `curl http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/gateway -H "Metadata-Flavor:Google"` dev ${interface.name}
+%{ endif ~}
```
wiktorn, line 56: I guess you can also move this after `%{ endfor ~}`s below, and skip the if, as you want this only for interface 0 which always exists?

wiktorn (approved):

PR #2266, opened 2024-05-12
Make 0-bootstrap IAM bindings merge(..) a deep-merge

wiktorn:
When identity provided as  `bootstrap_user` is used in other IAM permissions, it looses its permissions defined in `local.iam_user_bootstrap_bindings` which breaks bootstrap deployment. 

The merge we had works as long as IAM principals are distinct across the merged maps, if they are provided in multiple ones, we need to concat manually the lists.

I do not like how the code looks, so open to any ideas how to improve that. Also, I don't understand why we do not take any authoritative bindings from `local.iam_principal_bindings`, but I kept this as it was. 

---
**Checklist**

### DIFF
--- a/fast/stages/0-bootstrap/organization.tf
+++ b/fast/stages/0-bootstrap/organization.tf
@@ -18,17 +18,28 @@
 
 locals {
   # reassemble logical bindings into the formats expected by the module
-  _iam_bindings = merge(
-    local.iam_domain_bindings,
-    local.iam_sa_bindings,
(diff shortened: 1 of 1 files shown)


ludoo:
Wiktor this is working as idended: from our point of view there should be no user-level permissions at the org level. We also never saw this fail in 3 years of using FAST.

Can you give a specific use case you're trying to address? I'm a bit wary of hurting legibility to address a problem we've never faced.

PR #2280, opened 2024-05-14
Secret manager e2etests

dibaskar-google:
---
**Checklist**

- [X ] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X ] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)


tests/examples_e2e/setup_module/main.tf
```diff
@@ -15,8 +15,9 @@
 locals {
   prefix = "${var.prefix}-${var.timestamp}${var.suffix}"
   jit_services = [
-    "storage.googleapis.com",  # no permissions granted by default
-    "sqladmin.googleapis.com", # roles/cloudsql.serviceAgent
+    "storage.googleapis.com",       # no permissions granted by default
+    "sqladmin.googleapis.com",      # roles/cloudsql.serviceAgent
+    "secretmanager.googleapis.com", # roles/cloudkms.cryptoKeyEncrypterDecrypter
```
wiktorn, line 20: According to [IAM docs](https://cloud.google.com/iam/docs/service-agents), this service account doesn't have any role granted by default, if true - then change comment to `# no permissions granted by default`

tests/examples_e2e/setup_module/main.tf
```diff
@@ -200,6 +201,11 @@ resource "local_file" "terraform_tfvars" {
       email     = google_service_account.service_account.email
       iam_email = "serviceAccount:${google_service_account.service_account.email}"
     }
+    service_identities = {
+      secret = {
+        email = resource.google_project_service_identity.jit_si["secretmanager.googleapis.com"].email
+      }
+    }
```
wiktorn, line 208: I'm reluctant about adding those here, as we can do that only for JIT-ed service accounts, for all others we need to generate the name by hand. 

So I'd prefer to create a identity address by hand using `var.project_number`

tests/fixtures/secret.tf
wiktorn, line 1: This fixture deserves better name. Maybe `kms-global-key`.

We do have regional key available in `var.kms_key` so no need to create another one.

wiktorn (commented):
[E2E run](https://github.com/GoogleCloudPlatform/fast-e2e/actions/runs/9095008577)

wiktorn:
Hi Dinesh,

Thanks for this work. The E2E run showed few errors of looking like this:
```
module.secret-manager.google_secret_manager_secret.default["test-manual"]: Creating...
module.secret-manager.google_secret_manager_secret.default["test-auto"]: Creating...
module.secret-manager.google_secret_manager_secret.default["test-auto"]: Creation complete after 0s [id=projects/***-171577335114-prj/secrets/test-auto]
 
Error: Error creating Secret: googleapi: Error 400: Location [europe-west4] is provided more than once in replication policy.

  with module.secret-manager.google_secret_manager_secret.default["test-manual"],
  on fabric/modules/secret-manager/main.tf line 38, in resource "google_secret_manager_secret" "default":
  38: resource "google_secret_manager_secret" "default" {

', '
Error: Error creating Secret: googleapi: Error 400: Location [europe-west4] is provided more than once in replication policy.

  with module.secret-manager.google_secret_manager_secret.default["test-manual"],
  on fabric/modules/secret-manager/main.tf line 38, in resource "google_secret_manager_secret" "default":
  38: resource "google_secret_manager_secret" "default" {

')
```
Are you sure we have all the required changes?

dibaskar-google:
Hello @wiktorn,

Thanks for reviewing this one. This should be because the TFTEST_E2E_region env variable might be set to "europe-west4" while running the test. As same region is used for replication, this resulted in error.

tests/fixtures/secret.tf
dibaskar-google: There is a test for replication which forces us to have one more key in additional region.

tests/examples/variables.tf
```diff
@@ -62,6 +62,10 @@ variable "region" {
   default = "europe-west8"
 }
 
+variable "region_1" {
+  default = "europe-west9"
+}
+
```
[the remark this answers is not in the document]
ludoo: What about using the FAST convention?

```hcl
regions = {
  primary     = "europe-west1"
  secondary = "europe-west3"
}
```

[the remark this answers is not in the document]
dibaskar-google: Since its a list, thought there can be morethan two regions as well. Assumed that renaming as primary and secondary might give an impression to users that only 2 regions can be used for replication.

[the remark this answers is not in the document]
dibaskar-google: > Let's call it `region_a` and `region_b`. Also, we need to add those to setup module. (with assumption that `var.region_a == var.region` and `var.region_a != var.region_b`)

Sure, let me try this.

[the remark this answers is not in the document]
dibaskar-google: > What about using the FAST convention?
> 
> ```terraform
> regions = {
>   primary   = "europe-west1"
>   secondary = "europe-west3"
> }
> ```

Thanks Ludo, 

Since its a list, thought there can be morethan two regions as well. Assumed that renaming as primary and secondary might give an impression to users that only 2 regions can be used for replication.

tests/examples_e2e/setup_module/main.tf
```diff
@@ -187,14 +187,16 @@ resource "local_file" "terraform_tfvars" {
...
-    project_id      = google_project.project.project_id
-    project_number  = google_project.project.number
-    region          = var.region
+    kms_key_id       = google_kms_crypto_key.key.id
+    organization_id  = var.organization_id
+    project_id       = google_project.project.project_id
+    project_number   = google_project.project.number
+    region           = var.region
+    region_primary   = var.region
+    region_secondary = var.region_secondary
```
wiktorn, line 199:
```suggestion
   regions = {
       primary = var.region
       secondary = var.region_secondary
   }
```

And adjust examples and tests/examples/variables.tf

dibaskar-google: Thanks for the suggestion Wiktor. Its modified now.

tests/fixtures/secret.tf
dibaskar-google: > This fixture deserves better name. Maybe `kms-global-key`.
> 
> We do have regional key available in `var.kms_key` so no need to create another one.

This is fixed now.

wiktorn (approved):

PR #2284, opened 2024-05-15
Unify VPN and Peering FAST stages

sruffilli:
2-networking-a-peering and 2-networking-b-vpn are basically identical, with the only difference being how spokes are connected to the hub.
This PR unifies the two stages into 2-networking-a-simple, where the spoke connectivity method is controlled by variable spoke_configs, which allows a user to easily toggle between peerings (default) and VPNs.


sruffilli:
I don't particularly like the name 2-networking-a-**simple** but I ruled out everything else that came to my mind, so open to suggestions.

ludoo:
> I don't particularly like the name 2-networking-a-**simple** but I ruled out everything else that came to my mind, so open to suggestions.

It works, and names can always be changed later :)

ludoo (approved):

PR #2287, opened 2024-05-16
FAST: Cleanup/harmonization of Simple and NVA net stages

sruffilli:
This PR aligns `2-networking-a-simple` and `2-networking-c-nva` to 

- harmonize resource names and configurations
- smooth going back and forth from one design to the other

... and is propedeutic to a potential unification of `2-networking-c-nva` and `2-networking-e-nva-bgp`


fast/stages/2-networking-c-nva/test-resources.tf
```diff
@@ -14,242 +14,112 @@
...
-#     termination_action = "STOP"
-#   }
-#   metadata = {
-#     startup-script = <<EOF
-#       apt update
-#       apt install iputils-ping bind9-dnsutils
-#     EOF
-#   }
-# }
+output "ping_commands" {
```
juliocc, line 122: Move to outputs.tf?

juliocc (approved):

fast/stages/2-networking-c-nva/test-resources.tf
sruffilli: I thought about it and I don't mind having this specific output in this file, which makes it easier to delete it altogether - but the same should also apply to its variable which would make me a bit more nervous.

wdyt?

ludoo: I would either place both output and variable in the file, or put them in variables.tf and outputs.tf. Both work from my PoV, but I would be consistent.

ludoo (approved):
👍

fast/stages/2-networking-c-nva/test-resources.tf
sruffilli: `tfdoc.py` has rightfully decided that things should exist where they are supposed to exist (i.e. misplaced outputs and variables are not tracked in `README.md`), so moving I'm moving the output to outputs.tf

ludoo: tfdocs reads any variables*.tf and outputs*tf but it seems a bit overkill to have separate files for test vms. Good point, agree on moving them like J suggested.

PR #2305, opened 2024-05-24
FAST MT: Readme updates and more prefix validation 

sruffilli:
This PR documents the process of deploying FAST on a tenant-factory bootstrapped tenant.

It also fixes changes the validation logic for prefix as follows: 

- 0-bootstrap: 9 chars or less
- 1-resman/1-tenant-factory: 9 chars or less if ran at org-level, else 11
- else 11

It also uniforms across all stages the `variables.tf` and `variables-fast.tf` breakdown.


ludoo (approved):
❤️

juliocc (approved):

Issue #2308, opened 2024-05-27
FAST: Companion cleanup (or removal)


sruffilli:
`fast/COMPANION.md` content is currently obsolete. 
While it can be easily fixed, we might take the chance to discuss whether to eliminate it or not.

My proposal would be to eliminate it, while refactoring the `How to run this stage` chapter to document the bare minimum required to deploy each stage.

ludoo:
Oh noes, I'm trying to get issues down to zero, and a new one pops up! 🤣 

Jokes aside, 👍 on the plan. We should have a hard, olistic look at all our FAST docs.

juliocc:
+1, let's remove it and update the main docs.

PR #2313, opened 2024-05-28
FAST: Remove obsolete assets, fixes/updates factory validations

sruffilli:
- Remove obsolete files
  - [COMPANION.md](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/3d668170e5c7fb1f6e0028e53aa59dfabaa150a3/fast/stages/COMPANION.md) per #2308 
  - [IaC_SA.png](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/3d668170e5c7fb1f6e0028e53aa59dfabaa150a3/fast/stages/IaC_SA.png) not anymore referenced across the board
- Updates factory data schema validation, which was obsolete
- Adds factory data schema validation to the linting workflow


ludoo:
TBH I would not want to have validation going forward: the new approach to factories is to only wrap module interfaces, and if we stick to that (which I absolutely would) schemas are redundant and just something extra to manage.

Plus, yamls are now embedded in doc examples, and tested via plan.

Let's please decide, but from my PoV I would nuke all YAML schemas and validation tools.

juliocc:
> TBH I would not want to have validation going forward: the new approach to factories is to only wrap module interfaces, and if we stick to that (which I absolutely would) schemas are redundant and just something extra to manage.
> 

I think validation is a good thing. Just wrapping the module doesn't catch syntactically correct yaml that happens to use incorrect attributes.

> Plus, yamls are now embedded in doc examples, and tested via plan.

We should actually validate those too :)

ludoo:
> > TBH I would not want to have validation going forward: the new approach to factories is to only wrap module interfaces, and if we stick to that (which I absolutely would) schemas are redundant and just something extra to manage.
> 
> I think validation is a good thing. Just wrapping the module doesn't catch syntactically correct yaml that happens to use incorrect attributes.
> 
> > Plus, yamls are now embedded in doc examples, and tested via plan.
> 
> We should actually validate those too :)

Let's discuss, as this is just extra tooling for no real  purpose. The factory and the module interface are the validation.

sruffilli:
> TBH I would not want to have validation going forward: the new approach to factories is to only wrap module interfaces, and if we stick to that (which I absolutely would) schemas are redundant and just something extra to manage.

I'm on the fence: a potential side effect of avoiding validation is that a typo on an optional field wouldn't be caught by terraform (attribute would be ignored) with potential distructive side effects.

OTOH I don't love obsolete validators (which we had RN) and I think the solution would be to add tests to the documentation examples to ensure PR can't go through with an outdated validator.
Last and mostly cosmetic, I like having validation examples we can use to educate users about CI/CD pipelines.

Happy with whatever the decision is.

sruffilli:
Per our chat, I'm abandoning this PR, and sending a new one which simply removes obsolete assets. 
I'll also open an issue/discussion to continue the conversation re:Validation.

PR #2317, opened 2024-05-28
resource_labels added to the node_config nodepool

fulyagonultas:
Resource labels is needed in node_config block of nodepool module

---
**Checklist**


juliocc:
hi @fulyagonultas, we just had an internal discussion about this and we think the best approach is the following:
- Keep `var.labels` for what GKE resources call `resource_labels`. This is what we do in all other modules (including `gke-cluster-standard`) and we don't want to introduce different semantics here.
- Introduce a new variable for gke/k8s labels (e.g. `k8s_labels`) and use that for what GKE nodes call `labels`.

So in summary, keep var.labels as it is and add a new variable/field for kubernetes labels.

Can you make this changes? Sorry for mixup, but I didn't fully understand what you were trying to do initially.

ludoo (approved):

juliocc (approved):

PR #2341, opened 2024-06-04
Alloydb add support for psc

simonebruzzechesse:


modules/alloydb/variables.tf
```diff
@@ -277,6 +277,15 @@ variable "project_id" {
   type        = string
 }
 
+variable "psc_config" {
+  description = "PSC config for cluster and instance. Only one between network_config and psc_config can be used."
+  type = object({
+    enabled                   = optional(bool, null)
```
wiktorn, line 283: We can remove this variable and either psc_config is null (then it is disabled), or not null - then enabled.

Though we need to make `psc_config` default to null, and then have more dancing around, when checking for `allowed_consumer_projects` as parent may be null.

Also - if in the end you want to keep current structure, it should be `optional(bool, false)`

modules/alloydb/variables.tf
```diff
@@ -277,6 +277,15 @@ variable "project_id" {
   type        = string
 }
 
+variable "psc_config" {
```
wiktorn, line 280: I think we can merge it with `network_config`, especially given the comment: `Network configuration for cluster and instance. Only one between cluster_network_config and cluster_psc_config can be used.`

If the above comment means either `network_config` or `psc_config` should be provided - we can have network_config of following type:
```hcl
type= object({
psa_config = optional(object({ <current network_config>}), null)
psc_config = optional(object({<what we need for psc_config}), null)
})
```

And add validation, that only one of the above is provided.

modules/alloydb/variables.tf
```diff
@@ -245,21 +245,33 @@ variable "name" {
...
-    network                      = string
-    allocated_ip_range           = optional(string, null)
-    authorized_external_networks = optional(list(string), null)
-    enable_public_ip             = optional(bool, false)
+    psa_config = optional(object({
+      network                      = optional(string, null)
+      allocated_ip_range           = optional(string, null)
+      authorized_external_networks = optional(list(string), null)
+      enable_public_ip             = optional(bool, false)
+    }), null)
```
wiktorn, line 255:
```suggestion
    psa_config = optional(object({
      network                      = optional(string)
      allocated_ip_range           = optional(string)
      authorized_external_networks = optional(list(string), [])
      enable_public_ip             = optional(bool, false)
    }))
```
`null` is implied, if skipped in the optional

modules/alloydb/variables.tf
```diff
@@ -245,21 +245,33 @@ variable "name" {
...
   })
   nullable = false
   validation {
     condition = (
-      (try(length(var.network_config.authorized_external_networks), 0) != 0 && var.network_config.enable_public_ip)
-      || try(length(var.network_config.authorized_external_networks), 0) == 0
+      var.network_config.psa_config == null || (
+        (try(var.network_config.psa_config.enable_public_ip, false) &&
+        try(length(var.network_config.psa_config.authorized_external_networks), 0) > 0) ||
+        try(length(var.network_config.authorized_external_networks), 0) == 0
```
wiktorn, line 266:
```suggestion
        try(length(var.network_config.psa_config.authorized_external_networks), 0) == 0
```
This was the intention?

modules/alloydb/variables.tf
```diff
@@ -245,21 +245,33 @@ variable "name" {
...
+      )
```
wiktorn, line 267:
```suggestion
      var.network_config.psa_config == null || (
        (
          try(var.network_config.psa_config.enable_public_ip, false) &&
          try(length(var.network_config.psa_config.authorized_external_networks), 0) > 0
         ) || (
          try(length(var.network_config.psa_config.authorized_external_networks), 0) == 0
         )
      )
```
Reformatted for readability.

modules/alloydb/README.md
```diff
@@ -87,6 +93,23 @@ module "alloydb" {
...
+
+```hcl
+module "alloydb" {
+  source       = "./fabric/modules/alloydb"
+  project_id   = var.project_id
+  cluster_name = "db"
+  location     = var.region
+  name         = "db"
+  network_config = {
+    psc_config = {}
```
wiktorn, line 107:
```suggestion
    psc_config = { allowed_consumer_projects = [var.project_id] }
```
Without `allowed_consumer_projects` nobody can connect to the instance, so let's educate users how they can specify  that.

wiktorn (approved):

PR #2350, opened 2024-06-08
Add network tags outputs and examples to project module

### LINKED ISSUE #2348: Project and Organization module differences in support for Network Tags
The organization module supports creating secure tags for firewall policies at the org level. The project module can create secure tags, however the usage pattern seems to be a tad different.

Additionally while creation of Tags seems to work it is not included in the output making referring to the Tag Keys and Values through the module not possible, as of now.

**Environment**
Module using the Project module to create Secure Tags.

Terraform v1.8.2

**Expected behaviour**
The Project module should (imo) have the same way of using and referring to Secure Tags. Currently while it is not avail, I think we'd have a win if we configure it so that the referring to the tags is in the same style as with the Org module.

**Additional Context**
Discussion on usage [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/discussions/2345)

ludoo:
Fixes #2348


sruffilli (approved):
Not relevant to this PR specifically, but I'd rename "Network tags" to "Firewall tags", given the first has a specific meaning within GCP.

LGTM

PR #2360, opened 2024-06-14
Validate bundle, use pathexpand in cloud function modules

ludoo:


modules/cloud-function-v1/bundle.tf
```diff
@@ -22,7 +22,7 @@ locals {
     )
```
wiktorn, line 22: One question - when using zip files, we add sha256 which I guess triggers Cloud Function redeployment when code is changed. Is the `bundle_config.path` approach with the zip file also updating running CF with the new version?

wiktorn (approved):

modules/cloud-function-v1/bundle.tf
ludoo: if you change the zip name it will reupload. TL;DR if you use a zip, it's on you to manage this.

PR #2363, opened 2024-06-14
Add support for different endpoint types for Cloud NAT

wiktorn:
Open question: should we expose `endpoint_types` as a list (as provider does), if the API in the runtime says that only one value is allowed.

---
**Checklist**


modules/net-cloudnat/variables.tf
```diff
@@ -68,6 +64,25 @@ variable "config_timeouts" {
   nullable = false
 }
 
+variable "endpoint_types" {
+  description = "Specifies the endpoint Types supported by the NAT Gateway. Supported values include: ENDPOINT_TYPE_VM, ENDPOINT_TYPE_SWG, ENDPOINT_TYPE_MANAGED_PROXY_LB"
+  type        = list(string)
+  default     = null
+  validation {
+    condition = (var.endpoint_types == null ? true : setunion([
```
ludoo, line 72: nit: why not use `contains` in half the lines? smart way of doing it though, made me smile :)

ludoo (approved):

modules/net-cloudnat/variables.tf
ludoo: ignore me :)

PR #2372, opened 2024-06-22
Added spanner-instance module

apichick:
---
**Checklist**


modules/spanner-instance/variables.tf
```diff
@@ -0,0 +1,167 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "autoscaling_config" {
+  description = "Autoscaling config."
+  type = object({
+    autoscaling_limits = optional(object({
```
ludoo, line 20: maybe rename to `limits` and `targets`?

modules/spanner-instance/iam.tf
```diff
@@ -0,0 +1,93 @@
+/**
```
ludoo, line 1: can you use the same iam variables and main file as the other modules? kms would be a good example, keeping  everything identical and just replacing resource names and attributes which reference the main resources?

this will allow easier maintenance of the IAM interface across modules

ludoo (changes requested):

modules/spanner-instance/variables.tf
apichick: done

modules/spanner-instance/iam.tf
apichick: done, but kms iam file has issues

apichick:
do not approve yet, I will do a change to you can have have instance_create flag like in the kms module

ludoo:
> do not approve yet, I will do a change to you can have have instance_create flag like in the kms module

I'll add a useless comment and approve so you can resolve and merge once you're ready :)

modules/spanner-instance/README.md
```diff
@@ -0,0 +1,168 @@
+# Spanner instance
+
+This module allows to create a spanner instance with associated spanner instance config and databases in the instance. Additionally it allows creating instance IAM bindings and database IAM bindings.
+
+## Examples
```
ludoo, line 5: just resolve this once you're ready :)

ludoo (approved):

modules/spanner-instance/README.md
apichick: done

PR #2384, opened 2024-06-26
Allow Cloud NAT to only use secondary ranges

### LINKED ISSUE #2383: Add ability to change source_ip_ranges_to_nat to allow for multiple NATs
**Describe the bug**
We have a use case where we need to have two NATs for a GKE cluster: one that routes 90% of the traffic through a well-known IP, and another 10% that needs to route via a dynamic IP. In the process of doing this, we tried to declare the net-cloudnat module twice, one handling the 90% case (and would have `["PRIMARY_IP_RANGES", "LIST_OF_SECONDARY_IP_RANGES"]` for one set of pods), another handling the 10% case (which should have `["LIST_OF_SECONDARY_IP_RANGES"]` for another set of pods.)

Code to be changed here: https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/750cff01cd42564512dccb62f32ac5caeeb35620/modules/net-cloudnat/main.tf#L103-L107

Trying to use the net-cloudnat module as-is will result in this error:

```
Error: Error creating RouterNat: googleapi: Error 400: Invalid value for field 'resource.nats[1].subnetworks[0].sourceIpRangesToNat[0]': 'PRIMARY_IP_RANGE'. Can not specify PRIMARY_IP_RANGE for the Subnetwork since it is already configured for Nat with the option PRIMARY_IP_RANGE., invalid
```

We're currently looking into using https://github.com/terraform-google-modules/terraform-google-cloud-nat as a way to work around this limitation.

juliocc:
Fixes #2383

### DIFF
--- a/modules/net-cloudnat/README.md
+++ b/modules/net-cloudnat/README.md
@@ -51,14 +51,22 @@ module "nat" {
       },
       {
         # primary range only
-        self_link  = "projects/${var.project_id}/regions/${var.region}/subnetworks/net-1"
-        all_ranges = false
+        self_link     = "projects/${var.project_id}/regions/${var.region}/subnetworks/net-1"
+        all_ranges    = false
+        primary_range = true
       },
       {
         # both primary and specified secondary ranges
         self_link        = "projects/${var.project_id}/regions/${var.region}/subnetworks/net-2"
         all_ranges       = false
+        primary_range    = true
         secondary_ranges = ["pods"]
+      },
+      {
+        # only primary range
+        self_link     = "projects/${var.project_id}/regions/${var.region}/subnetworks/net-3"
+        all_ranges    = false
+        primary_range = true
       }
     ]
   }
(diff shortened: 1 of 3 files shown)


modules/net-cloudnat/variables.tf
```diff
@@ -44,11 +44,26 @@ variable "config_source_subnetworks" {
     subnetworks = optional(list(object({
       self_link        = string
       all_ranges       = optional(bool, true)
+      primary_range    = optional(bool)
```
wiktorn, line 47:
```suggestion
      primary_range    = optional(bool, false)
```
And then use bool expressions where checking for `primary_range` above?

modules/net-cloudnat/main.tf
```diff
@@ -100,7 +100,12 @@ resource "google_compute_router_nat" "nat" {
         subnetwork.value.all_ranges == true
         ? ["ALL_IP_RANGES"]
         : concat(
-          ["PRIMARY_IP_RANGE"],
+          (
+            subnetwork.value.primary_range == null
+            ? []
+            : ["PRIMARY_IP_RANGE"]
```
wiktorn, line 106:
```suggestion
            subnetwork.value.primary_range 
            ? ["PRIMARY_IP_RANGE"]
            :  []
```

modules/net-cloudnat/README.md
```diff
@@ -51,14 +51,22 @@ module "nat" {
...
         self_link        = "projects/${var.project_id}/regions/${var.region}/subnetworks/net-2"
         all_ranges       = false
+        primary_range    = true
         secondary_ranges = ["pods"]
+      },
+      {
+        # only primary range
+        self_link     = "projects/${var.project_id}/regions/${var.region}/subnetworks/net-3"
+        all_ranges    = false
+        primary_range = true
```
wiktorn, line 69:
```suggestion
      {
        # secondary range only
        self_link     = "projects/${var.project_id}/regions/${var.region}/subnetworks/net-3"
        all_ranges    = false
        primary_range = false
        secondary_ranges = ["pods"]
```

`primary range only` is already shown in lines 54-56

wiktorn (approved):

PR #2387, opened 2024-06-26
Added certificate-manager module

apichick:
---
**Checklist**


modules/certificate-manager/README.md
```diff
@@ -0,0 +1,263 @@
+# Certificate manager
+
+This module allows you to create a certificate manager map and associated entries, certificate, DNS authorizations and issueance configs.
```
ludoo, line 3: issuance (typo)

I would also mention that maps are optional

modules/certificate-manager/variables.tf
```diff
@@ -0,0 +1,106 @@
...
+      hostname     = optional(string)
+      labels       = optional(map(string), {})
+      matcher      = optional(string)
+      certificates = list(string)
+    })), {})
+  })
+  default = null
+
+  validation {
+    condition     = var.map == null ? true : alltrue([for k, v in var.map.entries : v.hostname == null && v.matcher != null || v.hostname != null && v.matcher == null])
```
juliocc, line 97: Split in a few lines to make more readable. Furthermore, I think your expression can be simplified to `(v.hostname == null) != (v.matcher == null)` (untested),  

something like
```
    condition = alltrue([
     for k, v in try(var.map.entries, {}) : 
       (v.hostname == null) != (v.matcher == null)
     ])
   )
```

juliocc (approved):
Left a couple of small comments. Thanks @apichick!

PR #2398, opened 2024-06-29
FAST top-level folder profiles

ludoo:
This is a work in progress PR that implements a core change in resource management discussed with @juliocc and originating from an insight by @eliamaldini.

The core features implemented here are

- [ ] make project factory a stage 2
- [x] support for "FAST folder profiles" in top-level folders, that allow users to flag arbitrary folders so that FAST makes them ready to support any stage 2 (IAM bindings, automation, optional CI/CD)
- [ ] support for CI/CD in top-level folders

Once the above features are complete, the plan is to drop direct support in code for stage 2s from the resource management stage (the `branch_xxx` files), and replace it with YAML folder definitions that implement the same resources. This will allow users to rename, remove, or add additional stage 2 folders.

Stage 3s will follow the same pattern, it remains to see how to properly implement them. One idea would be to depend them on the project factory, which would remove the need of assigning billing and project creator roles to their service accounts.

### DIFF
--- a/fast/stages/1-resman/billing.tf
+++ b/fast/stages/1-resman/billing.tf
@@ -23,6 +23,11 @@ locals {
       module.branch-network-sa.iam_email,
       module.branch-security-sa.iam_email,
     ],
+    # folder profiles
+    [
+      for k, v in local.folder_profile_iam :
+      module.top-level-sa[k].iam_email
+    ],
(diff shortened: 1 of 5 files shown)


ludoo:
Closing this pending discussions, as there's a lot more than needs to be implemented with this.

PR #2406, opened 2024-07-03
Adding TPU limits for GKE cluster node auto-provisioning (NAP)

aurelienlegrand:
Adding TPU limits for GKE cluster node auto-provisioning (NAP)

---
**Checklist**


modules/gke-cluster-standard/variables.tf
```diff
@@ -85,6 +85,11 @@ variable "cluster_autoscaling" {
       min           = number
       max           = number
     })))
+    tpu_resources = optional(list(object({
+      resource_type = string
+      min           = number
+      max           = number
+    })))
```
wiktorn, line 92: Why do we need to have separate structure for PUs? I think  we can use `gpu_resources` to set those

aurelienlegrand: I was also thinking about reusing the existing gpu_resources block but the name is not explicit and neither is our documentation that you can use the same block to configure TPU limits.

Would it be fine to rename this block to "accelerator_resources" and include in the documentation that this can include GPU or TPU? This would break existing use of the module though. 
Or just to update the README to say that you can configure TPU in the GPU block?

wiktorn: We do not have any autoscaling example. Maybe you can add one, and say there, that `gpu_resources` supports both. Renaming this variable to `accelerator_resources` sound good too.

modules/gke-cluster-standard/README.md
```diff
@@ -305,6 +306,46 @@ module "cluster-1" {
...
+  }
+  cluster_autoscaling = {
+    cpu_limits = {
+      max = 48
+    }
+    mem_limits = {
+      max = 182
+    }
+    # Can be GPUs or TPUs
+    accelerator_resources = [{
```
wiktorn, line 336:
```suggestion
    accelerator_resources = [
    {
```
To match indentation at the end of the block :-)

wiktorn (approved):

PR #2407, opened 2024-07-04
Allow project factory projects to override name

juliocc:
Allow project factory projects to override name from yaml definition

### DIFF
--- a/modules/project-factory/factory-projects.tf
+++ b/modules/project-factory/factory-projects.tf
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 Google LLC
+ * Copyright 2024 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


modules/project-factory/factory-projects.tf
```diff
@@ -39,15 +39,15 @@ locals {
...
+        project = lookup(v, "name", k)
       }
     ]
   ])
   project_budgets = {
     for v in local._project_budgets : v.budget => v.project...
   }
   projects = {
-    for k, v in local._projects : k => merge(v, {
+    for k, v in local._projects : lookup(v, "name", k) => merge(v, {
```
ludoo, line 50: Should the filename or name win as a key?

ludoo (approved):

modules/project-factory/factory-projects.tf
juliocc: if name present in yaml, use that, otherwise filename

ludoo: seems counterintuitive, but you came up with the use case so 👍

PR #2410, opened 2024-07-06
[FAST] Add basic NGFW enterprise stage

LucaPrete:
A few notes:

- This is a separate L3 stage, so we don't duplicate code in networking stages
- This is using network firewall policies in addition to legacy firewall rules deployed in stage 2. In the future, if and when we'll migrate to firewall policies, we'll be able create the policies in stage 2 and add the new rules we need to the same policy in this stage.
- This PR doesn't take into account TLS inspection yet. It's complex and it would add many more pieces. To be added after this is merged, as a second step.

**Checklist**


fast/stages/2-networking-a-simple/data/firewall-policy-rules/dev/ingress.yaml
```diff
@@ -0,0 +1,10 @@
+# skip boilerplate check
+
+ingress-default-deny:
+  description: "Deny and log any unmatched ingress traffic."
+  priority: 2147483643
+  action: "deny"
+  match:
+    source_ranges:
+      - "0.0.0.0/0"
+  enable_logging: true
```
LucaPrete, line 10: This doesn't seem to exactly have the effect of what we were doing in vpc firewall rules. Thoughts?

fast/stages/3-network-security/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "factories_config" {
+  description = "Configuration for network resource factories."
+  type = object({
+    data_dir = optional(string, "data")
```
ludoo, line 20: nit: this should be named after the resource type and point to the folder inside data (so one level deeper)

ludoo: and it looks like dev and prod folders are both needed, so I would do something like

```
type = object({
  policy_rules = optional(object({
    dev = string
```

fast/stages/3-network-security/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+  validation {
+    condition     = var.factories_config.data_dir != null
+    error_message = "Data folder needs to be non-null."
+  }
+}
+
+variable "ngfw_enterprise_config" {
+  description = "NGFW Enterprise configuration."
+  type = object({
+    endpoint_zones   = optional(list(string), ["europe-west1-b", "europe-west1-c", "europe-west1-d"])
```
ludoo, line 33: move the regions to the default, not as optional default

plus, I would use primary/secondary and inherit region mappings from the networking stage output file

fast/stages/3-network-security/main.tf
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# tfdoc:file:description Next-Generation Firewall Enterprise configuration.
+
+locals {
+  enabled_vpcs = {
+    dev-spoke-0  = try(var.vpc_self_links.dev-spoke-0, null) != null
```
ludoo, line 21: maybe also check if you have a factory folder for each?

fast/stages/3-network-security/main.tf
```diff
@@ -0,0 +1,59 @@
...
+  count           = var.ngfw_enterprise_config.quota_project_id == null ? 1 : 0
+  source          = "../../../modules/project"
+  billing_account = var.billing_account.id
+  name            = "net-ngfw-0"
+  parent          = var.folder_ids.networking
+  prefix          = var.prefix
+  services        = ["networksecurity.googleapis.com"]
+}
+
+module "ext-ngfw-quota-project" {
```
ludoo, line 41: this should use the same module as the one used for creation, leveraging `project_create` and nulling billing and parent fields, etc.

juliocc, line 41: Why make this quota project optional?

ludoo (approved):
approving with a couple comments, we can discuss live if you want how/if to implement them (Julio also has comments)

fast/stages/3-network-security/main.tf
juliocc, line 21: is this try needed? Perhaps making `var.vpc_self_links` non-nullable would make things easier?

fast/stages/3-network-security/main.tf
LucaPrete: This is different from the one above. It is only used to enable the networksecurity api in an already existing project, created outside of here. Notice here project_create is false. I agree we should null other fields. I'll do that.

LucaPrete: You need a quota project. The user can choose either to pass an existing project or to make the stage create an ad-hoc project

fast/stages/3-network-security/variables.tf
LucaPrete: This is by design. There are two reasons I don't do this:
* IHAC where endpoints are not activate in all zones/regions automatically for different reasons, so I'd like to keep the two things (where workloads are and where endpoints are) separata
* inheir regions from the networking stage would imply that I create components based on the fact that there are one or two regions, which makes everything terribly complex

fast/stages/3-network-security/main.tf
ludoo: My point exactly. Use a single module invocation, with project_create false or true, and billing/parent null if create is not needed. The module takes care of the rest.

fast/stages/3-network-security/variables.tf
ludoo: My point is different: why do you want to hardwire regions in this stage, and then force users to change them, when that information is already present in the network stage, and you can infer it from there?

fast/stages/3-network-security/variables.tf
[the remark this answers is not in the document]
LucaPrete: Will do

PR #2413, opened 2024-07-08
Add E2E tests for Cloud Functions and fix perma-diff

### LINKED ISSUE #2403: Split build and service environment variables in cloud functions modules
We always took a shortcut in the CF modules with environment variables, when we used a single variable to define both build and service variables. Now this (which was not entirely ok to start with) generates a permadiff with the introduction of [log execution](https://cloud.google.com/functions/docs/release-notes#June_18_2024).

We need to change the variable so it has optional `build` and `service` members, and default `service_config.environment_variables["LOG_EXECUTION_ID"]` to avoid a permadiff.

wiktorn:
Closes #2403 and adds end-to-end tests for CFv1 and CFv2

Incompatible change
* `modules/cloud-function-v2` - environment variables for build are now taken from separate variable - `build_environment_variables`

---
**Checklist**


ludoo (approved):
thanks Wiktor!

Issue #2419, opened 2024-07-10
Peering names exceed limits when combined


Beaverkilla:
**Describe the bug**
When the names of the local and peer network are combined, even excluding the optional prefix the length of the names can exceed the allowed 63 character limit set within GCP. There is no option to override or truncate this when the limit is exceeded and as such some networks will never be able to be configured using the peering module.

I.e.
Local network name = "webcore-net1-atlas-web01-production"
Peering network name = "database-net1-apollo-production"
Prefix = null

Error: "name" ("webcore-net1-atlas-web01-production-database-net1-apollo-production") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$"

**Environment**
```
Terraform v1.8.4
on windows_amd64
+ provider registry.terraform.io/hashicorp/google v5.33.0
+ provider registry.terraform.io/hashicorp/google-beta v5.33.0
+ provider registry.terraform.io/hashicorp/random v3.6.2
```

```
6c06adb9
```

**To Reproduce**
1. Create two networks with names exceeding 32 characters in length.
2. Attempt to peer them via the net-vpc-peering module
3. Witness error with no recourse

**Expected behavior**
The peering should be accepted with either a truncated name or the terraform should allow for an override name less than the 64 character limit.

**Result**
```
╷
│ Error: "name" ("webcore-net1-atlas-web01-production-database-net1-apollo-production") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$"
│
│   with module.peering["webcore-net1-atlas-web01-production"].google_compute_network_peering.local_network_peering,
│   on ..\cloud-foundation-fabric\modules\net-vpc-peering\main.tf line 24, in resource "google_compute_network_peering" "local_network_peering":
│   24:   name         = "${local.prefix}${local.local_network_name}-${local.peer_network_name}"
│
╵
```

PR #2420, opened 2024-07-11
Add name overrides for Internal and External Load Balancers

cvanwijck-hub24:
When developing IaC for existing infrastructure and attempting to achieve parity between the two, the current naming assignments of multiple modules in CFF causes issues.

For example in `net-lb-ext` the name of the load balancer added additional items to the name if a `forwarding_rules_config` was passed. 
```terraform
name = (
  each.key == "" ? var.name : "${var.name}-${each.key}"
)
```
This can cause issues where a specific name is preferred, such as when bringing existing infrastructure under the control of IaC. Changing the name of a load balancer, backend service or health check would cause them to be replaced by Terraform. 

This is also relevant to #2419 where peerings have a name set by combining the local and peering network names with a prefix. A similar `override_name` variable would resolve this issue.

Adding an optional `override_name` variable allows the user to specify a name if the generated name is not helpful for their use case. For users who would like to use the module as before and have names generated as before, leaving the `override_name` as `null` will use the default name assignment.

---
**Checklist**

I acknowledge that I have:


modules/net-lb-ext/main.tf
```diff
@@ -34,7 +34,7 @@ resource "google_compute_forwarding_rule" "default" {
   provider = google-beta
   project  = var.project_id
   region   = var.region
-  name = (
+  name = var.override_name != null ? var.override_name : (
```
ludoo, line 37: I'm not following the logic:

- if you have multiple forwarding rules, the override will work for one and fail creation of the others?
- if you have a single one with an empty key, the plain name will be used which is probably what you want?

I think you need to do the same for forwarding rules as you did for health checks, allowing the override for each rule.

ludoo:
It's probably easier if you run checks locally without waiting for the workflow to kick in. Details are in the contributing guide in the root of the repo. :)

modules/net-lb-ext/main.tf
cvanwijck-hub24: See changes. Allowed the override to be used for each rule. Override would now not fail where there are multiple forwarding rules in the list

modules/net-lb-int/variables.tf
```diff
@@ -186,6 +188,12 @@ variable "name" {
   type        = string
 }
 
+variable "override_name" {
+  description = "Optional override name to specify exact name without changes."
+  type        = string
+  default     = null
+}
+
```
wiktorn, line 196: If I read correctly, this is no longer used, is it?

wiktorn (commented):
I think that to be fully comprehensive, this requires allowing additional overrides in `net-lb-ext` for names of:
* google_compute_region_backend_service
* google_compute_service_attachment

modules/net-lb-int/variables.tf
cvanwijck-hub24: Correct, moved into each individual forwarding rule now.

cvanwijck-hub24:
> I think that to be fully comprehensive, this requires allowing additional overrides in `net-lb-ext` for names of:
> 
>     * google_compute_region_backend_service
> 
>     * google_compute_service_attachment

If "fully comprehensive" is the goal, this could be expanded to adding override names across multiple other modules. This would also partially resolve #2419. We have encountered this across many areas of our deployment in taking existing click-ops configured infrastructure under control of Terraform but trying to use CFF to accelerate development. Overrides would be useful for many brownfield deployments.

modules/net-vpc-peering/variables.tf
```diff
@@ -19,6 +19,12 @@ variable "local_network" {
   type        = string
 }
 
+variable "override_name" {
```
juliocc, line 22: This module handles up to two peerings, so I'd make this variable a object with two keys (`local` and `peer`, just like `routes_config`) with `optional(string)` values.

Also, I'd simply call it `name`

modules/net-vpc-peering/main.tf
```diff
@@ -21,7 +21,7 @@ locals {
 }
 
 resource "google_compute_network_peering" "local_network_peering" {
-  name         = "${local.prefix}${local.local_network_name}-${local.peer_network_name}"
+  name         = var.override_name != null ? var.override_name : "${local.prefix}${local.local_network_name}-${local.peer_network_name}"
```
juliocc, line 24: this will end up being something like
```terraform
  name = coalesce(
    var.name.local,
    "${local.prefix}${local.local_network_name}-${local.peer_network_name}"
  )
```
As a plus, you could add a [precondition](https://developer.hashicorp.com/terraform/language/expressions/custom-conditions#preconditions-and-postconditions) to ensure the user provides a name if the default is longer than the maximum allowed lenght.

modules/net-vpc-peering/main.tf
juliocc, line 1: Please update `peer_network_peering` with the same changes as `local_network_peering`

juliocc (changes requested):

juliocc:
This PR is getting messy. I extracted the changes to the peering module to #2459.

juliocc:
@cvanwijck-hub24, just checking in to see if you plan on finishing this PR. If not, we'll go ahead and close it

coopervw:
> @cvanwijck-hub24, just checking in to see if you plan on finishing this PR. If not, we'll go ahead and close it

Yes I will resolve the outstanding issues this week if possible.

cvanwijck-hub24 (commented):
All use coalesce and changed `override_name` to use just `name` for clarity

modules/net-lb-ext/README.md
```diff
@@ -234,6 +233,6 @@ For deploying changes to load balancer configuration please refer to [net-lb-app
 
 ## Fixtures
 
-- [compute-mig.tf](../../tests/fixtures/compute-mig.tf)
-- [compute-vm-group-bc.tf](../../tests/fixtures/compute-vm-group-bc.tf)
+- [compute-mig.tf](..\..\tests\fixtures\compute-mig.tf)
+- [compute-vm-group-bc.tf](..\..\tests\fixtures\compute-vm-group-bc.tf)
```
juliocc, line 237: please revert these changes

modules/net-lb-ext/variables.tf
```diff
@@ -30,6 +30,7 @@ variable "backend_service_config" {
     }))
     locality_lb_policy = optional(string)
     log_sample_rate    = optional(number)
+    name               = optional(string, null)
```
juliocc, line 33: the null is not needed

modules/net-vpn-ha/variables.tf
```diff
@@ -63,8 +64,9 @@ variable "router_config" {
       all_subnets = bool
       ip_ranges   = map(string)
     }))
-    keepalive = optional(number)
-    name      = optional(string)
+    keepalive     = optional(number)
+    name          = optional(string)
+    override_name = optional(string, null)
```
juliocc, line 69: still using override name in this file

juliocc (changes requested):

modules/net-vpn-ha/variables.tf
jacklever-hub24: Hey mate, sorry I don't understand what you want me to do with this, it's been a few weeks since we've worked on this. Can you give me a quick heads up and I'll get this fixed please?

juliocc (approved):
lgtm now. Thanks for your patience @cvanwijck-hub24

PR #2421, opened 2024-07-11
fix cloudbuild service account email

nathou:
The correct format for this service account is documented in https://cloud.google.com/build/docs/cloud-build-service-account

---
**Checklist**

### DIFF
--- a/modules/project/service-agents.yaml
+++ b/modules/project/service-agents.yaml
@@ -93,7 +93,7 @@
   service_agent: "service-%s@gcp-sa-cloudbuild.iam.gserviceaccount.com"
   jit: true # roles/cloudbuild.builds.builder
 - name: "cloudbuild-builder"
-  service_agent: "%s@cloudbuild.gserviceaccount.com.iam.gserviceaccount.com"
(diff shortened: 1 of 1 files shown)


wiktorn:
Thank you for reporting and fixing!

ludoo (commented):
The cloud build service account only exists for legacy projects.

https://cloud.google.com/build/docs/cloud-build-service-account-updates

Can you point to the documentation that explains your change?

ludoo:
@wiktorn are you sure this is ok?

juliocc:
This is correct. Cloud Build has two service agents according to the [docs](https://cloud.google.com/iam/docs/service-agents):
image

Merging this, but it will be deprecated by my service agents PR ;)

nathou:
Yes, I know this is a legacy service agent, but there are still existing projects using it. Thanks!

ludoo:
> This is correct. Cloud Build has two service agents
> Merging this, but it will be deprecated by my service agents PR ;)

ok, perfect, I was going by the "we don't have a cb service agent anymore", which is true but only for new projects using default org policies

PR #2430, opened 2024-07-18
FAST: IAM cleanups to reflect PF changes

sruffilli:
This PR retrospect-fixes a few IAM leftovers on the FAST networking stages.

In particular: 

- the `service_project_network_admin` custom role is assigned to the PF SA by 1-resman at folder level, so the project-level assignment done by 2-net is not needed
- `roles/dns.admin` is not required anymore since the PF doesn't deal with DNS management anymore (even though it could arguably be configured to assign DNS roles to other principals)


ludoo:
Good points, I also tackle them in the resman-ng branch, if you're not in a hurry can we discuss this holistically next Mon? :)

ludoo (approved):

PR #2440, opened 2024-07-23
FAST ng: stage 0 environments and VPC-SC IaC resources

ludoo:
This is the first PR adding supporting resources for a different resource management stage.

This only changes stage 0 with no actual impact on existing installations, by

- adding an `environments` variable and using the default environment name defined there instead of the static `prod` name
- adding the environments variable to globals outputs for consumption by the following stages (implemented in later PRs)
- adding service account, bucket and CI/CD resources for the dedicated VPC SC stage (implemented in a later PR)


fast/stages/0-bootstrap/variables.tf
```diff
@@ -85,6 +91,30 @@ variable "custom_roles" {
...
+variable "environments" {
+  description = "Environment names."
+  type = map(object({
+    name       = string
+    is_default = optional(bool, false)
+  }))
+  nullable = false
+  default = {
+    dev = {
+      name = "Development"
```
juliocc, line 103: - Why the nested map? are you expecting more attributes in the future?
- Can you have more than one default? Would it make more sense to have the default outside? Something like this:
  ```terraform
  type = object({
    names = map(string)
    default = string
  })
  ```
  This also removes the need for a check

juliocc (approved):

fast/stages/0-bootstrap/variables.tf
ludoo: but then you need to be consistent, and I would still need to check that default is in names...

TBH I prefer the idea of an environment as "an object", like we do for everything else in modules

PR #2460, opened 2024-08-01
VPC-SC as separate FAST stage 1

ludoo:
This PR removes VPC-SC functionality from the 2-level Security stage, and moves it to a new 1-level dedicated stage that only does VPC-SC.

This allows decoupling VPC-SC from KMS and other resources we might add in the future, and allows applying VPC-SC right after bootstrap in order to immediately switch to enforced mode protection.

Our default perimeter design also changed to a single perimeter, to more closely mimick what we usually do in recent times, where we privilege coarser granularity and immediate enforcement for new organizations, vs finer granularity and a longer delay before enforcement (or no enforcement ever) due to the additional complexity and operational burden.


fast/stages/1-vpc-sc/variables.tf
```diff
@@ -0,0 +1,174 @@
...
+  type        = string
+  default     = null
+}
+
+variable "perimeters" {
+  description = "Perimeter definitions."
+  type = object({
+    default = optional(object({
+      access_levels    = optional(list(string), [])
+      dry_run          = optional(bool, false)
```
juliocc, line 148: This, together with the default access level, can lock you out if you don't update the default regions.

fast/stages/1-vpc-sc/variables.tf
```diff
@@ -0,0 +1,174 @@
...
+variable "outputs_location" {
+  description = "Path where providers, tfvars files, and lists for the following stages are written. Leave empty to disable."
+  type        = string
+  default     = null
+}
+
+variable "perimeters" {
+  description = "Perimeter definitions."
+  type = object({
+    default = optional(object({
```
juliocc, line 146: nit `type=optional(map(object(...), {default={}})`

fast/stages/1-vpc-sc/variables.tf
```diff
@@ -0,0 +1,174 @@
...
+      resources        = optional(list(string), [])
+      vpc_accessible_services = optional(object({
+        allowed_services   = list(string)
+        enable_restriction = optional(bool, true)
+      }))
+    }))
+  })
+  nullable = false
+  default = {
+    access_levels = ["geo"]
```
juliocc, line 160: check this. I think you're missing the default key inside the default value

fast/stages/1-vpc-sc/README.md
```diff
@@ -0,0 +1,308 @@
+# VPC Service Controls
+
+This stage sets up VPC Service Controls (VPC-SC) for the whole organization and is a thing FAST-compliant wrapper on the [VPC-SC module](../../../modules/vpc-sc/), with some minimal defaults.
+
+<p align="center">
+  <img src="diagram.png" alt="Security diagram">
```
juliocc, line 6: remove projects from diagram

juliocc (approved):

fast/stages/1-vpc-sc/variables.tf
ludoo: set default to `dry_run = true`

fast/stages/1-vpc-sc/variables.tf
ludoo: done

fast/stages/1-vpc-sc/variables.tf
ludoo: done

fast/stages/1-vpc-sc/README.md
ludoo: removed diagram as I found no way of doing a good representation of VPC-SC resources, we can of course revisit later

PR #2477, opened 2024-08-07
VPC-SC factory JSON Schemas

ludoo:
As discussed, this only covers the VPC-SC module and does not yet integrate the schemas in FAST.

### DIFF
--- a/modules/vpc-sc/schemas/ingress-policy.schema.json
+++ b/modules/vpc-sc/schemas/ingress-policy.schema.json
@@ -1,0 +1,80 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "VPC-SC ingress policy",
+  "type": "object",
+  "required": [
+    "from",
+    "to"
+  ],
+  "properties": {
+    "from": {
+      "type": "object",
+      "items": {
+        "access_levels": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "identity_type": {
+          "enum": [
+            "IDENTITY_TYPE_UNSPECIFIED",
+            "ANY_IDENTITY",
+            "ANY_USER_ACCOUNT",
+            "ANY_SERVICE_ACCOUNT",
+            ""
+          ]
+        },
+        "identities": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "resources": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "to": {
+      "type": "object",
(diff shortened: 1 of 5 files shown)


modules/vpc-sc/schemas/ingress-policy.schema.json
```diff
@@ -0,0 +1,80 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "VPC-SC ingress policy",
+  "type": "object",
+  "required": [
+    "from",
+    "to"
+  ],
+  "properties": {
+    "from": {
+      "type": "object",
```
wiktorn, line 11:
```suggestion
      "type": "object",
      "additionalProperties": false,
      "required": [
        "access_levels"
      ],
```
Additional properties to catch if someone misspells the attribute.

Required - I vaguely remember that not providing access level bite me some time ago with not working policy. Encoding that in schema will be of a great help.

wiktorn (commented):
Generally - :heart: :heart: :heart: - as this may help me a lot in the future and I was bitten by this beast already.

One comment regarding making the schemas a bit more specific - set "additionalProperties" to false on all objects in JSON schema.

The other improvement I see, is to mark more fields as required based on experience (but this may easily explode into big schema, if we need to use "anyOf" or similar approaches).

modules/vpc-sc/schemas/ingress-policy.schema.json
ludoo: for ingress policies, access levels are not required: if you specify resources (i.e. projects) then you don't need access levels; if you don't have resources, then you need either a star of a list of access levels

TL;DR variable type is correct

ludoo:
> Generally - ❤️ ❤️ ❤️ - as this may help me a lot in the future and I was bitten by this beast already.
> 
> One comment regarding making the schemas a bit more specific - set "additionalProperties" to false on all objects in JSON schema.
> 
> The other improvement I see, is to mark more fields as required based on experience (but this may easily explode into big schema, if we need to use "anyOf" or similar approaches).

schema needs to be a 1:1 translation of the variable type :)

modules/vpc-sc/schemas/ingress-policy.schema.json
wiktorn: I'm not saying it's not correct, I'm just saying it could be better, but IIUC, this means going into `anyOf`, where one requires `access_levels`  and the other that requires `resources`, so then we encode that one of them is provided.

But it will be PITA to encode those rules in JSON Schema, goes outside what we can express in HCL, so I would just keep in mind of this limitation.

PR #2488, opened 2024-08-08
Introduce YAML schema validation for YAML examples

juliocc:
@wiktorn I'm gonna need your help to ensure I'm not breaking the e2e stuff

### DIFF
--- a/modules/project-factory/README.md
+++ b/modules/project-factory/README.md
@@ -215,23 +215,23 @@ name: Foo (level 1)
 iam:
   roles/viewer:
     - group:a@example.com
-# tftest-file id=h-0-0 path=data/hierarchy/foo/_config.yaml
+# tftest-file id=h-0-0 path=data/hierarchy/foo/_config.yaml schema=folder.schema.json
 ```
 
 ```yaml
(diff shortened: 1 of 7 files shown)


modules/project-factory/README.md
```diff
@@ -215,23 +215,23 @@ name: Foo (level 1)
 iam:
   roles/viewer:
     - group:a@example.com
-# tftest-file id=h-0-0 path=data/hierarchy/foo/_config.yaml
+# tftest-file id=h-0-0 path=data/hierarchy/foo/_config.yaml schema=folder.schema.json
```
ludoo, line 218: how do you look for the schema without a path? can we support cases where the schema lives outside the module's folder, like in FAST stages?

ludoo (approved):

modules/project-factory/README.md
juliocc: I look inside the schemas folder of the containing module. 

> can we support cases where the schema lives outside the module's folder, like in FAST stages?

Of course. We can do whatever we want :) When get to FAST I'll take care of updating it

PR #2490, opened 2024-08-09
Bind schemas to factory files, add support for groups in VPC-SC schema

wiktorn:
Groups and principals are [available in preview](https://cloud.google.com/vpc-service-controls/docs/release-notes#July_31_2024)

Inspired by #2489, though this means that JSON schema is more strict than `variables.tf`

---
**Checklist**


.vscode/settings.json
```diff
@@ -6,5 +6,25 @@
...
-  ]
-}
\ No newline at end of file
+  ],
+  "yaml.schemas": {
+    "modules/project-factory/schemas/budget.schema.json": [
+      "budgets/**/*yaml"
+    ],
+    "modules/project-factory/schemas/folder.schema.json": [
+      "folders/**/*yaml"
```
ludoo, line 15: this should be either `folders` or `hierarchy`

ludoo:
Ca we retrofit the validation to variables.tf in both vpc-sc module and FAST stage?

ludoo (approved):

.vscode/settings.json
```diff
@@ -6,5 +6,25 @@
       ],
       "url": "http://json-schema.org/draft-07/schema#"
     }
-  ]
-}
\ No newline at end of file
+  ],
+  "yaml.schemas": {
```
ludoo, line 10: I would use a `data/` prefix for values, so as to pinpoint factory data.

ludoo:
`yaml.schemas` does not seem to work
image

wiktorn:
> `yaml.schemas` does not seem to work
> 
> image

Are you running multi-root workspace? Then [this may be applicable](https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#multi-root-schema-association), and you need specify `yaml.schemas` in  root of the multi root workspace project.

.vscode/settings.json
wiktorn: Added `data/**/` prefix, so we can have `data/<stage-name>/<object_class>` structure when pointing to directory in `factories_config`.

ludoo: `stage-name` is not needed, paths will always be `data/object-class` for both modules and FAST :)

ludoo: well I lied, I point FAST factories for my org to my own data folder which is structured exactly like you wrote 😊, so we should support both `data/object-class` and `data/stage/object-class`

wiktorn: both works with `data/**/`

ludoo: yep, absolutely that was the intention of my last message

PR #2491, opened 2024-08-09
Organization module factory schemas

ludoo:


fast/stages/0-bootstrap/data/custom-roles/gcve_network_admin.yaml
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# yaml-language-server: $schema=../../../../../modules/organization/schemas/custom-role.schema.json
```
wiktorn, line 15: I'm kind of reluctant about adding explicit clauses to YAML examples, as if you move the files and relative path changes, you need to adapt also the header as it takes precedence over `.vscode/settings.json`

juliocc (approved):

fast/stages/0-bootstrap/data/custom-roles/gcve_network_admin.yaml
ludoo: I'm not and it's the whole purpose of this. This is not an example, but an integral part of FAST.

PR #2508, opened 2024-08-20
Add support for disable default snat

okguru1:
Adding support for `disable_default_snat`, which (as the name suggests) disables the default snat. This option is available on the root resource, so all this PR does is expose this option.

---
**Checklist**


ludoo (commented):
Thanks for this. A couple comments below.

Main one is we work very hard to minimize the variable surface, and I would prefer not adding one variable for a tiny feature. This should be an additional optional attribute in one of the existing `_config` variables, possibly `vpc_config` as it's the one most closely related to this.

The second comment is I would prefer using a dynamic block and a `null` default, so that the block is transparently skipped when the attribute is not explicitly set. This prevented issues with recreation of existing clusters for other attributes, and we might want to adopt the same approache here, even just for consistency's sake.

And last, do you know if this is also supported for autopilot? If it is, we should update both cluster modules together.

okguru1:
@ludoo thanks for the prompt feedback. adjusted the PR to utilize a dynamic block (with conditional on null), had the variable moved inside `vpc_config` (with default null), and aligned the autopilot module the same way.

ludoo:
> @ludoo thanks for the prompt feedback. adjusted the PR to utilize a dynamic block (with conditional on null), had the variable moved inside `vpc_config` (with default null), and aligned the autopilot module the same way.

super quick :)

ludoo:
Can you pull and merge master? I cannot use your fork...

modules/gke-cluster-autopilot/main.tf
```diff
@@ -94,6 +94,13 @@ resource "google_container_cluster" "cluster" {
     }
   }
 
+  dynamic "default_snat_status" {
+    for_each = var.vpc_config.disable_default_snat == null ? [""] : []
```
ludoo, line 98: I would use something like

```hcl
dynamic "default_snat_status" {
  for_each = var.vpc_config.disable_default_snat == null ? [] : [""]
  content {
    disabled = var.vpc_config.disable_default_snat
  }
}
```

okguru1: good catch 😅

okguru1:
> Can you pull and merge master? I cannot use your fork...

yes, updated my master branch, or did you mean something else?

ludoo (approved):
last fix on redundant default optional value then this is good to go, many thanks :)

modules/gke-cluster-standard/variables.tf
```diff
@@ -421,6 +422,7 @@ variable "release_channel" {
 variable "vpc_config" {
   description = "VPC-level configuration."
   type = object({
+    disable_default_snat       = optional(bool, null)
```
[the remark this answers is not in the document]
okguru1: done

ludoo:
Thanks a lot for your patience :)

ludoo:
ach, tfdoc strikes again :)

okguru1:
> ach, tfdoc strikes again :)

hahaha yea missed this, if I continue contributing I should probably make a precommit with all of these :)

ludoo:
> > ach, tfdoc strikes again :)
> 
> hahaha yea missed this, if I continue contributing I should probably make a precommit with all of these :)

hope you do :)

ludoo:
Oh God, one more merge :) if you set your fork to allow changes from this repo I could push the button for you

okguru1:
> Oh God, one more merge :) if you set your fork to allow changes from this repo I could push the button for you

done, I have to learn how to clickops those fork permissions, will do for next times

PR #2520, opened 2024-08-26
Add e2e pubusb errors

juliocc:
Adds pubsub to the list of jitsi service accounts for the default project. 

I wonder how this was working in the past...

### DIFF
--- a/tests/examples_e2e/setup_module/main.tf
+++ b/tests/examples_e2e/setup_module/main.tf
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


wiktorn:
I was thinking about different approach for that - to create explicit depends_on for IAM grants, that are in fixtures for the tests that are failing. It looks like:
* sometimes pubsub SA is missing permissions for GCS bucket
* sometimes cloud function SA is missing permissions on pubsub

(at least based on last execution).

And this could explain, why it works sometimes.

wiktorn:
I mean, test case should depend on grants given in: `fixtures/functions-default-sa-iam-grants.tf`  and `fixtures/gcs.tf`.

There might be other cases for that.

juliocc:
It's not just the Ian binding, you also have to trigger the creation of the service agent

wiktorn (approved):

PR #2524, opened 2024-08-27
Add dependencies on IAM grants, make E2E tests depend on IAM grants

wiktorn:
Fix missing IAM dependencies.

---
**Checklist**

### DIFF
--- a/modules/alloydb/README.md
+++ b/modules/alloydb/README.md
@@ -161,6 +161,9 @@ module "alloydb" {
   encryption_config = {
     primary_kms_key_name = var.kms_key.id
   }
+  depends_on = [
+    google_kms_crypto_key_iam_member.alloydb_encrypt_decrypt
+  ]
 }
 
 # tftest modules=1 resources=3 fixtures=fixtures/alloydb-kms-iam-grant.tf inventory=cmek.yaml e2e
--- a/modules/alloydb/README.md
+++ b/modules/alloydb/README.md
@@ -163,7 +163,7 @@ module "alloydb" {
   }
 }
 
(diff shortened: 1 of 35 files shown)


wiktorn:
[E2E test run](https://github.com/GoogleCloudPlatform/fast-e2e/actions/runs/10648462602)

juliocc:
@wiktorn I would really prefer if we avoid using `depends_on` on examples. Having a `depends_on` there is an artifact of the e2e architecture. This is not how you would add those bindings it if you use the project module. 

Same thing with these bindings:
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/ecb83e9c4b26a018472d21473153f8c437e7f2c8/tests/fixtures/kms-global-regional-keys.tf#L26-L32

These are not as bad since they're in a fixture but examples are intended to show idiomatic code. I'd argue that we should drop e2e from those examples if we have to introduce code to support it

ludoo:
> @wiktorn I would really prefer if we avoid using `depends_on` on examples. Having a `depends_on` there is an artifact of the e2e architecture. This is not how you would add those bindings it if you use the project module.
> 
> Same thing with these bindings:
> 
> https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/ecb83e9c4b26a018472d21473153f8c437e7f2c8/tests/fixtures/kms-global-regional-keys.tf#L26-L32
> 
> These are not as bad since they're in a fixture but examples are intended to show idiomatic code. I'd argue that we should drop e2e from those examples if we have to introduce code to support it

100% agree, e2e needs to take second not first place in cases like these

wiktorn:
Closing this to save your cycles. For PR prioritizing removal of `depends_on` see #2546

PR #2530, opened 2024-08-27
Add managed folders support to `gcs` module

juliocc:
- Add support for managed folders
- Add support for RPO
- Make STANDARD the default storage class
- Add `iam_by_principals`


ludoo (approved):
very nice, my only nit is on `Keys must end in /`, can we interpolate it and have more readable keys? ignore me ofc if the current approach makes more sense

juliocc:
> very nice, my only nit is on `Keys must end in /`, can we interpolate it and have more readable keys? ignore me ofc if the current approach makes more sense

Removed the need for the trailing /

PR #2532, opened 2024-08-28
Refactor of FAST resource management stage

ludoo:
Action items before this is ready to be merged

- [x] documentation
  - [x] stage overview
  - [x] stage 1 (include some patterns for top level and stage 3s)
  - [x] change `stage-links.sh` to new format in all stage READMES
- [x] ngfw
  - [x] move to stage 2
  - [x] check that interfaces from and to other stage 2 work
- [x] stage 2
  - [x] align interface (optional env folders, tag bindings)
  - [x] validate apply from scratch in new org
  - [ ] validate apply in existing org with moved blocks
- [x] stage 3
  - [x] gcve
    - [x] align interface and remove wrapper module and blueprint
    - ~[ ] validate apply from scratch in new org~
    - ~[ ] validate apply in existing org with moved blocks~
  - [x] gke
    - [x] remove wrapper module and blueprint (add factory after merging)
  - [x] data platform
    - [x] remove stage in preparation of new code
- [x] CI/CD
  - [x] test stage 2 and 3 workflows
  ~- [ ] reintegrate TFE support (after merging)~
- [x] multitenant
  - [x] test stage 1mt works
  - [x] test one tenant for each type

Highlight of the approach copy/pasted below from internal discussions.

- stage 2s are important and typically end up being customized, they keep their own individuality here and are just made compliant to a standard / configurable via a `fast_stage2` variable; these are "named stages" that cross environments and interact with the org in their own way
- stage 3s are a mob of similar things tied to an environment, and they are now rubber stamped from an identical pattern; the pattern accounts for some differences (e.g. no folders as in Luca's netsec), some "guardrailed" reqs they can demand from stage 2s and org in terms of IAM

As Julio put it, resman stage now becomes a broker between stage 2s and 3s. And as Simo observed this actually brings us closer to blueprint === stage 3.

Putting this out in draft to start getting reviews especially on locals which hide all the complexity of the new approach.


fast/stages/1-resman/billing.tf
```diff
@@ -17,20 +17,38 @@
...
-    try(module.branch-dp-dev-sa[0].iam_email, null),
-    try(module.branch-dp-prod-sa[0].iam_email, null),
-    try(module.branch-gcve-dev-sa[0].iam_email, null),
-    try(module.branch-gcve-prod-sa[0].iam_email, null),
-    try(module.branch-gke-dev-sa[0].iam_email, null),
-    try(module.branch-gke-prod-sa[0].iam_email, null)
-  ])
+  billing_iam = merge(
+    # stage 2
+    var.fast_stage_2.networking.enabled != true ? {} : {
```
juliocc, line 22: nit: I'm sure there's a reason but this condition is different to the others (`x != true` vs `!x`)

fast/stages/1-resman/organization.tf
```diff
@@ -48,60 +52,51 @@ module "organization" {
...
-      }
-    }
+      )
+    },
     (var.tag_names.environment) = {
       description = "Environment definition."
       iam         = try(local.tags.environment.iam, {})
       values = {
-        development = {
+        (var.environment_names["dev"]) = {
```
juliocc, line 69: Would it be better to iterate `var.environment_names` here and create values for each one?

fast/stages/1-resman/variables-fast.tf
```diff
@@ -56,8 +56,8 @@ variable "custom_roles" {
   type = object({
     gcve_network_admin              = string
     network_firewall_policies_admin = string
-    ngfw_enterprise_admin           = string
-    ngfw_enterprise_viewer          = string
+    ngfw_enterprise_admin           = optional(string)
+    ngfw_enterprise_viewer          = optional(string)
```
juliocc, line 60: is this `optional` ok?

fast/stages/1-resman/organization.tf
ludoo: Ah yes, this was before I added the environments variable, good point.

ludoo: pfff you're right :)

fast/stages/1-resman/variables-fast.tf
ludoo: it's only used if you apply netsec, but we can make it mandatory

juliocc: actually let's make the other s3 (only gcve I think) roles optional too

ludoo:
Closing this, and reopening against fast-dev

PR #2536, opened 2024-08-28
Add support for google provider 6.x

sruffilli:
---
**Checklist**


ludoo:
this should wait for the release

juliocc:
lgtm. As @ludoo said, I'd do a release just before mergin this

juliocc (approved):

ludoo (approved):
Update release comment in versions.tf to say

`# Fabric release: v35.0.0-pre`

wiktorn:
Before moving forward with this - there is a few resources that added `deletion_policy`:
- folder
- cloud run service v2

Plus others, and they do default to `true`. It means, that for all E2E tests, we need to explicitly set to false, or take make it default to true in modules.

You can check where we are at this moment in [this run](https://github.com/GoogleCloudPlatform/fast-e2e/actions/runs/10648336304)

wiktorn:
In #2502 @juliocc  added `deletion_policy` to `modules/project`  with default value `DELETE`.

Shall we do the same for `cloudrun-v2` and `folder`?

juliocc:
> In #2502 @juliocc added `deletion_policy` to `modules/project` with default value `DELETE`.
> 
> Shall we do the same for `cloudrun-v2` and `folder`?

@sruffilli already added support to `modules/folder` and `modules/cloud-run-v2` in this same PR: 
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2536/files#diff-e07eeb8ca41c3021453f0ba9b664e6ef632e826cdf1246179d0163be76847ffc
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2536/files#diff-b93d21eeff163e147d028de14911d6319d59df27f5389606b808700ade655f46

And I just pushed a [small commit](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2536/commits/a651f41fbaec9d935048ba4bab031e7d381a9b03) for `google_cloud_run_v2_job` resources.

wiktorn:
> > In #2502 @juliocc added `deletion_policy` to `modules/project` with default value `DELETE`.
> > Shall we do the same for `cloudrun-v2` and `folder`?
> 
> @sruffilli already added support to `modules/folder` and `modules/cloud-run-v2` in this same PR:
> 
> * https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2536/files#diff-e07eeb8ca41c3021453f0ba9b664e6ef632e826cdf1246179d0163be76847ffc
> * https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2536/files#diff-b93d21eeff163e147d028de14911d6319d59df27f5389606b808700ade655f46
> 
> And I just pushed a [small commit](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2536/commits/a651f41fbaec9d935048ba4bab031e7d381a9b03) for `google_cloud_run_v2_job` resources.

Yep, but as of now, we have with default value of `null`, that is using provider default (enforce deletion protection), which is different from what we do in `modules/project` where we override provider default.

juliocc:
> Yep, but as of now, we have with default value of `null`, that is using provider default (enforce deletion protection), which is different from what we do in `modules/project` where we override provider default.

I see. For `project` I tried to keep to behavior we've always had because it's a core module and didn't want to break everyone's code.  For `folder` we should probably do the same. 

For `cloud-run-v2` I'm I don't know. It' not used that much so I think we can keep the provider's default and document the change. wdyt?

Issue #2512, opened 2024-08-21
fast: allow existing users to migrate when incompatible changes are included in new versions


gustavovalverde:
**Describe the bug**
Some users would like to keep their changes in sync with this upstream, but the complexity of doing it might outweigh some of the benefits (eg. time saving) if the time to migrate is bigger than the time to build a full FAST structure from scratch.

**To Reproduce**
After doing a production implementation using [v29 ](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/releases/tag/v29.0.0) try to update with upstream and use [v33](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/releases/tag/v33.0.0)

This requires going through a few months of changes, and hundreds of lines of codes to identify each `mv` command, and in some cases it might not be enough as new logic or structures would require further analysis.

**Expected behavior**
Having clear instructions on how to migrate, or use [moved blocks syntax](https://developer.hashicorp.com/terraform/language/modules/develop/refactoring#moved-block-syntax) to allow Terraform to automatically identify the changes and make the migrations.

This could be complemented with migration instructions in the release changelog.

**Result**
Terraform output and/or error messages

**Additional context**
FAST is meant to be used in organizations for production use cases, and part of this production lifecycle should be keeping the FAST implementation up to date with the best practices implemented in this repo.

Related PRs to this issue: 
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1843
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2115
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2203
- https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2254

ludoo:
Gustavo, this will never happen and it was discussed many times before: FAST is a toolkit not a product, and its primary goal is to allow our team to quickly spin up standards compliant LZs.

Once an LZ is deployed, by definition it matches requirements and changes should just follow its natural evolution as requirements change.

Dealing with version upgrades is irrelevant to our primary goal, and would put an impossible strain on our limited resources. We share this code as we hope it's useful to the wider community, but this can only be maintained if it does not become a product but remains a toolkit. Our main job is working with customers, not maintaining code.

gustavovalverde:
[...]

PR #2541, opened 2024-08-30
Moved blocks and fix to resman for FAST v33-v34 transition

ludoo:
This is a followup from #2512 and a simple way we should provide guidance on updating FAST for every release. The fast-test-05 org is pinned to last version, and running apply on bootstrap, resman, and net-a allows us to write moved blocks and note down major changes.

The release notes for v34 should add a section

### Bootstrap stage

No destructive changes. A few IAM bindings are reapplied cleanly.

### Resource management stage

Network security IaC resources change names from `resman-netsec` to `resman-nsec` and need recreation. Network security state should be transitioned to local before applying resource management, and re-transitioned to remote after refreshing resman output files and netsec provider.

Project factory dev and prod resources will change internal names, the moved blocks in the provided file should seamlessly rename them in state.


fast/stages/1-resman/branch-nsec.tf
```diff
@@ -16,13 +16,25 @@
 
 # tfdoc:file:description Network security stage resources.
 
-# automation service account
+# TODO: remove in v35.0.0
 
 moved {
```
sruffilli, line 21: Why aren't those in a `moved` file?

ludoo: because they are not a delta between versions

fast/stages/1-resman/v33.0.0-v34.0.0.tf
juliocc, line 1: why the symlink?

juliocc (approved):

fast/stages/1-resman/v33.0.0-v34.0.0.tf
ludoo: there's a symlink? :)

ludoo: good catch

PR #2544, opened 2024-08-30
GCVE network mode for 2-networking-b-nva  stage

eliamaldini:
Hi Team, this is a draft of a NVA stage change to add GCVE compatibility.

The code is almost ready, I'm working on the Readme file and tests....

Main changes:
- added a variable to manage the network mode
- code adaptations for simple and ncc_ra options to support the new variable
- added two new regional truted VPC
- added a new NVA deployment option to support regional VPCs
- VPC routing

---
**Checklist**


ludoo:
I'm not sure we should have a ton of GCVE-specific resources in the network stage. But I have no idea about an alternative. Let's sit down and discuss all together.

fast/stages/2-networking-b-nva/net-landing.tf
```diff
@@ -64,6 +69,20 @@ module "dmz-vpc" {
       next_hop_type = "gateway"
       priority      = 1000
     }
+    to-gcve-primary = {
```
ludoo, line 72: this will fail if gcve is not enabled

fast/stages/2-networking-b-nva/variables.tf
```diff
@@ -111,6 +113,17 @@ variable "gcp_ranges" {
   }
 }
 
+variable "network_mode" {
+  description = "Selction of the network mode to deploy"
+  type        = string
+  default     = "simple"
+  nullable    = false
+  validation {
+    condition     = contains(["simple", "ncc_ra", "gcve"], var.network_mode)
```
ludoo, line 122: this is mixing support for a product (gcve) with a network design switch; we need to find a better option IMHO

fast/stages/2-networking-b-nva/net-landing.tf
eliamaldini: yes I noticed.....just fixed

eliamaldini:
> I'm not sure we should have a ton of GCVE-specific resources in the network stage. But I have no idea about an alternative. Let's sit down and discuss all together.

An idea could be, instead of referring to GCVE we can call the option in a generic way like "regional_trusted". It would be a design option that can be used not only for GCVE but in all the cases where regional trusted are required

ludoo:
> > I'm not sure we should have a ton of GCVE-specific resources in the network stage. But I have no idea about an alternative. Let's sit down and discuss all together.
> 
> An idea could be, instead of referring to GCVE we can call the option in a generic way like "regional_trusted". It would be a design option that can be used not only for GCVE but in all the cases where regional trusted are required

Makes a lot of sense, but then would this be an additional net design or a "flavor" of the current one? And how would it combine with ncc-ra support?

eliamaldini:
> > > I'm not sure we should have a ton of GCVE-specific resources in the network stage. But I have no idea about an alternative. Let's sit down and discuss all together.
> > 
> > 
> > An idea could be, instead of referring to GCVE we can call the option in a generic way like "regional_trusted". It would be a design option that can be used not only for GCVE but in all the cases where regional trusted are required
> 
> Makes a lot of sense, but then would this be an additional net design or a "flavor" of the current one? And how would it combine with ncc-ra support?

I would leave the net experts to undersatnd if a dedicated stage is better. 

GCVE mode cannot be combined with ncc-ra (regional VPC doesn't require ncc/bgp routing) if you see the code,  the three net modes  (simple, ncc-ra, gcve) are mutually exclusive. you can have just one of them enabled.

ludoo:
> > > > I'm not sure we should have a ton of GCVE-specific resources in the network stage. But I have no idea about an alternative. Let's sit down and discuss all together.
> > > 
> > > 
> > > An idea could be, instead of referring to GCVE we can call the option in a generic way like "regional_trusted". It would be a design option that can be used not only for GCVE but in all the cases where regional trusted are required
> > 
> > 
> > Makes a lot of sense, but then would this be an additional net design or a "flavor" of the current one? And how would it combine with ncc-ra support?
> 
> I would leave the net experts to undersatnd if a dedicated stage is better.
> 
> GCVE mode cannot be combined with ncc-ra (regional VPC doesn't require ncc/bgp routing) if you see the code, the three net modes (simple, ncc-ra, gcve) are mutually exclusive. you can have just one of them enabled.

sure, but the point is gcve *depends* on simple mode, it's not *a different* net mode so we need  different way of expressing this

fast/stages/2-networking-b-nva/variables.tf
sruffilli: I agree, this is a "regional VPC" networking design, there's nothing really gcve specific.

So, `regional_vpc`?

eliamaldini: renamed in regional_vpc

fast/stages/2-networking-b-nva/diagram-regional.svg
sruffilli, line 1: Nit: can you reroute the connectors so that they don't overlap? Feel free to pushback :)

fast/stages/2-networking-b-nva/net-dev.tf
```diff
@@ -19,7 +19,7 @@
 module "dev-spoke-project" {
   source          = "../../../modules/project"
   billing_account = var.billing_account.id
-  name            = "dev-net-spoke-0"
+  name            = "dev-net-spoke-00"
```
sruffilli, line 22: revert to -0

fast/stages/2-networking-b-nva/net-landing.tf
```diff
@@ -19,7 +19,7 @@
 module "landing-project" {
   source          = "../../../modules/project"
   billing_account = var.billing_account.id
-  name            = "prod-net-landing-0"
+  name            = "prod-net-landing-00"
```
sruffilli, line 22: revert to -0

fast/stages/2-networking-b-nva/net-landing.tf
```diff
@@ -30,9 +30,14 @@ module "landing-project" {
     "stackdriver.googleapis.com",
     ],
     (
-      var.enable_ncc_ra
+      (var.network_mode == "ncc_ra")
```
sruffilli, line 33: You don't need brackets here

fast/stages/2-networking-b-nva/net-landing.tf
```diff
@@ -57,14 +62,31 @@ module "dmz-vpc" {
...
+  routes = merge(
+    {
+      default = {
+        dest_range    = "0.0.0.0/0"
+        next_hop      = "default-internet-gateway"
+        next_hop_type = "gateway"
+        priority      = 1000
+      }
+    },
+    # Uncomment to enable cross regional VPC traffic
```
sruffilli, line 74: Is uncommenting enough or does the NVA require some additional programming?
Add a comment in that case, resolve this otherwise.

fast/stages/2-networking-b-nva/net-prod.tf
```diff
@@ -15,11 +15,21 @@
...
+      ? module.ilb-nva-landing["secondary"].forwarding_rule_addresses[""]
+    : module.ilb-regional-nva-landing["secondary"].forwarding_rule_addresses[""])
+  }
+}
 
 module "prod-spoke-project" {
   source          = "../../../modules/project"
   billing_account = var.billing_account.id
-  name            = "prod-net-spoke-0"
+  name            = "prod-net-spoke-00"
```
sruffilli, line 32: revert to -0

fast/stages/2-networking-b-nva/net-prod.tf
```diff
@@ -15,11 +15,21 @@
  */
 
 # tfdoc:file:description Production spoke VPC and related resources.
+locals {
+  nva_load_balancers = (var.network_mode == "ncc_ra") ? null : {
```
sruffilli, line 19: This is really opaque. 
I would instead define two locals (e.g. `_simple_nva_lb` and `_regional_nva_lb`, and conditionally assign their value to `nva_load_balancers`.
As an alternative write a verbose comment to what is happening there 😆

sruffilli (changes requested):
LGTM pending the highlighted changes

fast/stages/2-networking-b-nva/net-landing.tf
eliamaldini: the traffic at the NVA is managed through the default route... no changes are requries at the NVA level to implement cross regiona VPC traffic

fast/stages/2-networking-b-nva/net-landing.tf
eliamaldini: fixed

fast/stages/2-networking-b-nva/net-landing.tf
eliamaldini: fixed

fast/stages/2-networking-b-nva/net-dev.tf
eliamaldini: fixed

fast/stages/2-networking-b-nva/net-prod.tf
eliamaldini: fixed

fast/stages/2-networking-b-nva/diagram-regional.svg
eliamaldini: fixed

fast/stages/2-networking-b-nva/net-prod.tf
eliamaldini: fixed

PR #2571, opened 2024-09-15
Bigtable authorized views

apichick:
Added authorized views to tables in bigtable-instance module
---
**Checklist**


modules/bigtable-instance/README.md
```diff
@@ -68,10 +66,52 @@ module "bigtable-instance" {
...
+  }
+  tables = {
+    test = {
+      split_keys = ["a", "b", "c"]
+      column_families = {
+        cf1 = {}
+        cf2 = {}
+        cf3 = {}
+      }
+    }
```
wiktorn, line 89:
```suggestion
```
Shouldn't `authorized_views` be under the table in the structure?

Then I get failure in example, that `deletion_protection` is necessary (which looks good, according to variables.tf

wiktorn (commented):
Thank you, spotted one bug in the test. Also, took a quick look at how it should work, and I fear, that there is little value of having `authorized_views` without the possibility to set IAM policies on them in provider. Implementation of [setIamPolicy](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.tables.authorizedViews/setIamPolicy) on authorized view is still missing. So I wonder, how do you grant access to authorized view without granting access to the source table.

The other thing I'm curious about is `deletion_protection` - not sure why it is important on the view, when view should not hold any data.

PR #2549, opened 2024-09-02
Fix FAST stages to support multi-tenant resman

ZZelle:
The FAST stage 1-resman fails in a tenant for multiple reasons:
- it tries to create top level folders with the organization as parent instead of the `local.root_node`
- it uses the tag "project-factory" without creating it (it's done correctly without multi-tenancy)
- the stage 1-tenant-factory generates 0-bootstrap.auto.tfvars.json without some data required by 1-resman

---
**Checklist**

I applicable, I acknowledge that I have:
- [X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X] Ran `terraform fmt` on all modified files
- [X] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [ ] Made sure all relevant tests pass

PR #2579, opened 2024-09-18
FAST resman mt fixes

ludoo:
Implements changes from stuck PR #2549


fast/stages/1-tenant-factory/variables-fast.tf
```diff
@@ -56,11 +56,14 @@ variable "custom_roles" {
   type = object({
     gcve_network_admin              = string
```
sruffilli, line 57: Following the same logic, should also at least gcve_network_admin be optional?

sruffilli: _I'm guessing this is a merge artifact_

sruffilli (approved):

fast/stages/1-tenant-factory/variables-fast.tf
ludoo: yep, we don't have a rule for custom roles only needed by specific stages atm, this is all going to change anyway once we move to new resman, resolvign this but let's revisit ASAP and define a rule

PR #2594, opened 2024-09-26
Added unique suffix to the certname to prevent conflict when updating

SadriG91:
I added unique suffix to the certname to prevent conflict when updating domains
---
**Checklist**

### DIFF
--- a/modules/net-lb-app-ext/main.tf
+++ b/modules/net-lb-app-ext/main.tf
@@ -50,18 +50,39 @@ resource "google_compute_global_forwarding_rule" "default" {
 resource "google_compute_ssl_certificate" "default" {
   for_each    = var.ssl_certificates.create_configs
   project     = var.project_id
-  name        = "${var.name}-${each.key}"
+  name        = "${var.name}-${each.key}-${random_id.suffix_ssl[each.key].hex}"
   certificate = trimspace(each.value.certificate)
   private_key = trimspace(each.value.private_key)
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "random_id" "suffix_ssl" {
+  for_each    = var.ssl_certificates.create_configs
+  byte_length = 4
+  keepers = {
+    cert_key_hash = base64sha256(trimspace(each.value.certificate) + trimspace(each.value.private_key))
+  }
 }
 
 resource "google_compute_managed_ssl_certificate" "default" {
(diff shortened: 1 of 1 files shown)


juliocc (changes requested):
Hi @SadriG91,

Thanks for the contribution! We generally avoid using random suffixes/prefixes. Could you instead add an optional `name` field to `var.ssl_certificates` and use `coalesce()` to determine the name? 

This would allow users to specify a custom name while providing a clear default.

wiktorn:
@SadriG91 did you try following the instructions [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/net-lb-app-ext#deploying-changes-to-load-balancer-configurations), that should help avoid this issue?

SadriG91:
@wiktorn the link basically says changes will fail if in use, so how would the change be done then? What if it's during a planned maintenance window for example then shouldn't the module allow for that?

SadriG91:
> Hi @SadriG91,
> 
> Thanks for the contribution! We generally avoid using random suffixes/prefixes. Could you instead add an optional `name` field to `var.ssl_certificates` and use `coalesce()` to determine the name?
> 
> This would allow users to specify a custom name while providing a clear default.

The idea is that if you for example add a subdomain and want to add that to the cert it will get recreated and since it can't be detached while in use we need to add the flag `create_before_destroy` and for that we need a unique id or name for the cert otherwise name conflict because of name exist already

wiktorn:
> @wiktorn the link basically says changes will fail if in use, so how would the change be done then? What if it's during a planned maintenance window for example then shouldn't the module allow for that?

if your configuration is:
```hcl
  ssl_certificates = {
    managed_configs = {
      default = {
        domains = ["${module.addresses.global_addresses["glb-0"].address}.nip.io"]
      }
  }
```
You change it to:
```hcl
  ssl_certificates = {
    managed_configs = {
      default = {
        domains = ["${module.addresses.global_addresses["glb-0"].address}.nip.io"]
      }
      new = {
        domains = [
          "${module.addresses.global_addresses["glb-0"].address}.nip.io",
          "${replace(module.addresses.global_addresses["glb-0"].address, ".", "-")}.nip.io",
        ]
      }
    }
  }
```
(or whatever new certificate should look like), and once that is provisioned and working, you can remove the `default` part.

ludoo:
closing this, feel free to reopen with a different feature set

PR #2595, opened 2024-09-27
Allow manage existing SSM instance

lnesteroff:
Modified the secure-source-manager-instance module to allow management of an existing instance (eg create repos and manage IAM). Also removed the location attr from the repositories var as it needs to be the same as the instance, so this would save having to enter a location for each repo being created.

---
**Checklist**

### DIFF
--- a/modules/secure-source-manager-instance/iam.tf
+++ b/modules/secure-source-manager-instance/iam.tf
@@ -33,19 +33,18 @@ locals {
 
 resource "google_secure_source_manager_instance_iam_binding" "authoritative" {
   for_each    = var.iam
-  project     = var.project_id
-  location    = var.location
-  instance_id = var.instance_id
+  project     = try(google_secure_source_manager_instance.instance.project, var.project_id)
+  location    = try(google_secure_source_manager_instance.instance.location, var.location)
+  instance_id = try(google_secure_source_manager_instance.instance.instance_id, var.instance_id)
   role        = each.key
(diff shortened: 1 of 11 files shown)


ludoo (approved):

modules/secure-source-manager-instance/iam.tf
```diff
@@ -33,29 +33,31 @@ locals {
 
 resource "google_secure_source_manager_instance_iam_binding" "authoritative" {
   for_each    = var.iam
-  project     = google_secure_source_manager_instance.instance.project
-  location    = google_secure_source_manager_instance.instance.location
-  instance_id = google_secure_source_manager_instance.instance.instance_id
+  project     = try(google_secure_source_manager_instance.instance[0].project, var.project_id)
+  location    = try(google_secure_source_manager_instance.instance[0].location, var.location)
+  instance_id = try(google_secure_source_manager_instance.instance[0].instance_id, var.instance_id)
```
juliocc, line 38: nit: is there any way to remove the try() calls. Maybe having a local for each using a ternary? Something like
```terraform
locals {
  location = var.instance_create ? google_secure_source_manager_instance.instance[0].location : var.location
}
```

with this you can also remove the depends_on on the other resources

juliocc (changes requested):
LGTM except for small comment. See inline

juliocc (approved):

PR #2602, opened 2024-10-03
WIP: Add Azure DevOps option for FAST CICD

lnesteroff:
Adding option to configure FAST CICD for Azure DevOps pipelines. Will be testing e2e over the next week. This creates the WIF provider and modifies the vars to include the Azure oid assertion as Azure doesn't seem to pass through repo. oid seems to be bound to an identity and a pipeline, so should be sufficiently unique/secure.

---
**Checklist**


fast/stages/0-bootstrap/cicd.tf
```diff
@@ -24,11 +24,18 @@ locals {
...
-      name             = v.name
-      principal_branch = local.workload_identity_providers[k].principal_branch
-      principal_repo   = local.workload_identity_providers[k].principal_repo
+      issuer     = local.workload_identity_providers[k].issuer
+      issuer_uri = try(v.oidc[0].issuer_uri, null)
+      name       = v.name
+      principal_branch = (
+        k == "azure" ? null : local.workload_identity_providers[k].principal_branch
+      )
+      principal_repo = (
```
ludoo, line 33: Can we use `principal_repo` for Azure, and then in the IAM bindings when we use it we coalesce repo and oid, so we don't have to support a third format? I would interpret the formats not as "this needs interpolating the repo" but as "this is not restricting to a branch", so the definition could also apply to Azure.

Or am I way off mark? (possible, never used Azure DevOps)

lnesteroff: There doesn't seem to be any Azure claims that we could use for repo/branch (https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference). We could reuse the variable, but it's not a repo or branch being passed unfortunately.

ludoo: I just mean using only two formats: one accepting a single arg (coalesce of repo or oid), and one accepting two (repo and branch). We might also rename them to make it clearer.

lnesteroff: Ahh, understood. Yeah that makes sense. We may need to rename it though, the other common option I'm exploring is with TFE which uses workspaces instead of repos. Maybe instead of az_oid it should be wif_subject, then it can be reused for TFE and others.

ludoo: Yep TFE has plan/apply which kind of map to our ro/rw service accounts (@averbuks used it recently with a customer), let's discuss how to find a new naming convention that fits all use cases.

fast/stages/0-bootstrap/cicd.tf
```diff
@@ -86,14 +86,14 @@ module "automation-tf-cicd-sa" {
     "roles/iam.workloadIdentityUser" = [
       each.value.branch == null
       ? format(
-        local.workload_identity_providers_defs[each.value.type].principal_repo,
+        local.workload_identity_providers_defs[each.value.type].principal_subject,
```
ludoo, line 89: one step forward, but I would still iterate on naming a bit; copy/pasting from chat: "I think we need one more naming iteration, as subject and branch seem two different abstractions; we had repo and branch which both refer to the actual SCM used, now we have subject and branch which refer one to an attribute in the token, the other to the SCM; let's settle on one or the other"

PR #2606, opened 2024-10-04
feat: implement the new iam interface in `artifact-registry`

NitriKx:
This PR implements the new IAM interface in the `atifact-registry` module (#2605)

In order to have a consistent interface across the different modules, this update will cause a delete/create of the existing permissions set in the `iam` variable: 

```
  # module.artifact_registry.google_artifact_registry_repository_iam_binding.authoritative["roles/artifactregistry.reader"] will be created
  + resource "google_artifact_registry_repository_iam_binding" "authoritative" {
      + etag       = (known after apply)
      + id         = (known after apply)
      + location   = "europe"
      + members    = [
          + "allUsers",
        ]
      + project    = "xxredactedxx"
      + repository = "public"
      + role       = "roles/artifactregistry.reader"
    }

  # module.artifact_registry.google_artifact_registry_repository_iam_binding.bindings["roles/artifactregistry.reader"] will be destroyed
  # (because key ["roles/artifactregistry.reader"] is not in for_each map)
  - resource "google_artifact_registry_repository_iam_binding" "bindings" {
      - etag       = "xxredactedxx" -> null
      - id         = "xxredactedxx" -> null
      - location   = "europe" -> null
      - members    = [
          - "allUsers",
        ] -> null
      - project    = "xxredactedxx" -> null
      - repository = "projects/xxredactedxx/locations/europe/repositories/public" -> null
      - role       = "roles/artifactregistry.reader" -> null
    }

```

---
**Checklist**

I applicable, I acknowledge that I have:
- [x] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [x] Ran `terraform fmt` on all modified files
- [x] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [x] Made sure all relevant tests pass

PR #2617, opened 2024-10-11
fix(artifact-registry): fix a move issue with tf>1.7

NitriKx:
Happy to be the first one to complain 🙂 

As mentionned by @juliocc in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2606 the move is failing if we transfer the resource to the non-beta provider:

```
│ The "google_artifact_registry_repository_iam_binding" resource type does
│ not support moving resource state across resource types.
```

The TF 1.8 has introduced a new feature that makes the move failing (I was testing with 1.7 in #2606):
> Providers can now transfer the ownership of a remote object between resources of different types, for situations where there are two different resource types that represent the same remote object type.
> This extends the moved block behavior to support moving between two resources of different types only if the provider for the target resource type declares that it can convert from the source resource type. Refer to provider documentation for details on which pairs of resource types are supported.
https://github.com/hashicorp/terraform/blob/v1.8/CHANGELOG.md#180-april-10-2024

---
**Checklist**

### DIFF
--- a/modules/artifact-registry/iam.tf
+++ b/modules/artifact-registry/iam.tf
@@ -20,6 +20,7 @@ moved {
 }
 
 resource "google_artifact_registry_repository_iam_binding" "authoritative" {
+  provider   = google-beta
   for_each   = local.iam
   project    = var.project_id
   location   = google_artifact_registry_repository.registry.location


juliocc (approved):
I won't say "I told you" ;)

PR #2627, opened 2024-10-16
added log_sinks to tfvars.logging in 0-bootstrap

lnesteroff:
Currently [var.logging 1-resman](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/47e74a1c52c53405dc067ede2bbfcf39337c3f4a/fast/stages/1-resman/variables-fast.tf#L98-L109) has the [logging structure from 1-tenants](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/47e74a1c52c53405dc067ede2bbfcf39337c3f4a/fast/stages/1-tenant-factory/outputs.tf#L99-L120) which includes log_sinks. This change adds var.log_sinks from 0-bootstrap in to local.tfvars.logging to align with 1-resman's expectations. Without this value, this causes issues for using state/tfe_outputs backends but not using the tenant-factory as 1-resman expects logging.log_sink to exist. 

---
**Checklist**

### DIFF
--- a/fast/stages/0-bootstrap/outputs.tf
+++ b/fast/stages/0-bootstrap/outputs.tf
@@ -112,6 +112,7 @@ locals {
     }
     custom_roles = module.organization.custom_role_id
     logging = {
+      log_sinks         = var.log_sinks
       project_id        = module.log-export-project.project_id
       project_number    = module.log-export-project.number
       writer_identities = module.organization.sink_writer_identities
--- a/fast/stages/0-bootstrap/README.md
+++ b/fast/stages/0-bootstrap/README.md
@@ -678,16 +678,16 @@ The remaining configuration is manual, as it regards the repositories themselves
 
 | name | description | sensitive | consumers |
 |---|---|:---:|---|
-| [automation](outputs.tf#L146) | Automation resources. |  |  |
-| [billing_dataset](outputs.tf#L151) | BigQuery dataset prepared for billing export. |  |  |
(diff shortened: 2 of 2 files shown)


fast/stages/0-bootstrap/outputs.tf
```diff
@@ -112,6 +112,7 @@ locals {
     }
     custom_roles = module.organization.custom_role_id
     logging = {
+      log_sinks         = var.log_sinks
```
ludoo, line 115: should we pass this on via globals output file instead? it's a static quantity, I'm a bit wary of breaking the model and adding it to the stage outputs

ludoo: ok, just read the PR description snd it would not make sense :)

but I don't understand the use case, the variable type in stage 1 is this

```hcl
variable "logging" {
  # tfdoc:variable:source 1-tenant-factory
  description = "Logging configuration for tenants."
  type = object({
    project_id = string
    log_sinks = optional(map(object({
      filter = string
      type   = string
    })), {})
  })
  nullable = false
}
```

and everything is optional, how can it cause issues in TFE?

ludoo: I am sure there was a good reason, but I am struggling to remember why we create tenant log sinks in resman. I would try and revisit that (so we can remove log_sinks from the tenants variable) before adding sinks for non-tenant use.

ludoo:
closing this, feel free to reopen or log an issue if you still need it

PR #2628, opened 2024-10-16
WIP: Custom env names

lnesteroff:
Working to use the var.environments from 0-bootstrap (that is included in tfvars_globals) to allow the use of custom environment namings (and additional envs as required). This would allow an org to use something like nonprod instead of dev for example. I've switched the syntax of "is_default" for the environment objects to primary/secondary which would map to prod/dev (in the default vars). I will then use these envs in the other stages where things like prod or dev are currently hard coded.

---
**Checklist**


ludoo:
Closing this, let's follow up on the discussion we had this week and open an ADR for this, with a plan on how to make envs dynamic in FAST.

PR #2639, opened 2024-10-28
Add option to attach multiple snapshot schedule to disks

shujaatsscripts:
Currently only one snapshot schedule is allowed  to attach to the disk. The variable take a string.

This PR added the option to give the list of snapshot schedules to attach to the disk/boot disk. This option is not allowed through GCP console but its available in terraform and gcloud cli. 

Example: If we need to take snapshots both daily and weekly, we need to attach 2 snapshot schedule to one disk.

---
**Checklist**


modules/compute-vm/resource-policies.tf
```diff
@@ -118,22 +141,25 @@ resource "google_compute_resource_policy" "snapshot" {
 }
 
 resource "google_compute_disk_resource_policy_attachment" "boot" {
-  count   = var.boot_disk.snapshot_schedule != null ? 1 : 0
-  project = var.project_id
-  zone    = var.zone
+  for_each = var.boot_disk.snapshot_schedule != null ? {
+    for i, v in var.boot_disk.snapshot_schedule : i => v
```
ludoo, line 145: nit: can you just use the variable instead of a `for` loop that returns the same?

ludoo (approved):
Thanks for this, very clean PR

PR #2654, opened 2024-11-01
added svpc subnet controls for project factory

lnesteroff:
Adding option in project factory to define subnet level IAM for shared vpc.

---
**Checklist**


modules/project-factory/main.tf
```diff
@@ -122,7 +122,10 @@ module "projects" {
         for v in try(each.value.shared_vpc_service_config.network_users, []) :
         lookup(local.context.iam_principals, v, v)
       ]
-      # TODO: network subnet users
+      network_subnet_users = {
```
ludoo, line 125: we should add subnets to `factories_config.contexts` so we can try a substitution here

lnesteroff: Do you mean something along these lines:
```HCL
factories_config = {
  context = {
    network_subnet_users = {
      default = {
        "region/subnet" = ["gcp-devops"]
      }
    }
  }
}
```
So that project factories could attach a user/group to a default subnet? If the expectation is that that group/user should always have access to that subnet, it would only need to be applied once.

I also just added a service_subnet_iam_grants, which is similar to service_iam_grants and does the permissions based on modules/projects/sharedvpc-agent-iam.yaml, but applied the networkUser roles at the subnet level. I think that might make sense to have as a default, so if an org wants to allow compute.googleapis.com to be enabled on a default subnet on all projects it would add the agent.

juliocc:
where are we with this?

ludoo:
I'm closing this. The project factory has evolved quite a bit in the meantime, and this would be really hard to merge.

PR #2656, opened 2024-11-01
Fix permadiff in bootstrap IAM

ludoo:
Role was managed by both stages 0 and 1, it should only be managed by 1.


wiktorn:
Shouldn't this target master branch (backward compatible bugfix), as this bug is also present in recent master?

ludoo:
> Shouldn't this target master branch (backward compatible bugfix), as this bug is also present in recent master?

Well we will merge fast-dev soon anyway, so yes but it changes very little 😁

wiktorn (approved):

PR #2661, opened 2024-11-05
Allow duplicate file names in project hierarchy

robrankin:
The Project Factory module doesn't allow duplicate filenames in the project hierarchy folder structure, as the filenames are currently used as keys in TF maps, and duplicates lead to duplicate key errors.

This change moves to building the map keys from either:

* The full path + filename with the separator of `/` replaced with `-` (in the abscence of an explicit `name:` key in the yaml)
* The `name:` key explicitly set in the project yaml definition file.

I've included two example project files, both named `ce-0.yaml`, but in different parts of the project hierarchy.

The first file, `team-a/dev/ce-0.yaml` has an explicitly set `name:` key, and produces the following resource:
```
  # module.projects.module.projects["ta-d-ce-0"].google_project.project[0] will be created
  + resource "google_project" "project" {
      + name                = "main-ta-d-ce-0"
      + project_id          = "main-ta-d-ce-0"
    }
```

The second file, `team-b/dev/ce-0.yaml` does not have an explicitly set `name:` key, and produces the following resource with the project name, ID and TF map key built from the path + filename:
```
  # module.projects.module.projects["team-b-dev-ce-0"].google_project.project[0] will be created
  + resource "google_project" "project" {
      + name                = "main-team-b-dev-ce-0"
      + project_id          = "main-team-b-dev-ce-0"
    }
```

---
**Checklist**


juliocc:
Not a bad idea but if we do this for PF we should probably do it for all other factories too. 

I actually like the idea and the fix is fairly simple.

@ludoo wdyt?

modules/project-factory/factory-projects.tf
```diff
@@ -20,7 +20,7 @@ locals {
   _hierarchy_projects = (
     {
       for f in try(fileset(local._folders_path, "**/*.yaml"), []) :
-      basename(trimsuffix(f, ".yaml")) => merge(
+      replace(trimsuffix(f, ".yaml"), "/", "-") => merge(
```
juliocc, line 23: do you really need the replace? I think you could keep '/' in the keys.

juliocc: Well yes if you want to use the filename as project id. 

Ignore

ludoo: Same comment I made above, path should be left intact

ludoo (commented):
Forgot to send comments lol

modules/project-factory/factory-projects.tf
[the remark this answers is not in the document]
juliocc: the problem (I think) is that '/' is invalid in project ids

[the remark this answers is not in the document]
ludoo: We need to keep the oath for keys, not project ids

robrankin:
Anything I can do here?  Happy to make changes if needed.

ludoo:
> Anything I can do here? Happy to make changes if needed.

Can you implement my suggestion above? Just remove basename and trim the suffix, and then we see how it works :)

ludoo:
Rob I'm closing this, let's reopen a discussion to flesh out what we want to do here.

Issue #2675, opened 2024-11-11
(modules/vpc-sc) Inconsistent conditional result types


joelvoss:
**Describe the bug**
When defining ingress/egress policies using the `ingress_policies`/`egress_policies` variable of the VPC-SC module and by using the factory (defining a policy by creating a `data/ingress-policies/my-policy.yaml` file), Terraform throws an "Inconsistent conditional result types" error if one policy defines the `identities` parameter and the other one the `indentity_type` parameter.

**Environment**
```
# terraform -version
Terraform v1.9.8
on darwin_arm64
```

```
# git rev-parse --short HEAD
b85711e8
```

**To Reproduce**
To reproduce the error, you can modify one of the examples from the VPC-SC readme.

1. Prepare an `access-levels` and `ingress-policies` factory configuration like so:

`data/access-levels/geo-it.yaml`
```yaml
conditions:
  - regions:
      - IT
```

`data/ingress-policies/i2.yaml`
```yaml
from:
  access_levels:
    - geo-it
  identity_type: ANY_IDENTITY
to:
  operations:
    - service_name: "*"
  resources:
    - projects/1234567890
```

2. Next, use the VPC-SC module like so:

```hcl
module "repro" {
  source        = "./fabric/modules/vpc-sc"
  access_policy = "12345678"
  factories_config = {
    access_levels    = "data/access-levels"
    egress_policies  = "data/egress-policies"
    ingress_policies = "data/ingress-policies"
  }
  ingress_policies = {
    i1 = {
      from = {
        identities = [
          "serviceAccount:sa-0@myproject.iam.gserviceaccount.com",
          "serviceAccount:sa-1@myproject.iam.gserviceaccount.com"
        ]
        access_levels = ["*"]
      }
      to = {
        operations = [{ service_name = "*" }]
        resources = ["*"]
      }
    }
  }
  service_perimeters_regular = {
    default = {
      status = {
        access_levels       = ["geo-it"]
        resources           = ["projects/11111", "projects/111111"]
        restricted_services = ["storage.googleapis.com"]
        egress_policies     = []
        ingress_policies    = ["i1", "i2"]
        vpc_accessible_services = {
          allowed_services   = ["storage.googleapis.com"]
          enable_restriction = true
        }
      }
    }
  }
}
```

**Expected behavior**
No terraform errors should be thrown when running `terraform plan` or `terraform apply`.

**Result**
```
Error: Inconsistent conditional result types

[...]

PR #2676, opened 2024-11-11
Fix "inconsistent conditional result types" error in `modules/vpc-sc`

joelvoss:
This PR fixes the issue described in issue #2675. I've also adapted one of the example based tests so that the incorrect behavior is checked in the future.

---
**Checklist**

### DIFF
--- a/modules/vpc-sc/README.md
+++ b/modules/vpc-sc/README.md
@@ -229,6 +229,24 @@ module "test" {
     egress_policies  = "data/egress-policies"
     ingress_policies = "data/ingress-policies"
   }
+  ingress_policies = {
+    sa-tf-test = {
+      from = {
+        identities = [
+          "serviceAccount:test-tf-0@myproject.iam.gserviceaccount.com",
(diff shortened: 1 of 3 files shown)


ludoo:
Thanks, I just wonder why you edit the ingress policy examples but the change to the factory code is for egress. :)

joelvoss:
I thought it would be enough to just customize the ingress part of the example-based test. I can also extend the egress part as well if you like.

The adjustments in the code affect both ingress and egress.

ludoo:
Thanks for clarifying. My preference would be to leave the current examples as is, and add an extra set of YAML/code rules to cover this case. It makes it easier to assess changes.

joelvoss:
> Thanks for clarifying. My preference would be to leave the current examples as is, and add an extra set of YAML/code rules to cover this case. It makes it easier to assess changes.

No problem. I have reverted the changes to the current examples and added an extra set of YAML/code rules to cover this case instead.

ludoo (approved):
thanks a lot for this!

PR #2689, opened 2024-11-14
add GPU options to comute-vm module

ooshrioo:
---
**Checklist**


modules/compute-vm/variables.tf
```diff
@@ -392,3 +392,12 @@ variable "zone" {
   description = "Compute zone."
   type        = string
 }
+
+variable "gpu" {
+  description = "GPU information."
+  type = object({
+    count = number
+    type  = string
```
ludoo, line 400: we typically treat strings that only accept a fixed set of values as enums, and enforce the possible values via avriable validation; check the DNS module for one of many examples

modules/compute-vm/main.tf
```diff
@@ -333,6 +334,14 @@ resource "google_compute_instance" "default" {
       resource_manager_tags = local.tags_combined
     }
   }
+
+  dynamic "guest_accelerator" {
```
ludoo, line 338: is this also supported for instance templates? if so the same block needs to be added to that resource

ludoo (commented):
Hey, glad you made it :) Dropped a couple comments, feel free to ping me if you need clarifications

ludoo:
Oshri, can you sign our CLA?

ludoo (approved):

PR #2693, opened 2024-11-15
Unify usage of top level folders short_name

juliocc:
1. Make `short_name` optional (use key name/filename if not provided)
2. Disable automation by default
3. Remove enable key from top_level_folders automation field

The idea here is to have make the following work
```
top_level_folders = {
  test = {
    name       = "test folder"
  }
}
```
Today this would enable automation, and fail because we're not providing a short name. With this change, the code above works and does **not** enable automation. IMO this makes more sense since the user is not mentioning automation at all


ludoo (approved):
very nice, can you also edit the schema?

ludoo:
I'm also curious about which IAM bindings we're dropping in tests with this PR.

ludoo (approved):

PR #2696, opened 2024-11-15
Add support in GKE for DNS endpoint

rosmo:
Add support in GKE for: control_plane_endpoints_config, gcp_public_cidrs_access, private_endpoint_enforcement

---
**Checklist**


modules/gke-cluster-autopilot/variables.tf
```diff
@@ -35,6 +35,16 @@ variable "backup_configs" {
...
+variable "control_plane_endpoints_config" {
+  description = "Configuration for all of the cluster's control plane endpoints."
+  type = object({
+    dns_endpoint_config = optional(object({
+      allow_external_traffic = optional(bool)
+    }))
+  })
+  default = null
+}
+
```
juliocc, line 47: I know this aligns with the API does, but I don't like the type of this variable. I'm also concerned that now we have endpoint configuration spread across multiple variables.

I really like what the UI is doing now:
image

Could we consider extending this variable to include all control plane endpoint configuration options (DNS, PSC, and PSA)?

Something like
```terraform
variable "control_plane_endpoints_config" {
  description = "Configuration for all of the cluster's control plane endpoints."
  type = object({
    dns_endpoint = optional(object({...}))
    psc_endpoint = optional(object({...}))
    private_endpoint = optional(object({...}))
  })
}
```

I realize this is probably a larger change than what you had in mind. What do you think?

juliocc (changes requested):

modules/gke-cluster-autopilot/variables.tf
rosmo: I think it would be fine, though it makes a bit wary to differ a lot from the underlying Terraform resource (though granted, it has become a bit messy over the years). But it could look something like this:

```hcl
variable "control_plane_endpoints_config" {
  description = "Configuration for all of the cluster's control plane endpoints."
  type = object({
    dns_endpoint = optional(object({
     allow_external_traffic = optional(bool) 
   }))
    psc_endpoint = optional(object({
      subnetwork = optional(string) # maps to vpc_config.master_endpoint_subnetwork
      # enabling PSC implies no master_ipv4_cidr_block, peering_configs etc
      master_global_access = optional(bool) #  maps to private_cluster_config.master_global_access, I think this is only for PSC?
   }))
    private_endpoint = optional(object({ # Maps to private_cluster_config.enable_private_endpoint 
     # put vpc_config.master_ipv4_cidr_block here too ?
     # and perhaps private_cluster_config.peering_config
   }))
   master_authorized_ranges = optional(map(string)) # maps to vpc_config.master_authorized_networks, top-level because it applies to public/private/PSC
  })
}
```

If we don't need backwards compatibility, it would be fairly straightforward I suppose.

ludoo:
This has too many conflicts and should be redone via a module interface change. Will open an issue to track work on this.

PR #2697, opened 2024-11-15
Remove REGIONAL/MULTI_REGIONAL buckets from FAST

juliocc:


ludoo (approved):
as discussed, maybe the should default to `null` in the GCS module

juliocc:
I'll address the GCS change in another PR

PR #2484, opened 2024-08-08
[FAST] TLS inspection support for NGFW Enterprise

LucaPrete:
Adds TLS inspection support for NGFW Enterprise.

* Creates in stage 2-security: CA configurations (CAS), Certificate Manager trust-configs, TLS inspection policies
* References in stage 3-network-security the TLS inspection policies created in stage-2 security from the firewall endpoint associations

---
**Checklist**

I applicable, I acknowledge that I have:
- [X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X] Ran `terraform fmt` on all modified files
- [X] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [x] Made sure all relevant tests pass

PR #2701, opened 2024-11-16
Allow disabling network security stage

juliocc:
This PR:
- Removes some incorrect bindings that were failing when the security stage is disabled.
- Some potential errors being masked by `try()`s.
- Binding to an non-existing `network_firewall_policies_viewer` custom role.


juliocc:
The test failure is expected. We where previously assigning the network_firewall_policies_admin role always, but it's only needed if network_security is enabled.

So we either extend the test to enable network_security or we keep it as is and just remove that binding from the inventory.

fast/stages/1-resman/stage-2-networking.tf
```diff
@@ -60,35 +60,29 @@ module "net-folder" {
     # network security stage 2 service accounts
     var.fast_stage_2.network_security.enabled != true ? {} : {
       "roles/serviceusage.serviceUsageAdmin" = [
-        try(module.nsec-sa-rw[0].iam_email, null)
+        module.nsec-sa-rw[0].iam_email
```
ludoo, line 63: nsec is network security, which can be enabled/disabled independently from security; the trys are needed here...

fast/stages/1-resman/stage-2-networking.tf
```diff
@@ -60,35 +60,29 @@ module "net-folder" {
...
       ]
       "roles/compute.orgFirewallPolicyUser" = [
-        try(module.nsec-sa-ro[0].iam_email, null)
+        module.nsec-sa-ro[0].iam_email
       ]
       "roles/serviceusage.serviceUsageConsumer" = [
-        try(module.nsec-sa-ro[0].iam_email, null)
+        module.nsec-sa-ro[0].iam_email
       ]
+      # FIXME: this role does not exist. And a try was hiding the error
```
ludoo, line 74: I think this is the correct role? the nextsec stage was pretty liberal in IAM handling, I sorted a few things but some might have escaped...

https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/fast-dev/fast/stages/0-bootstrap/data/custom-roles/ngfw_enterprise_viewer.yaml

ludoo (commented):
see comments, nsec != security

fast/stages/1-resman/stage-2-networking.tf
juliocc: I know they're different things, and I think the roles/service accounts are mixed up. Looking at the security stage, I don't see anything related to firewall policies. `nsec`, on the other hand, does create firewall policies.

I don't see why the `try`s are needed. The SAs are supposed to exist if their stage is enabled, and the conditional around them guarantees the stage is enabled. Am I missing something?

In any case, at the very least, all these `try`s need a `compact` around the outer list; otherwise, you might end up with null elements, which makes IAM throw errors.

fast/stages/1-resman/stage-2-networking.tf
juliocc: Based on this [PR](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/2484/files#diff-b8235172e37c0123df773c6f33983c692d1f46ecaf9958e510eabdf33ec9ae6aL41-L47) I think `network_firewall_policies_viewer` was actually replaced by `roles/compute.orgFirewallPolicyUser` which is already granted above.

fast/stages/1-resman/stage-2-networking.tf
ludoo: Well, look at the current master. I just ported what is there.

juliocc: In master it's also only granted when nsec is enabled:

https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/6b0745deed68b63ca0d69c891cb4141768a94a7e/fast/stages/1-resman/branch-networking.tf#L33-L48

ludoo: Try serves same purpose, on mobile will send proper comments tmw 😂

juliocc: no hurry, I'm just curious...

ludoo: ok I misread the PR (mobile, teaches me do detach when I'm not at my coputer)...

- the condition is correct and trys not needed
- the firewall policy admin role *is* needed as in Luca's design (which we will change eventually) the nsec stage manages fw policies and needs to attach them to the VPCs defined here
- the nsec ro SA was missing a role, and I'm pretty sure I added it but you're right, it's not in bootstrap so either we a) add it, or b) use a looser but still view only predefined role

fast/stages/1-resman/stage-2-networking.tf
```diff
@@ -60,35 +60,25 @@ module "net-folder" {
...
     var.fast_stage_2.security.enabled != true ? {} : {
       "roles/serviceusage.serviceUsageAdmin" = [
-        try(module.sec-sa-rw[0].iam_email, null)
+        module.sec-sa-rw[0].iam_email
       ]
       "roles/serviceusage.serviceUsageConsumer" = [
-        try(module.sec-sa-ro[0].iam_email, null)
-      ]
-    },
-    try(var.custom_roles["network_firewall_policies_admin"], null) == null ? {} : {
```
ludoo, line 84: as mentioned above, these are needed until we refactor the nsec stage and keep fw policies in the net stages, they are currently managed from nsec and attached to VPCs here

juliocc: Agreed, but the one I removed here is granting to the **security SA** (not nsec SA). 

The correct binding (to the nsec SA) is already done a few lines above together with all the other nsec bindings

fast/stages/1-resman/stage-2-networking.tf
juliocc: > the condition is correct and trys not needed

Awesome :)

> the firewall policy admin role is needed as in Luca's design (which we will change eventually) the nsec stage manages fw policies and needs to attach them to the VPCs defined here

Correct, and it's already there.

> the nsec ro SA was missing a role, and I'm pretty sure I added it but you're right, it's not in bootstrap so either we a) add it, or b) use a looser but still view only predefined role

Also correct. My impression is that Luca replaced it with a `roles/compute.orgFirewallPolicyUse`, which is also already granted.

ludoo (approved):
thanks for being patient lol

PR #2707, opened 2024-11-17
Expose ` factories_config` for resman top level folders

juliocc:
This PR: 
 - allows top level folders to use factories,
 - enables sandbox policies with the policies in data/org-policies/sandbox


ludoo (approved):
Inception 😂

PR #2715, opened 2024-11-18
Allow setting GCS location default/override in project factory

ludoo:
This allows customers with enforcement on locations to default or override the location for automation buckets.


juliocc:
Does this need a schema update somewhere?

juliocc (approved):

ludoo:
> Does this need a schema update somewhere?

Ah God I'm sloppy, yes

ludoo:
correction: no, it's just an extra default/override, not a schema change

PR #2724, opened 2024-11-20
Move stage 3 role creation to resman

juliocc:


fast/stages/1-resman/stage-2-networking.tf
```diff
@@ -83,10 +83,10 @@ module "net-folder" {
     },
     # project factory service accounts
     (var.fast_stage_2.project_factory.enabled) != true ? {} : {
-      (var.custom_roles.service_project_network_admin) = [
+      (var.custom_roles["service_project_network_admin"]) = [
```
wiktorn, line 86:
```suggestion
      (local.custom_roles["service_project_network_admin"]) = [
```

fast/stages/1-resman/stage-2-networking.tf
```diff
@@ -83,10 +83,10 @@ module "net-folder" {
...
         module.pf-sa-rw[0].iam_email
       ]
-      (var.custom_roles.project_iam_viewer) = [
+      (var.custom_roles["project_iam_viewer"]) = [
```
wiktorn, line 89:
```suggestion
      (local.custom_roles["project_iam_viewer"]) = [
```

fast/stages/1-resman/moved/v35.1.0-v36.0.0.tf
```diff
@@ -236,3 +236,21 @@ moved {
...
+  for_each = var.fast_stage_2.network_security.enabled ? [""] : []
+  to       = module.nsec-organization-roles[0].google_organization_iam_custom_role.roles["ngfw_enterprise_admin"]
+  id       = "organizations/${var.organization.id}/roles/ngfwEnterpriseAdmin"
+}
+
+import {
+  for_each = var.fast_stage_2.network_security.enabled ? [""] : []
+  to       = module.nsec-organization-roles[0].google_organization_iam_custom_role.roles["ngfw_enterprise_viewer"]
+  id       = "organizations/${var.organization.id}/roles/ngfwEnterpriseViewer"
+}
```
wiktorn, line 256: The migration scenario then assumes following actions:
* run `terraform state rm ...` on 0-bootstrap - to remove roles from state and prevent their deletion on `apply`
* run `terraform apply` on 1-resman

I guess, to keep the migration paths clean, we need to call out that manual step somewhere.

wiktorn (commented):
Not functional, but just to save myself wondering why sometime we use `local.custom_roles` and sometimes `var.custom_roels` I'd align `custom_roles` access to go through `local` everywhere. There are still left some leftovers in `billing.tf` and `stage*.tf`.

fast/stages/1-resman/stage-3.tf
```diff
@@ -108,6 +111,22 @@ locals {
...
+  stage3_factories = [
+    for k, v in local.stage3 :
+    v.factories_config
+    if v.factories_config != null
+  ]
+}
+
+# Stage 3 custom role factories
+
+module "organization-roles" {
```
ludoo, line 123: do we really need extra module calls for this? can't we roll up modules in the existing organization module?

ludoo (approved):
very nice, I would just try reusing the current organization module invocation and bundle up all roles deriving from stage 3 factories in an internal factory local

fast/stages/1-resman/moved/v35.1.0-v36.0.0.tf
juliocc: Correct. I tried adding a `removed` block  to bootstrap but couldn't figure it out.

PR #2746, opened 2024-12-05
Add Default API Alerts

joshw123:
---
**Checklist**

### DIFF
--- a/modules/project/api_metrics_alerts.tf
+++ b/modules/project/api_metrics_alerts.tf
@@ -1,0 +1,361 @@
+resource "google_monitoring_notification_channel" "email" {
+  count        = var.enable_default_api_alerts ? 1 : 0
+  project      = local.project.project_id
+  display_name = "Default Email Notification"
+  type         = "email"
+  labels = {
+    email_address = var.default_api_alerts_email
+  }
+}
+
+#
+# Route Changes Metric and Policy
+#
+resource "google_logging_metric" "route_changes" {
+  count       = var.enable_default_api_alerts ? 1 : 0
+  project     = local.project.project_id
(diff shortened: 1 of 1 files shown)


modules/project/README.md
```diff
@@ -1360,12 +1360,50 @@ module "bucket" {
...
+module "project" {
+  source          = "./fabric/modules/project"
+  billing_account = var.billing_account_id
+  name            = "project"
+  parent          = var.folder_id
+  prefix          = var.prefix
+  services = [
+    "stackdriver.googleapis.com"
+  ]
+  enable_api_alerts = true
```
ludoo, line 1396: can we combine both into a single `api_alerts` variable, with `.enabled` and `default_email` attributes?

ludoo (commented):
I'm not sure this should be in the project module: while this is useful, it's not a product feature but a "recipe" or collection of "recipes" which we typically never embed in modules. Plus, having them in code makes it very hard for users to change/tweak them.

At a minimum, recipes should be defined via YAML files in a factory. But I still don't like them in the base project module. What if we added support for it in the project factory, which is where you would typically enable/disable alerts?

joshw123:
> I'm not sure this should be in the project module: while this is useful, it's not a product feature but a "recipe" or collection of "recipes" which we typically never embed in modules. Plus, having them in code makes it very hard for users to change/tweak them.
> 
> At a minimum, recipes should be defined via YAML files in a factory. But I still don't like them in the base project module. What if we added support for it in the project factory, which is where you would typically enable/disable alerts?

Hey, Yeah I can look at putting it within the project-factory, The only reason I didn't do that was with compliance checks, these alerts are required in every project, and project-factory is only used in stage-2 project factory I believe? @ludoo

ludoo:
Wait :) Let's leave this open and have @juliocc and @wiktorn also comment :)

ludoo:
reopening

juliocc:
Agree with @ludoo, this shouldn't be part of the project module.

PR #2756, opened 2024-12-09
Add Logging Alerts Factory

joshw123:
---
**Checklist**


joshw123:
@ludoo First Stab at the Factory :) Could you take a look please?

fast/stages/1-resman/data/logging-alerts/compliance-alerts.yaml
juliocc, line 1: I would keep a single copy of this file (perhaps the one in bootstrap) and symlink it from other stages.

modules/project/variables.tf
```diff
@@ -78,12 +78,19 @@ variable "descriptive_name" {
...
 variable "factories_config" {
   description = "Paths to data files and folders that enable factory functionality."
   type = object({
-    custom_roles = optional(string)
-    org_policies = optional(string)
-    quotas       = optional(string)
+    custom_roles           = optional(string)
+    org_policies           = optional(string)
+    quotas                 = optional(string)
+    logging_metrics_alerts = optional(string)
```
juliocc, line 93: I'd prefer to have this split in 3 separate factories: `metrics`, `alerts`, `notification_channels`

modules/project/alert-metrics.tf
```diff
@@ -0,0 +1,87 @@
...
+resource "google_monitoring_notification_channel" "default" {
+  project      = local.project.project_id
+  display_name = "Default Email Notification"
+  type         = "email"
+  labels = {
+    email_address = try(var.default_alerts_email)
+  }
+}
+
+resource "google_logging_metric" "default" {
```
juliocc, line 49: This only allows the creation of the simplest metrics. The resource (`google_logging_metric`) provides much more flexibility which should be expose by the factory.

modules/project/alert-metrics.tf
```diff
@@ -0,0 +1,87 @@
...
+  filter      = each.value.filter
+  name        = each.value.name
+  description = each.value.description
+  metric_descriptor {
+    metric_kind = each.value.metric_descriptor.metric_kind
+    value_type  = each.value.metric_descriptor.value_type
+  }
+}
+
+resource "google_monitoring_alert_policy" "default" {
```
juliocc, line 61: Same as the `google_logging_metric`, this is assuming a given policy structure.

juliocc (changes requested):
Hey @joshw123, thanks for getting started with this.

I left a few specific comments below, but my overall proposals are:

1. **Decouple metrics and alerts:** Currently, metrics and alert policies are tightly coupled, assuming all alerts originate from custom log-based metrics. I'd prefer a more general approach where alerts can be created independently of these metrics, or even use  built-in metrics.
2. **Increase factory flexibility:** The factory exposes only a few fields from the underlying resources. We should allow more flexibility and user control over these resources.
3. **Enable user control over notifications:** We should allow users to manage and configure notification channels according to their preferences.

I think all 3 points can be addressed by having 3 factories instead of one: one for metrics, one for alerts, and one for notification channels

joshw123:
> Hey @joshw123, thanks for getting started with this.
> 
> I left a few specific comments below, but my overall proposals are:
> 
> 1. **Decouple metrics and alerts:** Currently, metrics and alert policies are tightly coupled, assuming all alerts originate from custom log-based metrics. I'd prefer a more general approach where alerts can be created independently of these metrics, or even use  built-in metrics.
> 2. **Increase factory flexibility:** The factory exposes only a few fields from the underlying resources. We should allow more flexibility and user control over these resources.
> 3. **Enable user control over notifications:** We should allow users to manage and configure notification channels according to their preferences.
> 
> I think all 3 points can be addressed by having 3 factories instead of one: one for metrics, one for alerts, and one for notification channels

Thank You for the feedback @juliocc In terms of Data, should I aim for `data/observability/alerts, data/observability/channels, data/observability/metrics` just don't want to convolute the data with more folders than needed? thank you!

juliocc:
> Thank You for the feedback @juliocc In terms of Data, should I aim for `data/observability/alerts, data/observability/channels, data/observability/metrics` just don't want to convolute the data with more folders than needed? thank you!

That sounds good. I'd even remove the `observability` folder and just drop the new folders inside `data/`

PR #2761, opened 2024-12-11
Refactor GKE cluster modules access configurations, add support for DNS endpoint

### LINKED ISSUE #2753: Add support for GKFE to gke modules.
Resume work started in #2696

ludoo:
Fixes #2753


juliocc (approved):

apekurovsky:
Is it possible to make new release for cloud-foundation-fabric? We want to use new GKE DNS endpoint, however we don't want to pin our TF to master, we would like to pin to tag with release.

ludoo:
Sure. In the meantime, you could use one of our daily tags.

PR #2762, opened 2024-12-12
Update default-versions.tf

JohnbelMDev:
Regularly audit and update provider versions for security patches. Use a secure backend with encryption for the Terraform state. Validate input variables and restrict provider permissions. Avoid exposing unnecessary metadata (like release notes) unless necessary for internal documentation.

---
**Checklist**


ludoo:
I think we're fine with our current approach. @juliocc @wiktorn WDYT?

wiktorn:
@JohnbelMDev I think that better approach to achieve what you're looking for is to use [.terraform.lock.hcl](https://developer.hashicorp.com/terraform/language/files/dependency-lock).

`versions.tf` file say about the compatibility - so what we do expect to work with the given code. While users may mix modules from different sources in one code they need to rely on more relaxed versioning to allow module upgrades. Otherwise all modules that they use has to agree on using exactly the same version of the provider which makes then upgrades close to impossible.

juliocc:
Actually I think this is a good idea. For one, it would remove the need for [pinning versions before running tests](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/.github/actions/fabric-tests/action.yml#L71). It is definitely more conservative than our current approach but I think it would prevent some issues we've seen in the past.

I'd do it only for the provider, leaving the Terraform version as we currently have it.

ludoo:
Well, decide between you two @juliocc and @wiktorn  :)

default-versions.tf
```diff
@@ -12,18 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Fabric release: v36.0.1
```
ludoo, line 15: Please do not remove this.

juliocc:
After a bit more discussion with @wiktorn we've decided not to go with this. Thanks

JohnbelMDev:
Thank you for your thoughtful response and guidance of insights for
providing relevant details . I will take into consideration of the
conservative to be more modern approach. Thank you indicating the approach
wasn't modern to modernize code standard of google that still getting
familiar to me.  Will be more global thinker for flexible for
situations like this of current. Terrafor.lock.hcl are usually there for
versions.


Best regards,


-Johnbel Mahautiere

PR #2787, opened 2024-12-27
Leverage environments for folder and project creation in FAST resman and security

ludoo:
This PR extends the recent work on resman and stages, so that environment-level resources are now driven by the `environments` interface variable, greatly reducing the need to change code when user environments do not match our default ones.

The networking stages have been left as is, as work is ongoing to refactor them and this kind of change will need to be considered after the refactor has been completed, so as not to complicate things.

The network security stage will also change its input interface, and be driven by variables.

If different environments are needed for a specific stage, tfvars overrides can be used to override the environments coming in via globals.


juliocc (approved):
Nice! cleaner than what I expected.

PR #2788, opened 2024-12-27
New blueprint for React & Python single page app

rosmo:
New blueprint for React & Python single page apps, both with global and regional load balancer. Includes automation for extracting Content-Security-Policy headers.


ludoo:
Hey Taneli, I don't think this is a good example to have here: not only because we're trying to move awayrationalize blueprints, but more importantly because we focus on Terraform. Anything else should at best have a supporting role, or not even be there.

@juliocc WDYT?

rosmo:
I have no particular opinion whether it is a good example or not, but 1) I've had multiple customers ask about this, 2) There is no documented pattern out there right now. Make of it what you will.

PR #2801, opened 2025-01-07
Add optional support for fw policies via new vpc_configs variable, refactor factories variable in net stages

ludoo:
This PR adds support for firewall policies to the `a` networking stage (the other stages will also be updated if the approach is agreed upon), via a `vpc_configs` variable.

The variable is also used to concentrate some existing functionality which was previously scattered among other variables, and allow flipping cloudnat, DNS, and firewall bits for individual VPCs where the same configuration was previously applied to all VPC unconditionally.

From the point of view of resources this does not change anything, but it allows leveraging firewall policies especially in connection with NGFW, controlling the firewall precedence between classic and policies, and turning off classic firewalls entirely if needed. I actually need this for a project I'm working on, and I think it's high time we started supporting firewall policies.


fast/stages/2-networking-a-simple/net-dev.tf
```diff
@@ -67,24 +83,26 @@ module "dev-spoke-project" {
...
-  source     = "../../../modules/net-vpc"
-  project_id = module.dev-spoke-project.project_id
-  name       = "dev-spoke-0"
-  mtu        = 1500
-  dns_policy = {
-    logging = var.dns.enable_logging
+  source                          = "../../../modules/net-vpc"
+  project_id                      = module.dev-spoke-project.project_id
+  name                            = "dev-spoke-0"
+  mtu                             = 1500
```
juliocc, line 89: move into vpc_configs?

juliocc (approved):

PR #2803, opened 2025-01-09
New tool `versions.py` to manage versions.tf/tofu 

juliocc:
...also adds a `provider_metadata` block in all modules

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,4 +1,4 @@
-# Copyright 2024 Google LLC
+# Copyright 2025 Google LLC
 #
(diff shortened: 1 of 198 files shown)


ludoo (approved):
Very nice.

One idea: could we have a mode where the tool parses versions from the default versions file? IMHO it's easier to get all major and minor versions right by looking at a file in an editor, rather than passing switches to a tool :)

juliocc:
> Very nice.
> 
> One idea: could we have a mode where the tool parses versions from the default versions file? IMHO it's easier to get all major and minor versions right by looking at a file in an editor, rather than passing switches to a tool :)

It already does exactly that :)

Issue #2805, opened 2025-01-09
[FR]: Introduce iam_additive_by_principal just like iam_by_principals but for additive bindings


tenminus11:
### The feature request

Introduce iam_additive_by_principal just like iam_by_principals but for additive bindings
Name iam_additive_by_principal can be different

### Proposed solution

Currently Iam Additive roles where giving multiple role to a common principal requires you to repeat the principal again which is hard to read
This can be replicated to across multiple modules
```
variable "iam_additive_by_principal" {
  description = "Additive IAM binding in {PRINCIPAL => [ROLES]} format."
  type        = map(list(string))
  default     = {}
  nullable    = false
}
```

juliocc:
> Introduce `iam_additive_by_principal` just like `iam_by_principals` but for additive bindings

This sounds like a nice idea. It'll probably be a bit tricky to implement. We could start with a single module and see how it goes.

Are you interested in this for any particular module?

> Name `iam_additive_by_principal` can be different

What do you mean with this?

tenminus11:
organization, project, folder modules in order is something I would like to get implemented first

iam_additive_by_principal variable name is upto the team to decide

PR #2814, opened 2025-01-13
Add `iam_by_principals_additive` to project, organization and folder modules

juliocc:
This pull request introduces the `iam_by_principals_additive` variable to the folder module, addressing the feature request outlined in #2805. 

This implementation merges the functionality of `iam_by_principals_additive` with the existing `iam_bindings_additive` logic, utilizing a single `google_folder_iam_member` resource. While the core implementation is straightforward, the validation logic requires further review and likely expansion. Currently, validation only prevents the same principal/role combination from being defined in both `iam_bindings_additive` and `iam_by_principals_additive`.

Open questions:
- should we maintain a single `google_folder_iam_member` resource for all additive bindings, or would a separate dedicated block for `iam_by_principals_additive` be preferable?
- what other validations, if any, are necessary to ensure correct usage. 
- I'm chose `iam_by_principals_additive` to be consistent with the existing `iam_bindings_additive` naming. 

Usage would be something like this:
```terraform
module "folder" {
  source              = "./fabric/modules/folder"
  parent              = var.folder_id
  name                = "folder"

  iam_by_principals_additive = {
    "user:xxx@example.com" = [
      "roles/logging.admin",
      "roles/logging.viewer"
    ]
  }
}
```

And this should fail:

```terraform
module "folder" {
  source              = "./fabric/modules/folder"
  parent              = var.folder_id
  name                = "folder"

  iam_bindings_additive = {
    "my-binding" = {
      role   = "roles/logging.admin"
      member = "user:xxx@example.com"
    }
  }

  iam_by_principals_additive = {
    "user:xxx@example.com" = [
      "roles/logging.admin",
      "roles/logging.viewer"
    ]
  }
}
```

Thoughts?


modules/folder/variables-iam.tf
```diff
@@ -51,6 +51,51 @@ variable "iam_bindings_additive" {
...
 
+variable "iam_by_principals_additive" {
+  description = "Additive IAM binding in {PRINCIPAL => [ROLES]} format. Principals need to be statically defined to avoid cycle errors."
+  type        = map(list(string))
+  default     = {}
+  nullable    = false
+
+  validation {
+    error_message = format(
+      "Cannot refer to the same principal/role pair from iam_by_principals_additive and iam_bindings_additive. The following pairs are present on both sides:\n%s",
```
ludoo, line 62: I would skip the format and just spit out a generic message: it should be pretty simple for the user to figure out the dupes. We never use context-aware validation messages, IMHO they add unnecessary complexity to what is essentially just a failsafe.

ludoo:
I like this, the implementation is simple and readable.

I'm just slightly concerned about the additional additive surface this introduces, which IMHO will just make it easier to create unintentional overlaps between authoritative and additive. But that's kind of inevitable and the new interface might actually be useful, so...

As for how many resources, my preference would be for a single one but I don't think it makes a lot of difference TBH, both should work.

ludoo (approved):

modules/folder/variables-iam.tf
juliocc: Actually, I'll just remove the validation

wiktorn:
> * what other validations, if any, are necessary to ensure correct usage.

If any, I'd add a check, if there is any overlap in roles between authoritative roles and additive ones, though in my experience, this was usually caused by IAM grants managed by separate modules.

adrs/modules/20230816-iam-refactor.md
```diff
@@ -202,6 +218,13 @@ variable "iam_bindings_additive" {
   nullable = false
 }
 
+variable "iam_by_principals_additive" {
+  description = "Additive IAM binding in {PRINCIPAL => [ROLES]} format. Principals need to be statically defined to avoid cycle errors. Merged internally with the `iam_bindings_additive` variable."
```
wiktorn, line 222: It's actually not a cycle errors, but "for_each will be known only after apply" error :-)

wiktorn (approved):

adrs/modules/20230816-iam-refactor.md
juliocc: Done. But we still have a lot of places where it shows up.

PR #2834, opened 2025-01-23
Allow multiple stage-2 project factories

juliocc:


fast/stages/1-resman/outputs-providers.tf
```diff
@@ -131,3 +132,7 @@ locals {
     },
   )
 }
+
+output "xxx" {
+  value = local.stage_addons
```
ludoo, line 137: remember to drop this :)

ludoo (approved):
I really like this. The only thing that bugs me is what we were discussing in chat: `project-factory-pf`.

Why don't we call our default project factory `teams`? So names would be

- `project-factory-teams` for files
- `-teams-0`, `teams-0r` etc for service accounts

fast/stages/1-resman/outputs-providers.tf
juliocc: done

PR #2878, opened 2025-02-11
Move DRS and essential contact domains to factory

juliocc:
*Upgrade note*:

```upgrade-note
fast/stages/0-boostrap: if you use `var.org_policies_config.constraints.allowed_policy_member_domains` or `var.org_policies_config.constraints.allowed_policy_member_domains`, move their values to a YAML file under bootstrap's org policy factory.
```


ludoo (approved):
This is awesome. Do we need to also change something in the docs?

Issue #2877, opened 2025-02-11
[FR]: Extend Load Balancers to support Cloud Armor security policy


juliocc:
### The feature request

Currently the load balancing modules don't support [Cloud Armor Security Policies](https://cloud.google.com/armor/docs/security-policy-overview). Extend modules to support security policies

### Additional context

[Related TODOs](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/v37.1.0/modules/net-lb-app-ext-regional/backend-service.tf#L39)
Terraform fields: [Regional](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_region_backend_service#security_policy-1) [Global](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_backend_service#security_policy-2)

b/395800661

ludoo:
Most LBs support it already, This is currently missing in the external app regional.

Issue #2866, opened 2025-02-04
[FR]: net-lb-app-ext-regional to support security policies and adjustable network tier


tardigrde:
### The feature request

`net-lb-app-ext-regional` to support:
1. `security_policy` field (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_region_backend_service#security_policy-1)
2. `network_tier` variable and attribute to be set to PREMIUM (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_forwarding_rule#network_tier-1) 

### Proposed solution

```
module "glb-0" {
  source     = "./fabric/modules/net-lb-app-ext-regional"
  project_id = var.project_id
  name       = "ralb-test-0"
  vpc        = var.vpc.self_link
  region     = var.region
  # 2nd part
  network_tier = "PREMIUM"
  backend_service_configs = {
    default = {
      # 1st part
      security_policy = "...."
      backends = [
        { backend = module.compute-vm-group.group.id }
      ]
    }
  }
}
```

### Additional context

For the 1st part, I udnerstand that it's still in beta, but the backend service resource is anyway uses the beta provider. Without the security policy, the backend service can not be properly secured.

For the 2nd part, I noticed a comment `external regional app lb only supports standard tier`. In my expereiences and also based on the provider documentation, this is not true. Please support the setting of the tier field (otherwise the external address must be also in STANDARD tier).

Issue #2865, opened 2025-02-03
[FR]: dual-stack (ipv4 + ipv6) on net-lb-app-ext module


matheusbragatti:
### The feature request

As of now there is no way to enable dual-stack on net-lb-app-ext it currently only accepts ipv4.

### Proposed solution

Enable multiple forwarding rules with multiple configurations.
The module net-lb-ext has a similar concept:

```
module "nlb" {
  source     = "./fabric/modules/net-lb-ext"
  project_id = var.project_id
  region     = var.region
  name       = "nlb-test"
  backends = [{
    group = module.nlb.groups.my-group.self_link
  }]
  forwarding_rules_config = {
    ipv4 = {
      version = "IPV4"
    }
    ipv6 = {
      version = "IPV6"
    }
  }
  group_configs = {
    my-group = {
      zone = "${var.region}-b"
      instances = [
        module.compute-vm-group-b.id,
      ]
    }
  }
}
```

PR #2879, opened 2025-02-11
Address outstanding load balancer FRs

ludoo:
This fixes several issues currently open on load balancers:

  - #2877
  - #2866
  - #2865

The approach followed for #2865 uses the same interface already in place for multiple forwarding rules in the `net-lb-int` module. The `""` name for the default forwarding rule allows having short names and being backward compatible, and has also been in use in the `net-lb-int` module for more than a year.


modules/net-lb-app-ext-regional/variables.tf
```diff
@@ -142,6 +142,13 @@ variable "neg_configs" {
   }
 }
 
+variable "network_tier_standard" {
+  description = "Use standard network tier."
+  type        = bool
+  nullable    = false
+  default     = true
```
wiktorn, line 149: Should this default to false? Provider assumes [PREMIUM](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_forwarding_rule#network_tier-1)

And there is also [default network tier](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_project_default_network_tier) which will be incompatible with this, but explicit is better than implicit :-)

ludoo: THis is the default we had hardcoded, so the we are keeping compatibility. It also makes sense to use the lower tier for regional.

image

ludoo: And there's only two supported tiers, so this needs to be a `bool`. :)

PR #2880, opened 2025-02-11
fix(iam): use managed constraints instead of predefined ones

Ameausoone:
---
**Checklist**

### DIFF
--- a/fast/stages/0-bootstrap/data/org-policies-iac/iam.yaml
+++ b/fast/stages/0-bootstrap/data/org-policies-iac/iam.yaml
@@ -18,7 +18,6 @@ iam.automaticIamGrantsForDefaultServiceAccounts:
   rules:
     - enforce: true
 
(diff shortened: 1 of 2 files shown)


ludoo:
We can't do this as new organizations still use the old constraints, and switching to the new ones will make it a lot harder to support new users. We will need to postpone until the new constraints are adopted GCP-wide.

Ameausoone:
OK thank you for the quick response !

It's not the day I finally commit to terraform-google-modules :')

ludoo:
Ahhhh sorry for breaking your flow. :) Don't get discouraged though!

If you want a simple but useful PR to break the ice, I'm pretty sure some of the modules' README files are missing a TOC (which is generated automatically with our `tfdoc` tool once a set of tags are added to the file).

That would be merged instantly. There are a lot more things that need refining in the repo, if you want I can spend a few minutes and come up with a list. You just were unlucky and picked a controversial one.

PR #2884, opened 2025-02-12
Add new set of org policies with managed constraints to FAST bootstrap

juliocc:
We're not updating the default implementation until the [default set of organization policies for new organizations](https://cloud.google.com/resource-manager/docs/secure-by-default-organizations) moves to managed policies. This PR provides an alternative set of org policies leveraging managed policies for those users that wish to make the switch.

This PR also enables `restrictProtocolForwardingCreationForTypes` to internal only by default.

**Upgrade Notes**

```upgrade-note
fast/stages/0-boostrap: enabled restrictProtocolForwardingCreationForTypes to internal only by default
```

```upgrade-note
fast/stages/0-boostrap/data/org-policies-managed: new set of org policies using managed constraints
```


ludoo (approved):
What about a `data/org-policies-managed` folder with an alternate set, so one just swaps path in `factories_config`? So we can also test it.

juliocc:
> What about a `data/org-policies-managed` folder with an alternate set, so one just swaps path in `factories_config`? So we can also test it.

Great idea. Stand by

Issue #2885, opened 2025-02-12
password_validation_policy in DB prevents creating Cloud SQL instance


lyricnz:
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

As mentioned in the FYI at https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/2722#issuecomment-2608497173 , adding the `password_validation_policy` stanza to the DB creation prevents it from actually creating

If I try and create a DB with

```
  #   root_password = null # generate password

  password_validation_policy = {
    enabled                     = true
    default_complexity          = true
    disallow_username_substring = true
    min_length                  = 10
    reuse_interval              = 5
  }
```

I get an error about root password doesn't meet password policy. I am forced to comment out the password_validation_policy block to create the DB, then enable it afterwards and (re) apply

### Release version

5e31633f3

### Terraform version

OpenTofu v1.9.0

### Steps to reproduce

**Create** a DB using the module, with `password_validation_policy` included (it works fine updating an existing db).

### Expected behavior

DB should be created, with any specified accounts obeying the password policy.

### Actual behavior

Sorry, I don't have the error to hand.  Will try to reproduce later and follow up with exact error text.

juliocc:
If you set `password_validation_policy.default_complexity`, then the password must include a combination of lowercase, uppercase, numeric, and non-alphanumeric characters. The random password you're getting probably doesn't meet the criteria.

Quick workaround: disable default_complexity
Better solution: update the code to ensure [random passwords](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/alloydb/main.tf#L419C11-L419C26) include [`min_lower`, `min_upper`, `min_special`, `min_numeric`](https://registry.terraform.io/providers/hashicorp/random/3.6.3/docs/resources/password#min_lower-1)

lyricnz:
To reproduce:

```
module "db" {
  source     = "git::https://github.com/GoogleCloudPlatform/cloud-foundation-fabric.git//modules/cloudsql-instance"
  project_id = MY_PROJECT_ID
  name       = "test-db"

  region = "australia-southeast2"
  tier   = "db-g1-small"
  network_config = {
    connectivity = {
      psa_config = {
        private_network = "projects/MY_PROJECT_ID/global/networks/default"
      }
    }
  }
  database_version = "MYSQL_8_0"


[...]

PR #2886, opened 2025-02-13
Increase the default complexity of Cloud SQL DB passwords

lyricnz:
Per #2885

NOTE: this changes the type of `root_password` variable from a string (default null) to an object with `password` (same behaviour as old string) and `random_password` attributes (override `password` with a randomly-generated password).  if value was not previously provided, default behaviour is the same.

~~NOTE: this is a change in behaviour. Currently omitting `root_password` parameter to `modules/cloudsql-instance` will pass `null` to the provider (which may not set a password at all).  With this PR, a new+random password will be generated and passed to the TF provider.
Please let me know if this is a bad idea.  Perhaps a magic-value for the password (or a separate variable) could trigger the generation, rather than value `null`?~~

---
**Checklist**


lyricnz:
Added code to explicitly generate a root_password also (using the same logic).  Otherwise the DB fails to create because the default (Google generated) password doesn't match the password policy.

lyricnz:
> Added code to explicitly generate a root_password also (using the same logic). Otherwise the DB fails to create because the default (Google generated) password doesn't match the password policy.

Maybe the module shouldn't do this?  And rely on the caller passing a valid root_password, rather than generating it here?  (but that's inconsistent with the pattern for non-root users, which _do_ generate passwords if not provided)

modules/cloudsql-instance/main.tf
```diff
@@ -52,7 +52,7 @@ resource "google_sql_database_instance" "primary" {
   region              = var.region
   database_version    = var.database_version
   encryption_key_name = var.encryption_key_name
-  root_password       = var.root_password
+  root_password       = var.root_password == null ? random_password.root_password[0].result : var.root_password
```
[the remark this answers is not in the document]
lyricnz: I was trying not to break existing usage of the module - do you think that's a worthwhile objective?  If so, we could just add a new top-level variable auto_generate_root_password, which (if root_password==null) would use the code in the PR?

modules/cloudsql-instance/main.tf
```diff
@@ -289,8 +289,22 @@ resource "random_password" "passwords" {
...
+  length      = try(var.password_validation_policy.min_length, 16)
+  special     = true
+  min_lower   = 1
+  min_numeric = 1
+  min_special = 1
+  min_upper   = 1
+}
+
+resource "random_password" "root_password" {
+  count       = var.root_password == null ? 1 : 0
```
lyricnz, line 301: Should we bother using count to prevent this resource being created (and not used) in some cases?  It's only a local resource, and means that we need to use `[0]` when accessing it.

juliocc: No issues using count if you don't need the password. It's actually a pretty common pattern.

modules/cloudsql-instance/main.tf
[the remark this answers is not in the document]
lyricnz: So... which way should we go here, team?  In the meantime, I just cut/paste the random block into my app code, and passed root_password as a var; but feel like having the resource fail to create "just because" I turned on `password_validation_policy` (and don't have an explicit `root_password` is a bit uncool).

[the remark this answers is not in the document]
ludoo: > I was trying not to break existing usage of the module - do you think that's a worthwhile objective? If so, we could just add a new top-level variable auto_generate_root_password, which (if root_password==null) would use the code in the PR?

We prefer keepiing the variable surface small usually, even at the cost of breaking backward compatibility.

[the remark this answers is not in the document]
lyricnz: @ludoo updated PR.  Will add testing notes to the PR also.

lyricnz:
Changed the `root_password` input to an object, as suggested. Here is the results of my testing with this code (the first case is the original bug #2885)

# DB with no root_password config

Input:

```
module "db-none" {
  # no root_password config
  password_validation_policy = {
    enabled                     = true
    default_complexity          = true
    disallow_username_substring = true
    min_length                  = 20
    reuse_interval              = 5
  }
}
```

Output:

Fails to create: `googleapi: Error 400: Invalid request: Root user's password is not policy compliant: Length of password cannot be less than 20. Password does not meet the complexity criteria. ., invalid`

# DB with explicit password (too short)

Input:

```
module "db-explicit-short" {
  root_password = {
    password = "xxx"
  }
  password_validation_policy = { ... as above ... }
}
```

Output:

Fails to create: `googleapi: Error 400: Invalid request: Root user's password is not policy compliant: Length of password cannot be less than 20. ., invalid`

# DB with explicit password (OK)

Input:

```
module "db-explicit" {
  root_password = {
    password = "Qy[E7N&Jv49u[F9jf3Ec"
  }
  password_validation_policy = { ... as above ... }
}
```

Output:

```
$ terraform state show -show-sensitive module.db-explicit.google_sql_database_instance.primary | grep root
    root_password                  = "Qy[E7N&Jv49u[F9jf3Ec"
```

# DB with random_password

Input:

```
module "db-random" {
  root_password = {
    random_password = true
  }
  password_validation_policy = { ... as above ... }
}
```

Output:

```sh
$ terraform state show -show-sensitive module.db-random.google_sql_database_instance.primary | grep root
    root_password                  = "K*7bTarLtA-1!W=KK@$c"
```

# DB with both explicit password and random_password (override)

Input:

```
module "db-override" {
  root_password = {
    password = "Qy[E7N&Jv49u[F9jf3Ec"
    random_password = true
  }
  password_validation_policy = { ... as above ... }
}
```

Output:

```sh
$ terraform state show -show-sensitive module.db-override.google_sql_database_instance.primary | grep root
    root_password                  = "FZ1F4aQ3&qVKM5B<tUrx"
```

ludoo:
This is great, thanks for taking the time to run the changes. Can I ask you one last effort

- check the examples in the README to ensure the cases above are explained (if they are missing, a simple paragraph or code block is all we need)
- check that everything works without setting password validation

modules/cloudsql-instance/main.tf
```diff
@@ -52,7 +52,7 @@ resource "google_sql_database_instance" "primary" {
   region              = var.region
   database_version    = var.database_version
   encryption_key_name = var.encryption_key_name
-  root_password       = var.root_password
+  root_password       = var.root_password.random_password ? random_password.root_password[0].result : var.root_password.password
```
juliocc, line 55: This will fail if `var.root_password` is null. Make the variable `nullable = false` to avoid having to check for null values.

modules/cloudsql-instance/variables.tf
```diff
@@ -250,8 +250,11 @@ variable "replicas" {
 
 variable "root_password" {
   description = "Root password of the Cloud SQL instance. Required for MS SQL Server."
-  type        = string
-  default     = null
+  type = object({
+    password        = optional(string, null)
```
juliocc, line 254: nit: `optional(string)` (the null is redundant)

modules/cloudsql-instance/variables.tf
```diff
@@ -250,8 +250,11 @@ variable "replicas" {
...
+    random_password = optional(bool, false) # overrides password
+  })
+  default = {}
```
juliocc, line 257: make non-nullable to avoid errors mentioned above

modules/cloudsql-instance/variables.tf
```diff
@@ -250,8 +250,11 @@ variable "replicas" {
 
 variable "root_password" {
```
juliocc, line 251: nit: you could add a validation to ensure only one of password or random_password is provided.

Something like this (untested):
```
validation {
  condition = (var.root_password.password != null) != var.root_password.random_password
  error_message = "Cannot provide root_password.password and root_password.random_password at the same time"
}
```

juliocc (changes requested):

modules/cloudsql-instance/variables.tf
lyricnz: It's valid to provide password=null and random_password=false (this should pass null to the provider - current behaviour). I think the only invalid situation is password != null and random_password = true.

lyricnz: I used the following, because it seemed clearer than negating the terms in the expression. Thoughts?
```
  validation {
    condition     = !(var.root_password.password != null && var.root_password.random_password)
    error_message = "Cannot provide root_password.password and root_password.random_password at the same time"
  }
```

lyricnz:
I tested with no password_policy:

```hcl
module "db-none" {
  source     = "../modules/cloudsql-instance"
  name       = "test-db-no-root-password"
}

module "db-no-policy-password" {
  source     = "../modules/cloudsql-instance"
  name       = "test-db-no-policy-password"
  root_password = {
    password = "Qy[E7N&Jv49u[F9jf3Ec"
  }
}

module "db-no-policy-random" {
  source     = "../modules/cloudsql-instance"
  root_password = {
    random_password = true
  }
}
```

They all appeared to work fine.  The validation caught the following error:

```hcl
module "db-both" {
  source     = "../modules/cloudsql-instance"
  root_password = {
    password = "xyzzy"
    random_password = true
  }
}
```

error
```
│   on main.tf line 93, in module "db-both":
│   93:   root_password = {
│   94:     password = "xyzzy"
│   95:     random_password = true
│   96:   }
│     ├────────────────
│     │ var.root_password.password is "xyzzy"
│     │ var.root_password.random_password is true
│
│ Cannot provide root_password.password and root_password.random_password at the same time
│
│ This was checked by the validation rule at ../modules/cloudsql-instance/variables.tf:259,3-13.
```

juliocc (approved):
LGTM.

Thanks @lyricnz!

PR #2895, opened 2025-02-14
Add perimeter factory for VPCSC stage

karpok78:
A proposal to allow for the management of VPC-SC perimeters in yaml factory. 
Project discovery is extended to multiple perimeters thanks to query filtering.
Also adding support for bridge perimeters
The default behavior with a simple single perimeter remains.

---
**Checklist**


karpok78:
I forgot to update the schema. I will have a look at it

juliocc:
This is really cool.

I wonder if the core functionality should go in the vpc-sc module?

ludoo:
> This is really cool.
> 
> I wonder if the core functionality should go in the vpc-sc module?

Ah yes, the vpc-sc factories are all at the module level so the new ones should go there too. Great point.

karpok78:
I guess the project directory could also be passed down to the module.
And so the factory could be implemented in the module itself. I'll have a look at it

karpok78:
I gave it a first try. All the factory is now in the vpc-sc module.
resource auto-discovery result can be passed in as a parameter for the module.

Note : I realized the stage and the module had opposite defaults for dry run. I chose to keep with the module default. This means enforced mode would be the new default.

ludoo:
> I gave it a first try. All the factory is now in the vpc-sc module. resource auto-discovery result can be passed in as a parameter for the module.
> 
> Note : I realized the stage and the module had opposite defaults for dry run. I chose to keep with the module default. This means enforced mode would be the new default.

The stage defaults to dry run, as this is how we typically run it. Will take a look at the changes ASAP, hopefully by tonight / early tomorrow morning.

fast/stages/1-vpcsc/variables.tf
```diff
@@ -14,183 +14,31 @@
  * limitations under the License.
  */
 
-variable "access_levels" {
```
ludoo, line 17: Please keep the variables here and don't remove them.

fast/stages/1-vpcsc/main.tf
```diff
@@ -27,55 +59,29 @@ locals {
...
   source = "../../../modules/vpc-sc"
   # only enable if the default perimeter is defined
-  count         = var.perimeters.default == null ? 0 : 1
+  count         = length(local._data.perimeters) > 0 ? 1 : 0
   access_policy = var.access_policy
   access_policy_create = var.access_policy != null ? null : {
     parent = "organizations/${var.organization.id}"
     title  = "default"
   }
-  access_levels    = var.access_levels
```
ludoo, line 67: access levels should still be separate from perimeters, and referenced in perimeters by key

ludoo: same for ingress/egress policies

fast/stages/1-vpcsc/variables.tf
karpok78: My bad, I was a bit too enthusiastic.
I restored the missing variables.

ludoo: Thanks :)

To be clear: everything should remain the same from an interface point of view. The new factories -- like any other we have in modules -- will just be an additional source of data that is merged with the variable.

tests/fast/stages/s1_vpcsc/simple.tfvars
```diff
@@ -20,16 +20,7 @@ factories_config = {
...
-perimeters = {
-  default = {
-    access_levels    = ["geo_it", "identity_me"]
-    egress_policies  = ["test"]
-    ingress_policies = ["fast-org-log-sinks", "test"]
-    resources = [
-      "projects/1234567890"
-    ]
-  }
+  perimeters = "../../../tests/fast/stages/s1_vpcsc/data/vpc-sc/perimeters"
```
juliocc, line 23: Can you keep the old test as well?  You'll end up with two tests: one using the perimeters variable (as before) and another using the new factory

fast/stages/1-vpcsc/main.tf
```diff
@@ -15,6 +15,38 @@
...
+  dynamic_projects_map = {
+    for k, v in local._perimeter_filters :
+    k => (var.resource_discovery.enabled != true ? [] : [
+      for p in module.vpc-sc-discovery[k].project_numbers :
+      "projects/${p}"
+      ]
+    )
+    if try(v.type != "bridge", true)
+  }
+
```
juliocc, line 49: You don't need any of these locals anymore, right?

modules/vpc-sc/variables.tf
```diff
@@ -79,6 +79,12 @@ variable "access_policy_create" {
   default = null
 }
 
+variable "dynamic_projects_map" {
+  description = "External source of resources to perimeter mapping"
+  type        = map(list(string))
+  default     = {}
+}
+
```
juliocc, line 87: I don't understand what this is. Can you exaplin?

modules/vpc-sc/service-perimeters-regular.tf
```diff
@@ -26,7 +26,7 @@ locals {
 }
 
 resource "google_access_context_manager_service_perimeter" "regular" {
-  for_each                  = var.service_perimeters_regular
+  for_each                  = merge(var.service_perimeters_regular, local.data.service_perimeters_regular)
```
juliocc, line 29: nit: Can you do the merge in a local? (see just a few lines above this one)

juliocc (changes requested):

ludoo:
This is unmergeable in its current state. The module has evolved in the meantime. You are welcome to send a smaller PR if you are still interested, or we will look at implementing it soonish.

PR #2897, opened 2025-02-14
FAST project templates example

ludoo:
This introduces a new FAST concept for project-level templates, that is intended to replace blueprints and provide application-level setups that can be easily deployed bith via FAST and without.

One additional feature which is not present here, is a way to self-deploy the project based on the provided YAML file. This is meant to be used when deploying outside of FAST.


fast/stages/2-project-factory/templates/providers.tf.tpl
```diff
@@ -0,0 +1,30 @@
+/**
+ * Copyright 2022 Google LLC
```
juliocc, line 2: 2025

juliocc (approved):

fast/stages/2-project-factory/templates/providers.tf.tpl
ludoo: 🤦

PR #2910, opened 2025-02-19
Add ssl_mode support to cloudsql-instance replicas

sruffilli:
Adds `ssl_mode` support for cloudsql replicas, mirroring the same configuration of the primary instance


modules/cloudsql-instance/main.tf
```diff
@@ -234,6 +234,7 @@ resource "google_sql_database_instance" "replicas" {
       allocated_ip_range = try(
         var.network_config.connectivity.psa_config.allocated_ip_ranges.replica, null
       )
+      ssl_mode = var.ssl.ssl_mode
```
ludoo, line 237: I would rename this to `var.ssl.mode` :)

ludoo (approved):

modules/cloudsql-instance/main.tf
ludoo: I would do it :)

PR #2911, opened 2025-02-19
Add support for max workstations, refactor timeouts in workstation-cluster module

ludoo:
**Breaking Changes**

```upgrade-note
`modules/workstation-cluster`: changed the interface for configuration timeouts to object, timeouts are now specified as numbers.
```


sruffilli (approved):
LGTM - one qq: does this still need to use `google-beta`?

ludoo:
> LGTM - one qq: does this still need to use `google-beta`?

yep, according to the resource docs it does; good call, let's remember to revisit periodically (actually, we should do it for all modules once in a while)

PR #2919, opened 2025-02-21
Add perimeter factory to `modules/vpc-sc`

karpok78:
Proposal to introduce YAML factory to configure VPC-SC perimeters

Module VPC-SC:

- has been update to support both tfvars and yaml file for perimeters configuration
- has a new variable 'dynamic_projects_map' to leverage external dynamic resource membership

Stage 1-vpvsc has been updated:

- perimeters are now defined in a dedicated YAML factory
- project discovery now implements query filters to support for multiple perimeters configuration. The filter can be defined in yaml perimeter files (default to all active project)
- the result of project discovery is ingested in vpc-sc module with the new dynamic_projects_map variable

---
**Checklist**


modules/vpc-sc/variables.tf
```diff
@@ -79,6 +79,12 @@ variable "access_policy_create" {
   default = null
 }
 
+variable "dynamic_projects_map" {
+  description = "External source of resources to perimeter mapping"
+  type        = map(list(string))
+  default     = {}
+}
+
```
juliocc, line 87: What is this for?

modules/vpc-sc/factory.tf
```diff
@@ -89,6 +89,57 @@ locals {
         }
       }
     }
+    service_perimeters_bridge = {
+      for k, v in local._data.perimeters :
+      k => {
+        description = try(v.description, "")
```
juliocc, line 95: default to null, here

modules/vpc-sc/factory.tf
```diff
@@ -89,6 +89,57 @@ locals {
...
+        spec_resources = (
+          try(v.dry_run, false)
+          ? distinct(concat(v.resources,
+          flatten([for p in v.perimeters : var.dynamic_projects_map[p]])))
+          : null
```
juliocc, line 100: formatting is wrong here. Pleas pass your code through `terraform fmt`

fast/stages/1-vpcsc/schemas/perimeters.schema.json
```diff
@@ -0,0 +1 @@
+../../../../modules/vpc-sc/schemas/perimeters.schema.json
```
juliocc, line 1: broken link?

fast/stages/1-vpcsc/variables.tf
```diff
@@ -170,29 +173,6 @@ variable "outputs_location" {
...
-  }))
-  nullable = false
-  default = {
-    default = {
-      access_levels    = ["geo"]
-      ingress_policies = ["fast-org-log-sinks"]
-    }
-  }
-}
-
```
juliocc, line 195: the stage should still support the variable.

juliocc (dismissed):

modules/vpc-sc/variables.tf
karpok78: As mentioned in the comment, this allow to keep the auto-discovery feature.
It allows to surcharge the static resource to perimeter mapping in the yaml file.

fast/stages/1-vpcsc/schemas/perimeters.schema.json
karpok78: I don't get it. It's not broken for me and points to the schema in the module

fast/stages/1-vpcsc/variables.tf
karpok78: This one is much more tricky.
Currently the vpc-sc module does not consume the perimeters variable directly. Instead in relied directly on service_perimeters_bridge et service_perimeters_regular.
With the perimeter factory in the module, we could:
- keep the current perimeters variable in the stage. But it would introduce an inconsistency between tfvars (computed locally) and the factories (computed in the module)
- move the tfvars computation in the module too, but it will break compatibility for existing consumers
- only support factory in the module.

The latest seems the nicest to me. But I could still change it

juliocc:
@karpok78, we have some concerns regarding the current direction of this pull request.

We propose a slightly different approach:
1. First, modify the VPC-SC module independently of FAST. This change should be self-contained, allowing perimeter definitions exclusively through the factory.
2. Once we've updated the module, we update FAST in a separate pull request.

In addition to this, we believe there's a mix up wiuth definition of the perimeter and discovery of resources. The VPC-SC module does not and should not care how you pass in the projects belonging to the perimeters, that's up to the caller to figure out.

Please allow us a few hours to do a some changes to this PR to make sure we can align your idea to the practices in our codebase.

fast/stages/1-vpcsc/variables.tf
karpok78: I agree with you there should not be a big churn on this.

But I found interesting to:

- have a common interface for access-levels, policies and perimeters, rather than a mix of yaml and tfvars
- expand the auto-discovery feature to allow for more than a single perimeter.

And from a gitops point of view, it could indeed integrates with the project factory in some more complex pipelines:

1. You manage policies and core perimeter memberships in the vpc-sc factory
1. You create new project with the project-factory and give them a proper label
2. New projet perimeter membership is computed from the label
3. This could trigger an update of vpc-sc configuration (automatic or semi-automatic). Same could apply when project are deleted or even moved between perimeters.
4. Eventually auto-discoveries could be hard coded back in static file whenever secops want it to be documented somewhere

juliocc:
Here's a summary of the changes
- Aligned the factory to the variables types
- Add title and description to all fields
- Added context to the factory to interpolate lists of projects and services in the relevant fields
-

ludoo (approved):

PR #2959, opened 2025-03-16
Rationalize project factory context interpolations for automation service accounts

### LINKED ISSUE #2957: Project Factory improvements
- add support for interpolating project id in `automation.project`
- document the interpolation format in project IAM for automation service accounts (`project key/service account key`)
  - also support using plain keys from within the same project taking into account the new scope described below, so that both `automation/project key/service account key` and `automation/service account key` work
- scope automation service account and bucket keys, so that we don't risk overlap with project-level service accounts and keys (e.g. `automation/project key/service account key` instead of `project key/service account key`)

more to add later

ludoo:
Fixes #2957

### DIFF
--- a/modules/project-factory/automation.tf
+++ b/modules/project-factory/automation.tf
@@ -65,7 +65,13 @@ module "automation-bucket" {
           # project/automation/rw project/sa
           var.factories_config.context.iam_principals[vv],
           # fully specified principal
-          vv
+          vv,
+          # passthrough + error handling using tonumber until Terraform gets fail/raise function
+          (
+            strcontains(vv, ":")
+            ? vv
+            : tonumber("[Error] Invalid member: '${vv}' in automation bucket '${each.key}'")
+          )
         )
       ]
     })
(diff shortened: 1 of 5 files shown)


modules/project-factory/automation.tf
```diff
@@ -58,8 +58,13 @@ module "automation-bucket" {
...
       members = [
         for vv in v.members : try(
+          # rw (infer local project and automation prefix)
+          module.automation-service-accounts["${each.key}/automation/${vv}"].iam_email,
+          # automation/rw or sa (infer local project)
           module.automation-service-accounts["${each.key}/${vv}"].iam_email,
+          # project/automation/rw project/sa
           var.factories_config.context.iam_principals[vv],
+          # fully specified principal
           vv
```
juliocc, line 68: Should we do the `strcontains` thing here too?

juliocc (approved):

modules/project-factory/automation.tf
ludoo: done, good call

PR #2954, opened 2025-03-13
Mongodb part of the Cloud Foundation Fabric

theshanbhag:
This PR will consist of scripts to add MongoDB as part of Cloud Foundation Fabric

---
**Checklist**

I applicable, I acknowledge that I have:
- [ X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [ X] Ran `terraform fmt` on all modified files
- [ ] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [ X] Made sure all relevant tests pass

PR #2986, opened 2025-03-28
Mongodb Atlas cluster project template

ludoo:
Supersedes #2954

TODO

- [x] define meaningful outputs
- [x] improve README


fast/project-templates/mongodb/project.yaml
```diff
@@ -0,0 +1,47 @@
...
+        - automation/rw
+iam:
+  roles/compute.admin:
+    - automation/rw
+  roles/servicedirectory.admin:
+    - automation/rw
+shared_vpc_service_config:
+  host_project: dev-spoke-0
+  network_users:
+    - automation/rw
```
wiktorn, line 47: I'd comment out `parent`, `automation`, `iam`, and `shared_vpc_service_config`, as this is not necessary for this template

wiktorn (approved):

fast/project-templates/mongodb/project.yaml
ludoo: It is to have it work in fast, which is the primary use case

ludoo: I edited the file a bit, should be clearer to understand now.

PR #2997, opened 2025-04-01
Allow disabling GKE IP endpoints and setting GKE VPC scope DNS domain

juliocc:
* Allow disabling IP access to GKE clusters
* Allow setting Additive VPC scopo dns domain
* Make `gcp_public_cidrs_access_enabled` default to false

**Breaking Changes**

```upgrade-note
`modules/gke-cluster-standard`, `modules/gke-cluster-autopilot`: Default value for `access_config.ip_access` changed from `{}` to `null`; explicitly set to keep IP access enabled.
```


modules/gke-cluster-standard/variables.tf
```diff
@@ -21,12 +21,12 @@ variable "access_config" {
     ip_access = optional(object({
       authorized_ranges               = optional(map(string), {})
       disable_public_endpoint         = optional(bool, true)
-      gcp_public_cidrs_access_enabled = optional(bool, true)
+      gcp_public_cidrs_access_enabled = optional(bool, false)
```
ludoo, line 24: 👍

ludoo (approved):

PR #2998, opened 2025-04-01
Fix handling of data_overrides in project factory

wiktorn:
`coalesce([], ["a"]) == []` and `coalesce({}, {a = 1}) == {}`, so it's wrong default value for data_overrides. Changed that to `null` + added few comments and tests.

Allow multiple project files of the same name in factory, as long as they override the name. If duplicate is found, following error is raised:
```
Error: Duplicate object key

  on factory-projects.tf line 39, in locals:
  35:   _projects_input = {
  36:     # will raise error, if the same filename is raised multiple times
  37:     # and project name is not set via name in YAML
  38:     for k, v in merge(local._hierarchy_projects_full_path, local._projects_full_path) :
  39:     lookup(v, "name", basename(k)) => v
  40:   }

Two different items produced the key "automation" in this 'for' expression.
If duplicates are expected, use the ellipsis (...) after the value expression
to enable grouping by key.
```
Which hopefully will help users to understand where the error is.

---
**Checklist**

### DIFF
--- a/modules/project-factory/factory-projects-object.tf
+++ b/modules/project-factory/factory-projects-object.tf
@@ -101,8 +101,18 @@ locals {
       tag_bindings               = null
       services                   = null
       service_accounts           = null
-      vpc_sc                     = try(local._projects_config.data_overrides.vpc_sc, null)
-      logging_data_access        = null
+      vpc_sc = try(
+        merge(
+          {
+            perimeter_name    = null
+            perimeter_bridges = []
+            is_dry_run        = false
+          },
+          local._projects_config.data_overrides.vpc_sc
+        ),
+        null
+      )
+      logging_data_access = null
       },
       try(local._projects_config.data_overrides, {})
     )
--- a/modules/project-factory/factory-projects.tf
+++ b/modules/project-factory/factory-projects.tf
@@ -33,7 +33,7 @@ locals {
     trimsuffix(f, ".yaml") => yamldecode(file("${local._project_path}/${f}"))
   }
   _projects_input = {
-    # will raise error, if the same filename is raised multiple times
+    # will raise error, if the same filename is used multiple times
     # and project name is not set via name in YAML
     for k, v in merge(local._hierarchy_projects_full_path, local._projects_full_path) :
     lookup(v, "name", basename(k)) => v


modules/project-factory/factory-projects-object.tf
```diff
@@ -95,107 +96,111 @@ locals {
...
-          perimeter_name    = null
-          perimeter_bridges = []
-          is_dry_run        = false
-        })
-      )
-      logging_data_access = {}
+      tag_bindings               = null
+      services                   = null
+      service_accounts           = null
+      vpc_sc                     = try(local._projects_config.data_overrides.vpc_sc, null)
```
ludoo, line 104: you sure we don't need to prime the optional parts of vpc sc?

wiktorn: If this is null, it will be primed below

modules/project-factory/factory-projects-object.tf
```diff
@@ -95,107 +96,111 @@ locals {
...
-    for k, v in local._projects_input : lookup(v, "name", k) => merge(v, {
-      billing_account = try(coalesce(
+    # Semantics of the merges are:
+    # * if data_overrides.<field> is not null, use this value
+    # * if  _projects_inputs.<field> is not null, use this value
+    # * use data_default value, which if not set, will provide "empty" type
+    # This logic is easily implemented using coalesce, even on maps and list and allows to
+    # set data_overrides.<field> to "", [] or {} to ensure, that empty value is always passed, or do
+    # the same in _projects_input to prevent falling back to default value
+    for k, v in local._projects_input : k => merge(v, {
```
ludoo, line 118: I think we need to keep name as key. There are situations where the project data folder uses subfolders for organizing project files, and those are (rightly) stripped from they key. But then you risk duplicates in keys if you don't use the name as key.

Having the same file name across subfolders and then overriding with name inside the YAML is a valid pattern, and we should support it IMHO. The project factory is not a library, and the primary use case is supporting valid patterns in this module. If the net factory has an issue with this, it's a problem that needs to be addressed there.

modules/project-factory/factory-projects-object.tf
ludoo: I don't mean null, I mean an incomplete object.

modules/project-factory/factory-projects-object.tf
wiktorn: That's valid point, but risk of duplicates will happen before this stage, as here we already have a map on input. I'd opt to do that either while creating `local._projects_input` or post-processing `local._projects_output` if necessary and outside of this file.

Then we can provide guarantee, that keys here are stable between input and output, and you have the ability to cross-reference.

ludoo: I don't think that addresses support for the pattern described above. I don't see an issue using names as keys, control is still on the caller that defines both keys and names...

ludoo (approved):

PR #2999, opened 2025-04-02
Added variable for activating nat and implementation in google_apigee…

jacklever-hub24:
This allows you to set the activation status of an apigee NAT object. Previously it had to be manually activated by api calls or if it's imported and the status is different from the default (only option available at the moment) it will perpetually try to update this.

No breaking changes, defaults to false as-is currently the default.

### DIFF
--- a/tests/modules/apigee/test_apigee_nat_activate.tfvars
+++ b/tests/modules/apigee/test_apigee_nat_activate.tfvars
@@ -18,7 +18,7 @@ environments = {
 instances = {
   europe-west1 = {
     environments = ["europe-west1"]
-    enable_nat = true
+    enable_nat   = true
     activate_nat = true
   }
 }


juliocc (approved):

modules/apigee/main.tf
```diff
@@ -113,6 +113,7 @@ resource "google_apigee_nat_address" "apigee_nat" {
   }
   name        = each.key
   instance_id = each.value
+  activate    = coalesce(var.instances[each.key].activate_nat, false)
```
juliocc, line 116: This coalesce is redundant, `var.instances[each.key].activate_nat` will never be null

jacklever-hub24: I'll fix this, get the tests building and ready it for a merge.

jacklever-hub24: These changes fix it.

PR #3008, opened 2025-04-04
Add support to attach tags to service accounts

juliocc:
- Allow attaching tags to service accounts
- Remove service account key upload
- Expose `create_ignore_already_exists`

**Breaking Changes**

```upgrade-note
`modules/iam-service-account`: removed `public_keys_directory` variable. Use bare `google_service_account_key` resources if this functionality is needed.
```


ludoo (approved):

modules/iam-service-account/variables.tf
oscarpulido55, line 1: This change introduced next error:
The condition for variable "create_ignore_already_exists" can only refer to the variable itself, using var.create_ignore_already_exists.
 
Since Terraform v0.15, a validation block within a variable definition can only refer to the variable it's defined for (e.g., var.create_ignore_already_exists inside the validation for create_ignore_already_exists). This rule violates Terraform's constraint because it references var.service_account_create in addition to var.create_ignore_already_exists.

PR #3012, opened 2025-04-07
Make FAST aggregated sinks use a project destination

juliocc:
This PR changes FAST's top-level sinks to use a project destination. Now the logging project is the one routing log messages to their final destination.

**Breaking Changes**

```upgrade-note
`fast/stages/0-boostrap`: log sinks no longer support project destinations. By default, all logs are sent to the logging project.
```


ludoo:
I don't like having a single project as a destination. Please let's keep the current approach.

PR #3029, opened 2025-04-11
Allow IAP configuration with default IdP

stribioli:
Load balancers can be configured with IAP-enabled backends. They can either be configured to use external Identity Providers (IdP) or to use Cloud Identity. The latter is the default and is used when the OAuth2 parameters are not specified.

The `iap_config` parameter in the `backend_service_configs` variable already supported the external IdP option, but did not support the default one.

For reference: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_backend_service#nested_iap

---
**Checklist**

I acknowledge that I have:


ludoo (approved):
Thanks Stefano!

PR #3045, opened 2025-04-18
Allow direct assignment of project id without prefix

braincow:
This PR implements a feature that assists on deploying Cloud Foundation Fabric to a brownfield installation. 

An existing GCP organization adopting the Foundation Fabric will have, usually, projects that do not follow naming conventions in their project ids. With this feature addition a field called "project_id" will reflect the id of the project and thus allow adopting it to the fabric via Terraform imports.

---
**Checklist**

### DIFF
--- a/modules/project-factory/factory-projects-object.tf
+++ b/modules/project-factory/factory-projects-object.tf
@@ -43,6 +43,7 @@ locals {
       metric_scopes              = []
       parent                     = null
       prefix                     = null
+      project_id                 = null
       service_encryption_key_ids = {}
       services                   = []
       shared_vpc_service_config = merge({
@@ -96,6 +97,7 @@ locals {
       )
       parent                     = null
       prefix                     = null
+      project_id                 = null
       service_encryption_key_ids = null
       storage_location           = null
       tag_bindings               = null
@@ -181,6 +183,7 @@ locals {
         local.__projects_config.data_defaults.metric_scopes
       )
       name         = lookup(v, "name", basename(k)) # type: string
+      project_id   = lookup(v, "project_id", null)  # type: string, nullable
       org_policies = try(v.org_policies, {})        # type: map(object({...}))
       parent = try(                                 # type: string, nullable
         coalesce(
--- a/modules/project-factory/main.tf
+++ b/modules/project-factory/main.tf
@@ -37,11 +37,11 @@ module "projects" {
   source          = "../project"
   for_each        = local.projects
   billing_account = each.value.billing_account
-  name            = each.value.name
+  name            = each.value.project_id == null ? each.value.name : each.value.project_id # if project_id is present it takes precedence
   parent = lookup(
     local.context.folder_ids, each.value.parent, each.value.parent
   )
-  prefix              = each.value.prefix
+  prefix              = each.value.project_id == null ? each.value.prefix : null # if project_id is present no prefix should be used
   alerts              = try(each.value.alerts, null)
   auto_create_network = try(each.value.auto_create_network, false)
   compute_metadata    = try(each.value.compute_metadata, {})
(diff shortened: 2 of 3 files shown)


juliocc:
Why do you need a new field for this? I think setting the name without prefix has the same effect.

braincow:
@juliocc no, it does not when using fast apparently. Fast seems to have the data override for "prefix" field and thus it cant be set empty in the project factory data fields on stage 2. Only way to override that is to create a corner case and handle that by bypassing the name and prefix during project creation.

juliocc:
I guess you mean [this](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/4821ac2c07e9414af2601b8854ac255af26ee934/fast/stages/2-project-factory/main.tf#L32). Just remove that line and you're good to go

braincow:
> I guess you mean [this](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/4821ac2c07e9414af2601b8854ac255af26ee934/fast/stages/2-project-factory/main.tf#L32). Just remove that line and you're good to go

Yup. Thats the override. However removing that would break the "existing interface" which is against rules in the contribution guidelines is it not?

ludoo:
WDYM with breaking the existing interface? That stage is minimal and meant to be adapted. :)

braincow:
> WDYM with breaking the existing interface? That stage is minimal and meant to be adapted. :)

Adapting that per deployment basis is of course an option, but it should not be necessary in most use cases. It makes the fast framework more user fiendly to give the deployment team an option to keep their codebase more upstream compatible by not modifying that part of the code and still give them flexibility to deploy in brownfield org. If you remove that line in the code then you also have to remember to always also specify the prefix in every single project definition.

With my PR fast framework works as previously but gives the admins the option to handle existing legacy projects in the org gracefully as well.

braincow:
@ludoo pardon my ignorance, but that would still override the prefix for all projects handled by the stage 2 project factory? Again requiring admins to after that define the prefix for all projects and not just new ones.

Consider the following:
1. You have a cloud governance model that dictates naming scheme that you wish to follow at all times by default, therefore you enforce the prefix by default as the project factory currently does
2. You have existing projects in your gcp org that you wish to also govern now with the same Terraform stack as your future projects
3. You cant currently do that as the project factory dictates the naming and therefore is uncompatible with the legacy project naming scheme
4. With the feature provided by this PR a workaround is implemented for that use case and leaves the default behavior intact

The project of course decides which PRs get merged, but i have seen this particular problem on multiple customer engagements and have implemented this same behavior now multiple times and thought that since in my experience this is something that is needed many times over that it might therefore also benefit the community at large through the upstream project.

ludoo:
So, either you decide that prefix is important and you make it a default (which is our way of doing things), or you decide it isn't and neuter it, then in each file you have the option of using it or not (which is what you want).

Please stop pushing to have your PR accepted as it won't happen.

braincow:
@ludoo at this stage I am not expecting to see this PR merged, but i am still trying to understand the reasoning behind the decision and wish to discuss it from the projects perspective. Please feel free to say that you are not open to further discussion and I will shut up.

I understand that you say that you are not willing to accept additional fields to the schema, and that there already is a way to achieve non-enforced prefix naming of projects via removing default behavior. However to me this seems counterintuitive approach as the default behavior should guarantee sane defaults and only if there are situations where that needs to be relaxed then it should be possible in a guaranteed way that does not break other features (like the default behavior) or compatibility with upstream. And as I personally have seen this exact scenario multiple times on customer engagements I find it an odd design policy to enforce here.

Its also true that fast is intended to be modified by the deployers of fast to their liking, but on the field people who do not know how to work with Terraform also deploy this for their businesses as it is the recommended path to take. Therefore we as a community should be here to assist them on making the right decisions and help them govern their environments properly, right?

PR #3050, opened 2025-04-20
New Dataplex Aspect Types module

### LINKED ISSUE #3047: Fix and rename versions.tf file in secops stage
### Affected components

- [ ] Fabric modules
- [x] FAST
- [ ] Blueprints
- [ ] Other

### The problem

The secops stage used a versions.tf file which gets overwritten when we publish a release. The file needs to be renamed versions-override.tf, and the original content brought back.

### Release version

39.0.0

### Terraform version

whatever

### Steps to reproduce

---

### Expected behavior

--

### Actual behavior

---

### Additional context

_No response_

ludoo:
Prerequisite for the new data platform, due to the pending deprecation of Data Catalog Tag Templates.

image

This also fixes #3047 which is generating an unrelated test error.


lcaggio (approved):
Thanks!

LGTM

PR #3060, opened 2025-04-29
Add deletion_policy to project-factory module

tyler-sommer:
This pull request adds the ability to configure project `deletion_policy` from the project-factory module with no default value, the intent being to fallback to the default in the project module when not set.

Side note: curious about why the project default `deletion_policy` is set to `DELETE`? The default for the `google_project` resource is actually `PREVENT` -- upgrading from v30 to v39 we hit this where all our projects' deletion policies are changing from `PREVENT` to `DELETE`. Not a huge deal if we can override from the project-factory :)

---
**Checklist**

I acknowledge that I have:

I spent time trying to run tests locally but cannot figure out how. I get `ModuleNotFoundError: No module named 'tftest'` when running pytest 8.3.5 (python 3.11.6) even though I have installed tftest and all other deps in `tests/requirements.txt`.


modules/project-factory/variables.tf
```diff
@@ -19,6 +19,7 @@ variable "data_defaults" {
   type = object({
     billing_account = optional(string)
     contacts        = optional(map(list(string)), {})
+    deletion_policy = optional(string)
```
ludoo, line 22: I don't see these used in the factory normalization...

ludoo:
This came up recently, I completely forgot this field even existed. Out approach is to ignore it by default since AFAIK it's a provider-level feature and not part of the project API, and it introduces an asymmetry with gcloud etc. It's also fairly easy to undelete a project, so I don't see much benefit using this.

But yes, we should still offer a way to override the default so thanks for this PR. Can you make sure defaults/overrides are used when we normalize the YAML data?

tyler-sommer:
Thanks for the quick reply. I've added the necessary bits to the normalized object. It looks like some tests as well lint is failing. Running `./tools/lint.sh` locally does not raise any errors. Let me do a bit of poking around to try and get both those sorted out.

juliocc:
@ludoo I fixed the failing tests. Can you TAL an approve if it looks good to you?

ludoo (approved):

juliocc:
> Thanks for the quick reply. I've added the necessary bits to the normalized object. It looks like some tests as well lint is failing. Running `./tools/lint.sh` locally does not raise any errors. Let me do a bit of poking around to try and get both those sorted out.

I fixed it for you :)

PR #3061, opened 2025-04-30
AlloyDB read poll support and various usability fixes

viliampucik:
* Added support for read poll instances.
* Added support for public (outbound) IPs.
* Added new arguments: `machine_type`, `skip_await_major_version_upgrade`, and `subscription_type`.
* Added outputs for key cluster attributes, including ID and name.
* Improved input variable validations and simplified configurations.
* Fixed bug where AlloyDB user accounts always had null passwords.
* Resolved Terraform state drift issues for `network_config` and `psc_config`.
* Fixed `continuous_backup_config` always being implicitly or explicitly enabled.
* Fixed `automated_backup_policy` and `maintenance_update_policy`, which do not support non-zero minutes, seconds and nanos.
* Fixed various typos.
* Aligned the format of test examples.

---
**Checklist**

### DIFF
--- a/modules/alloydb/main.tf
+++ b/modules/alloydb/main.tf
@@ -17,42 +17,60 @@
 locals {
   prefix      = var.prefix == null ? "" : "${var.prefix}-"
   is_regional = var.availability_type == "REGIONAL"
-  # secondary instance type is aligned with cluster type unless apply is targeting a promotion, in that
-  # case cluster will be 'primary' while instance still 'secondary'.
+  flags       = try(var.network_config.psa_config.enable_outbound_public_ip, false) ? merge(var.flags, { "password.enforce_complexity" = "on" }) : var.flags # Outbound Public IP requires password complexity to be enabled.
+
+  require_connectors = try(var.client_connection_config.require_connectors, false) ? true : null
+  ssl_mode           = try(var.client_connection_config.ssl_config.ssl_mode, null)
+
+  has_public_ip                = try(var.network_config.psa_config.enable_public_ip, false) || try(var.network_config.psa_config.enable_outbound_public_ip, false)
+  authorized_external_networks = toset(try(var.network_config.psa_config.authorized_external_networks, []))
+  enable_public_ip             = try(var.network_config.psa_config.enable_public_ip, false) ? true : null
+  enable_outbound_public_ip    = try(var.network_config.psa_config.enable_outbound_public_ip, false) ? true : null
+  allowed_consumer_projects    = try(var.network_config.psc_config.allowed_consumer_projects, [])
+
   primary_cluster_name    = "${local.prefix}${var.cluster_name}"
   primary_instance_name   = "${local.prefix}${var.instance_name}"
+  primary_kms_key_name    = try(var.encryption_config.primary_kms_key_name, null)
   secondary_cluster_name  = coalesce(var.cross_region_replication.secondary_cluster_name, "${var.cluster_name}-sec")
   secondary_instance_name = coalesce(var.cross_region_replication.secondary_instance_name, "${var.instance_name}-sec")
+  secondary_kms_key_name  = try(var.encryption_config.secondary_kms_key_name, null)
+  # secondary instance type is aligned with cluster type unless apply is targeting a promotion, in that
+  # case cluster will be 'primary' while instance still 'secondary'.
   secondary_instance_type = try(
     var.cross_region_replication.promote_secondary && google_alloydb_cluster.secondary[0].cluster_type == "SECONDARY"
     ? "SECONDARY"
     : google_alloydb_cluster.secondary[0].cluster_type, null
   )
-  users = {
-    for k, v in coalesce(var.users, {}) :
-    k => {
-      name     = k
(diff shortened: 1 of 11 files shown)


juliocc (commented):
@viliampucik thanks a lot for this. Very nice PR. We'll need a some time to go over it.

cc @simonebruzzechesse

modules/alloydb/versions.tf
```diff
@@ -25,6 +25,10 @@ terraform {
       source  = "hashicorp/google-beta"
       version = ">= 6.33.0, < 7.0.0" # tftest
     }
+    random = {
+      source  = "hashicorp/random"
+      version = ">= 2.2.0" # tftest
+    }
```
juliocc, line 31: Don't make changes to individual versions.tf

modules/alloydb/main.tf
```diff
@@ -17,42 +17,60 @@
...
+  deletion_policy                  = var.deletion_policy
+  display_name                     = coalesce(var.cluster_display_name, local.primary_cluster_name)
+  labels                           = var.labels
+  location                         = var.location
+  skip_await_major_version_upgrade = var.skip_await_major_version_upgrade
+  subscription_type                = var.subscription_type
+
+  # network_config block should exist only when PSA VPC resource link is present to prevent Terraform state drift
+  dynamic "network_config" {
+    for_each = try(var.network_config.psa_config.network, null) != null ? [""] : []
```
juliocc, line 69: no need for try here. Just check `var.network_config.psa_config != null`

modules/alloydb/main.tf
```diff
@@ -407,31 +425,96 @@ resource "google_alloydb_instance" "secondary" {
...
 
 resource "google_alloydb_user" "users" {
-  for_each       = local.users
+  for_each       = var.users
   cluster        = google_alloydb_cluster.primary.id
-  user_id        = each.value.name
+  user_id        = each.key
   user_type      = each.value.type
-  password       = each.value.password
+  password       = each.value.type == "ALLOYDB_BUILT_IN" && each.value.password == null ? random_password.passwords[each.key].result : each.value.password
```
juliocc, line 517: Split in multiple lines

modules/alloydb/main.tf
```diff
@@ -407,31 +425,96 @@ resource "google_alloydb_instance" "secondary" {
...
+    if v.type == "ALLOYDB_BUILT_IN" && v.password == null
+  }
   length  = 16
   special = true
 }
 
 resource "google_alloydb_user" "users" {
-  for_each       = local.users
+  for_each       = var.users
   cluster        = google_alloydb_cluster.primary.id
```
juliocc, line 514: Should this be done in the secondary instances too?

modules/alloydb/main.tf
```diff
@@ -17,42 +17,60 @@
 locals {
   prefix      = var.prefix == null ? "" : "${var.prefix}-"
   is_regional = var.availability_type == "REGIONAL"
-  # secondary instance type is aligned with cluster type unless apply is targeting a promotion, in that
-  # case cluster will be 'primary' while instance still 'secondary'.
+  flags       = try(var.network_config.psa_config.enable_outbound_public_ip, false) ? merge(var.flags, { "password.enforce_complexity" = "on" }) : var.flags # Outbound Public IP requires password complexity to be enabled.
```
juliocc, line 20: Reformat as follows.
```terraform
  # Outbound Public IP requires password complexity to be enabled.
  flags = (
    try(var.network_config.psa_config.enable_outbound_public_ip, false) 
    ? merge(var.flags, { "password.enforce_complexity" = "on" }) 
    : var.flags 
  )
```

Does the API warn you about this? If yes, then the user should be the one taking care of this.

modules/alloydb/main.tf
```diff
@@ -17,42 +17,60 @@
...
   )
-  users = {
-    for k, v in coalesce(var.users, {}) :
-    k => {
-      name     = k
-      password = try(v.type, "ALLOYDB_BUILT_IN") == "ALLOYDB_BUILT_IN" ? try(random_password.passwords[k].result, v.password) : null
-      roles    = v.roles
-      type     = coalesce(v.type, "ALLOYDB_BUILT_IN")
-    }
+  secondary_machine_type = try(var.cross_region_replication.secondary_machine_config.machine_type, null) != null ? var.cross_region_replication.secondary_machine_config.machine_type : var.machine_config.machine_type
```
juliocc, line 44: split

juliocc (approved):
This LGTM. Left a few minor comments but otherwise I'm OK with this.

modules/alloydb/main.tf
viliampucik: My assumption is that the secondary cluster/instance is read-only replica, which asynchronously copies all the data including users from the primary, hence user management in the primary cluster is enough. However, I cannot fully test this assumption. Please, could you advise?

modules/alloydb/main.tf
viliampucik: API does want about this. I will remove auto-setting of this flag.

modules/alloydb/main.tf
juliocc: Makes sense

juliocc (approved):

PR #3062, opened 2025-05-01
VPC SC module refactor

juliocc:
- Remove support for bridges
- Allow project ids and convert them to project numbers where necessary
- Preserve order of directional policies

TODO:
- [x] Update README
- [x] Fix failing tests

---
**Checklist**

**Breaking Changes**

```upgrade-note
`modules/vpc-sc`: perimeter bridge are no longer supported. Please migrate to directional policies (ingress/egress rules) for more granular and secure perimeter configurations.
`modules/vpc-sc`: `service_perimeters_regular` renamed to `perimeters`

```


modules/vpc-sc/main.tf
```diff
@@ -19,6 +19,35 @@ locals {
...
+          try(v.from.resources),
+          try(v.to.resources)
+        ]
+      ],
+    ]
+  ]))
+  _project_ids = toset([
+    for x in local._all_project_identifiers :
+    x
+    if can(regex("^projects/[a-z]+", x))
```
wiktorn, line 45:
```suggestion
    if can(regex("^projects/[a-z]", x))
```
Is enough and more explicit about the intent.

modules/vpc-sc/main.tf
```diff
@@ -19,6 +19,35 @@ locals {
...
+  ])
+  project_number = {
+    for k, v in data.google_project.project :
```
ludoo, line 48: this should be optional and gated behind a var/var attribute, so that we can skip the data source for a) tests, b) situations where we don't want it to be triggered

juliocc: Yeah, that will come if I manage to get around dynamic values. TBH, I might end up removing the option to convert project ids to numbers

ludoo (approved):
very nice, came out great!

PR #3083, opened 2025-05-14
Remove default values for access_config.ip_config for gke cluster modules

jaiakt:
This ensures the [dynamic master_authorized_networks_config block](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/e937a562c78cb060ebb63ed0fdae5e8b008d7454/modules/gke-cluster-standard/main.tf#L395) is properly conditional on user input.

Previously, these fields would never be null due to their default values, causing the block to always be initialized even when users didn't intend to configure authorized networks.

---
**Checklist**


modules/gke-cluster-standard/variables.tf
```diff
@@ -19,9 +19,9 @@ variable "access_config" {
   type = object({
     dns_access = optional(bool, true)
     ip_access = optional(object({
-      authorized_ranges               = optional(map(string), {})
+      authorized_ranges               = optional(map(string), null)
```
juliocc, line 22: Null is redundant here. Leave only `optional(map(string))`.

Same for the other field

juliocc (changes requested):

modules/gke-cluster-standard/variables.tf
jaiakt: Awesome wasn't aware of that.  Updated!

jaiakt:
Cannot update the README on my work laptop (due to not being able to pull arbitrary pip packages).  Will do later today on my personal laptop.

juliocc:
Looking more closely at this PR, I think the original defaults are sensible. Can you provide more details on the rationale for this change?

jaiakt:
Yeah of course.  If you take a look at https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/e937a562c78cb060ebb63ed0fdae5e8b008d7454/modules/gke-cluster-standard/main.tf#L395, this for_each does not resolve to false even if `var.access_config` is left completely unset.  The defaults for the subfields resolve to non-null so we end up creating a `master_authorized_networks_config` resource which requires authorized networks for connectivity.  I feel like the intended behavior is if `var.access_config` is unset, we would want to not create a `master_authorized_networks_config` resource which then allows unauthorized networks to hit the cluster endpoint.

I think there are two ways of solving this.  Either we default the subfields to null like I did in this PR or we update the resource logic to check for `{}` and `false` rather than `null`.  The empty dictionary sounds reasonable but I'm not familiar enough with this codebase enough to say if the user explicitly passing `false` is the same as the user leaving out the boolean field.

ludoo:
Given the complexity of the GKE API and the amount of different cluster configurations, can you provide an actual example of using the module that requires this change, so we can properly test this PR?

juliocc:
@jaiakt I think understand the issue now and you're correct that the current type would not allow the creation of a cluster without authorized ranges.

Let me do a few tests and I'll get back to you

jaiakt:
Yeah.  The behavior we saw with an older version of `cloud-foundation-fabric` (specifically `v34.1.0`), when we unset `access_config` we believe the entire `master_authorized_networks_config` block was commented out and the behavior we experienced was ip access without authentication.  Then we use IAM and kube policies to control who is allowed to take particular actions on the cluster.  When using the `kubernetes` provider, our provider setup looks just like https://stackoverflow.com/a/75198973 which also connects directly to the ip address of the gke control plane.

When we bumped the `cloud-foundation-fabric` version to `v36.2.0` due to needing some other features, terraform was unable to connect directly to the ip and we weren't able to connect to the ip from our machines as well because now `master_authorized_networks_config` was being created with no way to stop it.  Our only solution to fix this was to fork `cloud-foundation-fabric` and apply the change in this PR which got us back to the old behavior we were experiencing.  Let me know if this all makes sense and sounds reasonable.  I've been a bit busy to write tests/readme changes here but if this all sounds reasonable, I'll fix tests/documentation for this change.

juliocc:
I'll do a few more tests with v34.1.0 to check what was happening back then. In any case, I think the current behavior is reasonable and probably more secure now.

In the meantime, can you try adding an authorized range for 0.0.0.0/0 and check if it gives you what you want?

PR #3116, opened 2025-05-28
Add support for binary authorization policy to cloud function v2 module

### LINKED ISSUE #3076: [FR]: Binary Authorization block in cloud run functions v2 module
### The feature request

The terraform module that the cloud run functions v2 uses has a optional binary_authorization_policy variable that can be set in the service config but this is not surfaced as a variable in cff's module.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions2_function

### Proposed solution

binary_authorization_policy is accessible as a variable within the cff's cloud run functions v2 variables.tf and main.tf

### Additional context

_No response_

ludoo:
Fixes #3076


modules/cloud-function-v2/variables.tf
```diff
@@ -104,12 +104,13 @@ variable "environment_variables" {
...
-    cpu             = optional(string, "0.166")
-    runtime         = optional(string, "python310")
-    timeout_seconds = optional(number, 180)
+    binary_authorization_policy = optional(string)
+    entry_point                 = optional(string, "main")
+    instance_count              = optional(number, 1)
+    memory_mb                   = optional(number, 256) # Memory in MB
+    cpu                         = optional(string, "0.166")
+    runtime                     = optional(string, "python310")
+    timeout_seconds             = optional(number, 180)
```
juliocc, line 113: Should we remove all the defaults and let the provider/api pick them?

juliocc (approved):
one small comment but feel free to ignore.

LGTM

PR #3129, opened 2025-06-01
Cloud Run with IAP recipe

apichick:
---
**Checklist**


modules/net-lb-app-ext/recipe-cloud-run-iap/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+```hcl
+module "test" {
+  source     = "./fabric/modules/net-lb-app-ext/recipe-cloud-run-iap"
+  project_id = "project-1"
+  _testing = {
+    name   = "project-1"
+    number = 1234567890
+  }
+  region = "europ-west1"
```
ludoo: `europe`

ludoo (commented):
Miren, we should avoid using JS code and external libraries when possible, as they will inevitably generate security alerts when vulnerabities are found in the deps.

Is this the only way of doing what you need?

modules/net-lb-app-ext/recipe-cloud-run-iap/README.md
apichick: ok, I'll see if i can do it in bash

ludoo (approved):

PR #3130, opened 2025-06-03
Conditionally creates billing sink if the billing account is of type resource

Ali-Aburub:
Conditionally creates billing sink if the billing account is of type resource

---
**Checklist**

### DIFF
--- a/tests/fast/stages/s0_bootstrap/external_billing_account.tfvars
+++ b/tests/fast/stages/s0_bootstrap/external_billing_account.tfvars
@@ -1,10 +1,10 @@
 billing_account = {
-  id = "000000-111111-222222"
+  id           = "000000-111111-222222"
   is_org_level = false
   force_create = {
-     dataset    = true
-      project    = true
-      log_bucket = true
+    dataset    = true
+    project    = true
+    log_bucket = true
   }
 }
 essential_contacts = "gcp-organization-admins@fast.example.com"


fast/stages/0-bootstrap/billing.tf
```diff
@@ -18,16 +18,21 @@
...
-    local.principals.gcp-billing-admins,
-    local.principals.gcp-organization-admins,
-    module.automation-tf-bootstrap-sa.iam_email,
-    module.automation-tf-resman-sa.iam_email
-  ]
-  billing_ext_viewers = [
-    module.automation-tf-bootstrap-r-sa.iam_email,
-    module.automation-tf-resman-r-sa.iam_email
-  ]
+  billing_iam = {
```
ludoo: we can skip using the local and put this directly below in the module call

fast/stages/0-bootstrap/billing.tf
```diff
@@ -81,20 +86,28 @@ module "billing-export-dataset" {
...
 
-resource "google_billing_account_iam_member" "billing_ext_viewer" {
-  for_each = toset(
-    local.billing_mode == "resource" ? local.billing_ext_viewers : []
-  )
-  billing_account_id = var.billing_account.id
-  role               = "roles/billing.viewer"
-  member             = each.key
-}
+module "billing-account-log-sink" {
```
ludoo: I would rename the module as `billing-account`

fast/stages/0-bootstrap/billing.tf
Ali-Aburub: Moved to module call directly

fast/stages/0-bootstrap/billing.tf
Ali-Aburub: Done

fast/stages/0-bootstrap/billing.tf
```diff
@@ -81,20 +70,42 @@ module "billing-export-dataset" {
...
-resource "google_billing_account_iam_member" "billing_ext_admin" {
-  for_each = toset(
-    local.billing_mode == "resource" ? local.billing_ext_admins : []
-  )
-  billing_account_id = var.billing_account.id
-  role               = "roles/billing.admin"
-  member             = each.key
+module "billing-account-logbucket" {
+  source        = "../../../modules/logging-bucket"
+  count         = local.billing_mode == "resource" ? 1 : 0
```
ludoo: Ali, sorry for forgetting about this. Can we make log bucket creation explicit? I would add a `log_bucket` attribute to `var.billing_account.force_create` and combine that in this condition.

ludoo:
> > One more comment: we should have a `moved` block to transition IAM from the resource to the module, if possible.
> 
> Currently the billing IAM assignments are additive. With the refactoring I did, I changed it to be authoritative since these roles will not be touched on later stages. Do you agree with this change? @ludoo

actually no, they should be additive as the assumption is an external billing account is not exclusively managed here

fast/stages/0-bootstrap/billing.tf
```diff
@@ -81,20 +70,42 @@ module "billing-export-dataset" {
...
-  )
-  billing_account_id = var.billing_account.id
-  role               = "roles/billing.viewer"
-  member             = each.key
-}
+module "billing-account" {
+  source = "../../../modules/billing-account"
+  count  = local.billing_mode == "resource" ? 1 : 0
+  id     = var.billing_account.id
+  iam = {
```
ludoo: this should be changed to additive

ludoo (approved):

Issue #3123, opened 2025-05-29
Module cloudsql-instance:  random_password.passwords resource is being created for users with "CLOUD_IAM_SERVICE_ACCOUNT" type


digitalake:
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

`random_password.passwords` resource is being created for users with "CLOUD_IAM_SERVICE_ACCOUNT" type but not used.

### Release version

v40.1.0

### Terraform version

v1.11.4

### Steps to reproduce

Use such `users` input value:

```
users = {
  "example@example.iam.gserviceaccount.com" = {
    type = "CLOUD_IAM_SERVICE_ACCOUNT"
  }
}
```

### Expected behavior

Let's say we are using `cloudsql_instance` name when calling the module and running terraform plan:

```
module.cloudsql_instance.google_sql_user.users["example@example.iam.gserviceaccount.com"] will be created
...
```

### Actual behavior

Let's say we are using `cloudsql_instance` name when calling the module and running terraform plan:

```
module.cloudsql_instance.google_sql_user.users["example@example.iam.gserviceaccount.com"] will be created
...
module.cloudsql_instance.random_password.passwords["example@example.iam.gserviceaccount.com"] will be created
...
```

### Additional context

The user settings are fine (including password), so the user itself is created with a set of expected parameters. At the same time `random_password.passwords` resource is being created and not used at all. The only way to prevent `random_password.passwords` resource creation is setting up a mock password value so that user key is not added to the input set for: 

```
resource "random_password" "passwords" {
  for_each = toset([
    for k, v in coalesce(var.users, {}) :
    k
    if v.password == null
  ])
  length      = try(var.password_validation_policy.min_length, 16)
  special     = true
  min_lower   = 1
  min_numeric = 1
  min_special = 1
  min_upper   = 1
}
```

like this:

```
users = {
  "example@example.iam.gserviceaccount.com" = {
    type = "CLOUD_IAM_SERVICE_ACCOUNT"
    password = "mock"
  }
}
```

PR #3135, opened 2025-06-05
CloudSQL - Create password resource only when needed

wiktorn:
* create random password resource only when it will be used

Closes: #3123

---
**Checklist**


modules/cloudsql-instance/main.tf
```diff
@@ -287,7 +289,7 @@ resource "random_password" "passwords" {
   for_each = toset([
     for k, v in coalesce(var.users, {}) :
     k
-    if v.password == null
+    if v.password == null && coalesce(v.type, "BUILT_IN") == "BUILT_IN"
   ])
```
juliocc, line 291: nit: You can probably simplify all these coalesce calls making BUILT_IN the default, and then while variable non nullable

PR #3149, opened 2025-06-10
Add support for DNS zones to Apigee module

apichick:
---
**Checklist**


modules/apigee/variables.tf
```diff
@@ -26,6 +26,18 @@ variable "addons_config" {
   default = null
 }
 
+variable "dns_zones" {
+  description = "DNS zones."
+  type = map(object({
+    domain            = string
```
ludoo, line 32: Would it make sense to use the domain as key? It would also ensure a domain maps to only one zone.

apichick: I would not do that. Some customers implement naming conventions for resources, by doing so you are preventing them for applying their naming conventions to the name of the managed DNS zone, as you would choose the name for them.

ludoo (approved):

PR #3066, opened 2025-05-02
New FAST data platform

ludoo:
This is now feature complete for the MVP.

The README needs more work and examples, but that will be done incrementally by the Data people after merging.

PR #3156, opened 2025-06-11
Allow to directly specify service agents for CMEK in project module (Composer v2 support)

jnahelou:
The recent change to the CMEK service agents mapping for composer.googleapis.com in PR #3066 breaks backward compatibility for Cloud Composer v2 environments.

The mapping was changed from:
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/c0bf32e797b7de9cb8ac1d161c7f1ab4432acbc2/modules/project/cmek.tf#L30-L33
To:
https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/e176f49a84ea6df43dacf2f9fc42625ecb7e0e08/modules/project/cmek.tf#L32

While this change aligns with Composer v3 requirements, it prevents users from enabling the necessary dependencies for Composer v2 clusters to function properly with CMEK encryption. 

Adding an explicit usage of dependencies in project manifests could works but "container-engine" ([doc](https://cloud.google.com/composer/docs/composer-2/configure-cmek-encryption#grant-roles-permissions) can not be set as it is neither a google service nor a container.googleapis dependency.

https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/e176f49a84ea6df43dacf2f9fc42625ecb7e0e08/modules/project/cmek.tf#L34

### Expected Behavior

The module should support both Composer v2 and v3 CMEK requirements, or provide a fallback mechanism for users to specify additional service agents when no explicit mapping is found,.

### DIFF
--- a/modules/project/README.md
+++ b/modules/project/README.md
@@ -838,6 +838,59 @@ module "kms" {
 # tftest modules=2 resources=10 e2e
 ```
 
+Services like Composer, Dataflow, and Datafusion require service agent dependencies from other services to function properly with CMEK encryption. These dependencies are automatically resolved based on a predefined mapping that follows the latest service requirements.
+
+In situations where the predefined mapping doesn't cover your specific use case (such as using older service versions or custom configurations), you can extend this mapping by explicitly declaring additional dependencies.
+
+The `service_encryption_key_ids` parameter accepts keys declared using either:
+
+- [Service Agents](#service-agents) API names (e.g., composer.googleapis.com)
+- [Service Agent aliases](#service-agent-aliases) (e.g., container-engine-robot)
+
+The following examples demonstrate how to configure CMEK encryption for different Composer versions:
+
+For composer v3:
+
+```
+module "project" {
+  source          = "./fabric/modules/project"
+  billing_account = var.billing_account_id
+  name            = "project"
+  prefix          = var.prefix
+  parent          = var.folder_id
+  services = [
+    "composer.googleapis.com",
+  ]
+  service_encryption_key_ids = {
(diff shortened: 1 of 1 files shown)


ludoo:
I don't think we can support both TBH. My suggestion would be to stay on the older release where you need an older composer version, or clone and change the latest version to suit your needs.

I'll leave it to @juliocc to figure out if the PR is ok, as that part of the code is his. :)

jnahelou:
With this change, users could handle mixed environments by explicitly configuring the required services:

For Composer v2:

```yaml
kms_service_agents:
  composer.googleapis.com:
    - <key>
  # Composer 2 dependencies
  artifactregistry.googleapis.com:
    - <key>
  container-engine-robot:
    - <key>
  container.googleapis.com:
    - <key>
  pubsub.googleapis.com:
    - <key>
 ```

For Composer v3:

```yaml
kms_service_agents:
  composer.googleapis.com:
    - <key>
 ```

This approach allows users to maintain both Composer versions in **the same project factory** while properly configuring CMEK encryption for each.

ludoo:
I think I get your change, and kudos for submitting this. Basically the last resort if a mapping from service to agent is not found, is to use the mapping key as the service agent. Can you please document it in the README under the service agent section? This is good for me, unless @juliocc disagrees.

juliocc:
A bit late to the party, but here are my thoughts. We consciously held off on supporting Composer v2 until a user explicitly asked for it. Congratulations, @jnahelou, you're the first!

I've thought about different ways to tackle this and never found a good solution. This PR is elegant, short, and fixes the problem perfectly.

My only ask is, could we provide an example in the project module showing how it works? The examples from the comments above would be sufficient.

PR #3163, opened 2025-06-14
Allow custom roles in context, add support for shared VPC IAM to project and project factory

ludoo:
This PR allows implementing a specific use case: allowing an automation service account created by the project factory to [attach a DNS zone to a Shared VPC host project](https://cloud.google.com/dns/docs/zones/cross-project-binding). This is needed is specific situations like installing OpenShift.

The pieces required to make this work are implemented in this PR:

- in the `project` module implement support for arbitrary IAM additive bindings on the Shared VPC host via `var.shared_vpc_service_config.iam_bindings_additive`
- in the `project-factory` module implement support for `shared_vpc_service_config.iam_bindings_additive` and add context replacement for custom roles, supported by `var.factories_config.context.custom_roles`
- in the FAST resource management stage add support for custom role expansion to IAM conditions in stages

The end result allows defining a condition like this in the networking stage 2 definition:

```yaml
  iam_bindings:
    project_factory:
      role: roles/resourcemanager.projectIamAdmin
      members:
        - project-factory-rw
      condition:
        title: Project factory delegated IAM grant.
        expression: |
          api.getAttribute('iam.googleapis.com/modifiedGrantsByRole', []).hasOnly([
            'roles/compute.networkUser', 'roles/composer.sharedVpcAgent',
            'roles/container.hostServiceAgentUser', 'roles/vpcaccess.user',
            '${custom_roles.dns_zone_binder}'
          ])
```

Which then allows the project factory to set the custom role on a Shared VPC host at the project level:

```yaml
shared_vpc_service_config:
  host_project: dev-spoke-0
  iam_bindings_additive:
    dns_zone_binder:
      member: ocp-install
      role: dns_zone_binder
  network_users:
    - automation/rw
    - ocp-install
    - ocp-master
```

A new DNS zone binder role has also been added to the FAST bootstrap stage.


juliocc:
very nice

PR #3166, opened 2025-06-17
add CI/CD automation support to project-factory projects

wolfie82:
### **Description**

This PR refactors and enhances the `project-factory` module to provide an automated CI/CD configuration path. The core goal is to enable secure, keyless, and automated deployments for vended projects, using a robust pattern centered around Workload Identity Federation (WIF) and a centralized "Controlling Project."

This change allows teams to define their CI/CD and automation resource needs declaratively within their project's YAML configuration, with a clear separation between user-defined variables and module-provided globals.

### **Key Changes**

* **`outputs_bucket` Configuration**: The `automation.outputs_bucket` supports either the creation of a new bucket via a `create_new` object or the use of a pre-existing bucket by specifying its `name`. This allows for better integration with existing infrastructure.

* **Implicit Provider File Generation**: The module uses the `cicd_config.impersonations` map as the **single source of truth** to automatically generate the necessary provider configuration files, which prevents configuration drift and reduces user error. A custom provider template can be specified via the optional `templates.provider` key.

* **Flexible and Explicit Workflow Templates**:
    * **Flexible Workflow Map**: `templates.workflow` uses flexible map, allowing users to define multiple, uniquely named CI/CD pipelines (e.g., `plan-and-apply`, `nightly-tests`).
    * **Namespaced Template Variables**: To prevent name collisions and improve clarity, variables injected into templates are namespaced. User-defined values are available under `vars.*`, while module-provided values are in a `global.*` namespace.
    * **Role-Keyed Provider Map**: The `global.provider_file` map available in templates is keyed by the target IAM **role** (e.g., `.ro`, `.rw`). This makes the workflow template's choice of provider file explicit and self-documenting, removing any "magic" from the module's implementation.

* **Decoupled Local File Generation**: The creation of local artifacts via the `outputs_location` variable is decoupled from GCS bucket creation, allowing developers to generate and test templates locally without needing to provision cloud resources.

### **Documentation**

The `modules/project-factory/README.md` and `fast/stages/2-project-factory/README.md` have been updated to:
* Provide a comprehensive example and detailed explanation of the enhanced `automation` block, including the `cicd_config`, `outputs_bucket`, and `templates` attributes.
* Add a new `Template Variables` section to the module README, fully documenting the `global` and `vars` namespaces available to template authors.
* Add new bootstrapping/workflow section to fast README describing a practical flow `plan on PR / apply on merge` CI/CD process.

This is related to https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/discussions/2775

---
**Checklist**


wolfie82:
I fixed the linting error, hopefully all checks should pass now when they run.

modules/project-factory/schemas/project.schema.json
```diff
@@ -11,15 +11,44 @@
         "project"
       ],
       "properties": {
-        "prefix": {
-          "type": "string"
+        "cicd_config": {
+          "type": "array",
```
ludoo: Why do you need an array here? What we typically do is map a single pipeline to a FAST stage or (in this case) a project. IMHO that should be enough, and also reduce the complexity both in code and usage.

modules/project-factory/variables.tf
```diff
@@ -243,6 +243,12 @@ variable "factories_data" {
     projects = optional(map(object({
       automation = optional(object({
         project = string
+        cicd_config = optional(list(object({
```
ludoo: Same comment I dropped in the schema: I would only support one CI/CD configuration here, like we do with FAST.

modules/project-factory/variables.tf
```diff
@@ -299,6 +317,22 @@ variable "factories_data" {
           iam_sa_roles           = optional(map(list(string)), {})
           iam_storage_roles      = optional(map(list(string)), {})
         })), {})
+        templates = optional(list(object({
```
ludoo: One CI/CD config maps to one template, so this too should not be a list.

modules/project-factory/variables.tf
```diff
@@ -408,3 +442,15 @@ variable "factories_data" {
   nullable = false
   default  = {}
 }
+
+variable "federated_identity_pool" {
```
ludoo: Here we should conform to the format we use in FAST, we have expose a map of federated identies. We don't need two variables one for the pool one for the provider. But I can see situations where a project factory manages projects that use different providers or pools.

ludoo (changes requested):
I onl checked the variables/schema changes, and before looking at the code I think we should agree on how many CI/CD configs per project are supported. IMHO there's no need to support more than one, which is what we do in FAST. While for WIF pools we should instead support an arbitrary number, and those should also be passed in via context and be available for substitutions.

wolfie82:
> I onl checked the variables/schema changes, and before looking at the code I think we should agree on how many CI/CD configs per project are supported. IMHO there's no need to support more than one, which is what we do in FAST. While for WIF pools we should instead support an arbitrary number, and those should also be passed in via context and be available for substitutions.

Thanks for the feedback @ludoo!

My initial thinking for making `cicd_config` a list was to provide maximum flexibility for different team structures and deployment patterns that we've seen in various environments. While a single config per project is definitely the most common and simple use case, the list structure was intended to natively support a few more advanced scenarios without requiring workarounds.

Here are the specific use cases I had in mind:

1.  **Multi-Repository Deployments**: When a single GCP project (e.g., a shared application environment) has its resources managed by IaC from multiple repositories. For example, a main application repo and a separate, centralized database management repo both needing to deploy to the same project.

    ```yaml
    # A single project being deployed to from two different repos
    cicd_config:
      - workload_identity_provider: github-main
        repository: my-org/app-repo
        impersonations: { cicd-rw: app-rw }
      - workload_identity_provider: github-main
        repository: my-org/database-repo
        impersonations: { cicd-rw: db-rw }
    ```

2.  **Multi-Provider Deployments**: During a migration (e.g., from GitLab to GitHub) or in organizations that use multiple SCM tools, both systems might temporarily need deployment access to the same project.

    ```yaml
    # A single project being deployed to from both GitHub and GitLab
    cicd_config:
      - workload_identity_provider: github-provider
        repository: my-org/app-repo
        impersonations: { cicd-rw: rw }
      - workload_identity_provider: gitlab-provider
        repository: my-group/legacy-job
        impersonations: { cicd-rw: rw }
    ```

3.  **Complex Branching Strategies**: Allowing a single project to be deployed from multiple branches in the same repository, which can be useful for pre-production or staging environments that test both `main` and `release-candidate` branches.

    ```yaml
    # A single project that is restricted to two different branches
    cicd_config:
      - workload_identity_provider: github-provider
        repository: my-org/my-app
        branch: main
        impersonations:
          cicd-rw: rw
      - workload_identity_provider: github-provider
        repository: my-org/my-app
        branch: release-candidate
        impersonations:
          cicd-rw: rw
    ```

I completely agree that for the standard FAST pattern, a single configuration is simpler and easier to manage. My goal was to treat that as the default while still allowing for these more complex scenarios if a team needed them.

Given your feedback, I'm happy to proceed in whichever direction you think is best for the project long-term. We can either:

1.  Keep the flexible list implementation and I can add documentation to clarify these advanced use cases.
2.  Refactor to a single object to strictly enforce the "one-pipeline-per-project" model, which I can do right away.

What are your thoughts on this trade-off? Let me know which path you prefer, and I'll update the PR accordingly.

Thanks!

ludoo:
Thanks for the very detailed reply, and not getting discouraged. :)

My personal preference is to limit the complexity, especially in an initial release. Coming out withh a simpler solution allows us to cover a large number of the most common use cases, and then wait for experiences coming from actual use to iterate on additional features. Which is what we always try to, as designing for maximum flexibility often ends up being unsustainable.

I would clone the exact same variable types for `cicd_config` and `federated_identity_providers` we use in FAST to start with, which brings the added benefit of making the new functionality familiar and instinctive to experienced users. And try to tackle any context replacement needs right from the start, by adding the relevant attributes to `factories_config.contexts`. Ideally this should be a drop-in in FAST, allowing the project factory to consume resources passed down from the previous stages, like we already do for VPCs/subnets/VPC SC perimeters.

wolfie82:
Thank you for the guidance @ludoo, that makes sense to me.  I will work on these refinements today.

modules/project-factory/schemas/project.schema.json
wolfie82: Updated

modules/project-factory/variables.tf
wolfie82: Updated

modules/project-factory/variables.tf
wolfie82: Updated

modules/project-factory/variables.tf
wolfie82: This is now passed down via context and we need the pool name to construct the `sa_wif_binding`.

wolfie82:
@ludoo I’ve pushed the latest updates based on your guidance. Let me know if anything needs adjusting.

modules/project-factory/variables.tf
```diff
@@ -299,6 +324,19 @@ variable "factories_data" {
...
+        templates = optional(object({
+          workflow = optional(object({
+            plan = object({
+              service_account = string
+            })
+            apply = object({
+              service_account = string
+            })
+          }))
+          provider_files = optional(list(object({
```
ludoo: Thinking about FAST here, where we're already generating the provider files and storing them in the FAST-level output bucket/location: I would remove this from the variables and always generate provider files if the output bucket is present, or an existing bucket name is passed in. FAST would then simply instruct the project factory to use the provided bucket for the files. There should also be an option to generate them locally.

modules/project-factory/variables.tf
```diff
@@ -260,6 +273,18 @@ variable "factories_data" {
               description = optional(string)
             }))
           })), {})
+          outputs_bucket = optional(object({
```
ludoo, line 287: I would change this to be either a bucket name, or bucket details for a new bucket. And add validation to ensure only one of the two is passed.

modules/project-factory/variables.tf
ludoo: Addendum: I would keep a single `templates.provider`, and use a default one if nothing is passed in and an output bucket or bucket details are present.

For workflows, I would make it a map and not an object, with `workflow_file` and `service_account` attributes for each element.

ludoo (changes requested):
Some nits on the variables interface, before tackling the actual code.

modules/project-factory/README.md
```diff
@@ -98,13 +98,16 @@ buckets:
...
+Other than creating automation resources within the project via the `service_accounts` and `buckets` attributes, this module also support management of automation resources created in a separate controlling project. This allows granting broad roles on the project, while still making sure that the automation resources used for Terraform cannot be manipulated from the same identities.
 
 Automation resources are defined via the `automation` attribute in project configurations, which supports:
 
 - a mandatory `project` attribute to define the external controlling project; this attribute does not support interpolation and needs to be explicit
 - an optional `service_accounts` list where each element defines a service account in the controlling project
 - an optional `bucket` which defines a bucket in the controlling project, and the map of roles/principals in the corresponding value assigned on the created bucket; principals can refer to the created service accounts by key
+- an optional `cicd_config` which defines the CI/CD automation configurations for the project that configures Workload Identity Federation (WIF) and impersonation roles
```
wiktorn:
```suggestion
- an optional `cicd_config` which defines the CI/CD automation configurations for the project that configures Workload Identity Federation (WIF) and grants impersonation permission
```

modules/project-factory/automation.tf
```diff
@@ -158,3 +249,23 @@ module "automation-service-accounts" {
   # we don't interpolate buckets here as we can't use a dynamic key
   iam_storage_roles = lookup(each.value, "iam_storage_roles", {})
 }
+
+module "automation_sa_impersonation" {
```
wiktorn:
```suggestion
module "automation-sa-impersonation" {
```

modules/project-factory/variables.tf
wolfie82: I've updated this.  Please take a look and let me know what you think.

wolfie82:
@ludoo Here is an update below.  Once we settle, I will go through the documentation to make sure its reflective of the design.  I paused on that because it was a lot to iterate and update the docs at the same time with all the changes.

Here’s a high-level summary of the improvements:

* **Logic Encapsulation:** All CI/CD artifact generation (workflows, provider files) has been moved from the FAST stage into the module itself. The module is now a self-contained.
* **Flexible Configuration:** The rigid object structures for workflows and providers have been replaced with flexible, data-driven maps and optional overrides.
* **Improved Clarity & Safety:** The final interface is more explicit about its intent and includes validation to prevent configuration mistakes.

**1. `outputs_bucket` Configuration**

* The `automation.outputs_bucket` attribute now accepts either a `name` for a pre-existing bucket or a `create_new` object to have the module provision a new one. This directly implements your feedback, allowing the module to be used seamlessly in environments where core infrastructure like CI/CD buckets are managed centrally.

**2. Provider File Generation**

I removed the explicit `templates.provider_files` list. The module now **implicitly** generates a provider file for every service account defined in the `cicd_config.impersonations` map.

The `impersonations` map is the single source of truth, eliminating the risk of configuration drift where the two lists could become out of sync. I also added an optional `templates.provider` key that allows a user to specify a custom provider template, which defaults to `providers.tf.tpl` in the stage's `templates` directory.

**3. Workflow Configuration (The `templates.workflow` map)**

This saw the most significant evolution based on your feedback.

I replaced the rigid `plan`/`apply` object with a flexible map, as you suggested. Each key in the map is a user-defined workflow name (e.g., `plan-and-apply`), which points to a configuration object.

Why I defined the interface this way...

**Namespaced Template Variables:** To ensure clarity and safety, variables passed to the template are namespaced.
        * `global.*`: For variables provided by the module (`identity_provider`, `outputs_bucket`, `audiences`.).
        * `vars.*`: For all user-defined variables from the YAML `vars:` block.
        This prevents user variables from ever colliding with module-generated ones and makes the template's data sources explicit.

**Explicit Provider File Mapping:** The `global.provider_file` map passed to the template is now keyed by the **target impersonation** (e.g., `.ro`, `.rw`). This was a key decision to make the template's intent clear. The template author must now explicitly ask for the provider file they need for a given step, making the workflow's IAM requirements self-documenting. For example: `env.provider_file: ${global.provider_file.ro}`.

### Final Configuration Example

```yaml
# data/projects/dev-tb-0.yaml
parent: tb/dev
iam:
  roles/owner: [rw]
  roles/viewer: [ro]

automation:
  project: wolfcoded-tb-iac-0
  service_accounts:
    rw: {}
    ro: {}
    cicd-ro: {}
    cicd-rw: {}
  cicd_config:
    identity_provider: tb
    impersonations:
      cicd-ro: ro
      cicd-rw: rw
    repository: Tranquil-Beats/tranquil-beats
  outputs_bucket:
    create_new:
      description: CI/CD outputs bucket DEV
  templates:
    # Optional: Override the default provider template
    provider: "custom-providers.tf.tpl"
    # Define a CI/CD workflow
    workflow:
      plan-and-apply:
        template_path: "workflow-github.yaml"
        vars:
          plan_service_account: cicd-ro
          apply_service_account: cicd-rw
```

ludoo:
Can you fix the CLA?

wolfie82:
Added a bunch of tests covering file generation across multiple apps using different federated identity providers. I'll start work on refining the readme's.

wolfie82:
> Can you fix the CLA? There's commits from `Co-Author: <gi***b​@vink.pl>`

@ludoo Should I squash the commit into another one?  I accepted a change from @wiktorn via the Github UI since it was super simple. That's when this popped.

ludoo:
> > Can you fix the CLA? There's commits from `Co-Author: <gi***b​@vink.pl>`
> 
> @ludoo Should I squash the commit into another one? I accepted a change from @wiktorn via the Github UI since it was super simple. That's when this popped.

Let me try to override that. :)

ludoo:
Ok, that has been fixed. Give me some time to review this and run a few tests in my org, unfortunately these days are a bit hellish. :) Will try to get to it by tonight.

wolfie82:
> Ok, that has been fixed. Give me some time to review this and run a few tests in my org, unfortunately these days are a bit hellish. :) Will try to get to it by tonight.

No stress!  I'm working on updating the user readme examples and info.  However, you can look at the tests in `modules/project-factory/README.md`.

ludoo:
We brainstormed a bit with Wiktor on the interface and features, to figure out how much complexity can be safely added to the project factory while still respecting its main use case.

We don't have consensus yet but can throw in here a few things for discussion, so we can hopefully get to the point where this PR can be finished and merged. :)

First, we both feel that cicd service accounts and impersonation add quite a bit of complexity to the interface. I also kind of dislike it as it's an abstraction that deviates from the interface approach of the project factory. One way to solve this would be to explicitly managing impersonation via a service account's `iam` block, but that might fail unless we split resources for "regular" and CI/CD service accounts. A better way IMHO would be to either create CI/CD service accounts so they and the related impersonation roles don't need to be explicitly defined. What about something like this:

```hcl
service_accounts:
    rw:
      description: Read/write automation sa for app example 0.
    ro:
      description: Read-only automation sa for app example 0.
cicd_config:
  service_accounts_for:
    - rw
    - ro
```

This would generate the service accounts and assign the impersonation role. The added benefit is to enforce a naming convention for CI/CD service accounts, in the same way that FAST does.

The second thing we're discussing is output files: this again complicates the interface, and is not immediately obvious how to use it. My personal preference would be to avoid generating files in the project factory module (something no other module does BTW) and just return sets of variables ready to be consumed in templates. This would also remove the need to deal with output buckets. For FAST this would mean keeping this machinery in FAST itself instead of hiding it in a module, which I much strongly prefer. For non-FAST project factory consumers this would mean having to wrap the module to generate output files, which is simple enough and allows much greater flexibility vs having all the machinery in the module.

wolfie82:
> We brainstormed a bit with Wiktor on the interface and features, to figure out how much complexity can be safely added to the project factory while still respecting its main use case.
> 
> 
> 
> We don't have consensus yet but can throw in here a few things for discussion, so we can hopefully get to the point where this PR can be finished and merged. :)
> 
> 
> 
> First, we both feel that cicd service accounts and impersonation add quite a bit of complexity to the interface. I also kind of dislike it as it's an abstraction that deviates from the interface approach of the project factory. One way to solve this would be to explicitly managing impersonation via a service account's `iam` block, but that might fail unless we split resources for "regular" and CI/CD service accounts. A better way IMHO would be to either create CI/CD service accounts so they and the related impersonation roles don't need to be explicitly defined. What about something like this:
> 
> 
> 
> ```hcl
> 
> service_accounts:
> 
>     rw:
> 
>       description: Read/write automation sa for app example 0.
> 
>     ro:
> 
>       description: Read-only automation sa for app example 0.
> 
> cicd_config:
> 
>   service_accounts_for:
> 
>     - rw
> 
>     - ro
> 
> ```
> 
> 
> 
> This would generate the service accounts and assign the impersonation role. The added benefit is to enforce a naming convention for CI/CD service accounts, in the same way that FAST does.
> 
> 
> 
> The second thing we're discussing is output files: this again complicates the interface, and is not immediately obvious how to use it. My personal preference would be to avoid generating files in the project factory module (something no other module does BTW) and just return sets of variables ready to be consumed in templates. This would also remove the need to deal with output buckets. For FAST this would mean keeping this machinery in FAST itself instead of hiding it in a module, which I much strongly prefer. For non-FAST project factory consumers this would mean having to wrap the module to generate output files, which is simple enough and allows much greater flexibility vs having all the machinery in the module.

I really appreciate this @ludoo. The CI/CD service accounts and impersonation interface makes sense to me and will be really easy to change.

The output file generation moving to FAST also makes sense to me.  Funny enough it's where I originally had it, but that's what happens as we iterate :). For the direct consumers of the module, they could always look to the FAST output implementation as an example.

Once we have consensus, I'll be happy to make the changes.

ludoo:
It's a Sunday so we'll need to wait for @wiktorn to get offline tomorrow, but I would be inclined to go with the following. And I know your original PR was drastically trimmed down, I hope that's not discouraging: I much prefer a new feature to come in as a lean addition, and then we wait for actual usage feedback to come in. Especially on the project factory which is our most complex module.

```hcl
automation:
 project: foo-prod-iac-core-0
 service_accounts:
   rw:
     description: Read/write automation sa for app example 0.
   ro:
     description: Read-only automation sa for app example 0.
 cicd_config:
   identity_provider: github-public-sample
   repository:
     type: github
     name: my-org/my-repo
   service_accounts_for:
     rw:
       branch: main
	ro: {}
```

The identity providers passed in via variable will already contain the principal/principalset templates(`principal_branch` and principal_repo` attributes), so the above should allow to

- create one CI/CD "wrapper" service account for each service account where that makes sense
- assign the Wokload Identity User permission so they can impersonate their corresponding "operations" service account
- provide a set of output data for each project that allows module callers to easily generate workflow files from templates, following their own logic

Note that branch is not set in `cicd_config.repository` but is set at the service account level, which allows a lot more flexibility like having service accounts that merge to different branches (dev, prod, etc.).

Are we missing anything if we go with the above? And more importantly, does this make sense to everyone in this thread?

wiktorn:
Make sense!

wiktorn:
> Do we want to forgo the template generation, do it some other way, or have it implicitly generated if a 'cicd_config' is defined?

What I had in mind, is that in the projects-factory output value `projects` you can add `repository.type` as a hint to pick a proper template. Then the pipeline generation can work solely on the outputs of the project-factory module.

ludoo:
Closing this as it has gone stale. We're also refactoring our whole context replacement system in an upgraded project factory, currently under development in the `fast/fast-lite` branch, so any new changes should be implemented there.

PR #3173, opened 2025-06-18
Add support for IPv6 only subnets and IP collections

cmm-cisco:
Add support for IPv6 only subnets and IP collections for BYOIPv6.

---
**Checklist**

I had some problems getting the test harness to work.  I'm using this with a fairly complex configuration and it works.  Please let me know if something breaks.

**Breaking Changes**

None.

### DIFF
--- a/modules/net-vpc/README.md
+++ b/modules/net-vpc/README.md
@@ -712,6 +712,46 @@ module "vpc" {
 }
 # tftest modules=1 resources=6 inventory=ipv6.yaml e2e
 ```
+
+### IPv6-Only and IP Collections
+
+An IPv6-only subnetwork can be specified by setting `ipv6_only` to `true` and
+setting or `ip_cidr_range` to `null` or omitting it entirely.  An IP Collection
+may be specified with `ip_collection` and a
+[reference](https://cloud.google.com/compute/docs/reference/rest/v1/subnetworks/insert)
+to a collection source, like a PublicDelegatedPrefix (PDP) for BYOIPv6. The PDP
+must be a sub-PDP in `EXTERNAL_IPV6_SUBNETWORK_CREATION` mode.
+
+```hcl
+module "vpc" {
+  source     = "./fabric/modules/net-vpc"
+  project_id = var.project_id
+  name       = "my-network"
+  ipv6_config = {
+    enable_ula_internal = true
+  }
+  subnets = [
+    {
+      ip_cidr_range = null
+      name          = "test-internal-v6only"
+      region        = "us-east1"
+      ipv6          = {
+        ipv6_only   = true
+      }
+    },
+    {
+      ip_cidr_range = null
+      name          = "test-external-v6only"
+      region        = "us-east1"
+      ipv6 = {
+        access_type = "EXTERNAL"
+        ipv6_only   = true
+      }
+    }
+  ]
+}
+# tftest modules=1 resources=6 inventory=ipv6.yaml e2e
+```
 <!-- BEGIN TFDOC -->
 ## Variables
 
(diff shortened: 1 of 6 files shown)


modules/net-vpc/subnets.tf
```diff
@@ -139,13 +141,21 @@ locals {
...
+  provider = google-beta
+  for_each = local.subnets
+  project  = var.project_id
+  network  = local.network.name
+  name     = each.value.name
+  region   = each.value.region
+  ip_cidr_range = (
+    try(each.value.ipv6, null) != null
+    ? (
+      each.value.ipv6.ipv6_only
```
ludoo: nit: why not `try`ing `ipv6.ipv6_only` directly?

modules/net-vpc/README.md
```diff
@@ -712,6 +712,46 @@ module "vpc" {
 }
 # tftest modules=1 resources=6 inventory=ipv6.yaml e2e
 ```
+
+### IPv6-Only and IP Collections
+
+An IPv6-only subnetwork can be specified by setting `ipv6_only` to `true` and
+setting or `ip_cidr_range` to `null` or omitting it entirely.  An IP Collection
```
ludoo: extra `or`?

ludoo (approved):
Thanks a lot for this, left a couple minor nits and @sruffilli will also have to give this an eye, but looks great to me!

modules/net-vpc/README.md
```diff
@@ -712,6 +712,46 @@ module "vpc" {
...
+
+```hcl
+module "vpc" {
+  source     = "./fabric/modules/net-vpc"
+  project_id = var.project_id
+  name       = "my-network"
+  ipv6_config = {
+    enable_ula_internal = true
+  }
+  subnets = [
```
sruffilli, line 733: I would also include a dummy reference to `ip_collection` to make sure to test that in the plan as well.

modules/net-vpc/versions.tf
```diff
@@ -18,12 +18,12 @@ terraform {
   required_version = ">= 1.11.4"
   required_providers {
     google = {
-      source  = "hashicorp/google"
-      version = ">= 6.33.0, < 7.0.0" # tftest
+      source = "hashicorp/google"
```
ludoo: Please do not change this on an individual module (ht @sruffilli as I had missed this). Change the top-level default versions files for both tf and tofu, and roll them out across all modules via the `tools/versions.py` script.

ludoo (changes requested):
versions issue (see above comment)

modules/net-vpc/versions.tf
cmm-cisco: It's a relatively new feature.  I'll try putting the version back and hopefully it will test okay.  Sorry about that.

modules/net-vpc/README.md
cmm-cisco: Hello.  I'm not sure how to test this, since it would require an IP collection to be created.  Do you already have a PDP available to test?  I could just comment it out if that would work okay.  Do you have a standard for a self-URL pointing to a project for documentation?

modules/net-vpc/subnets.tf
cmm-cisco: Good point, will fix.

modules/net-vpc/README.md
cmm-cisco: Will fix.

modules/net-vpc/versions.tf
cmm-cisco: I think I see where you have the version set across all modules.  Let me know if this was wrong.

modules/net-vpc/README.md
cmm-cisco: Set a commented out ip_collection.  Hopefully this works okay.

sruffilli: You can come up with an ID that mimics the ID for the resource so that we can test if the plan actually works with our testing framework. Apply would obviously fail but that's not relevant for examples.

modules/net-vpc/versions.tf
ludoo: It's not wrong, but you need to roll that out to every module :)

ludoo: and the versions file contain the module name, this is why I was sugegsting use of our tool which does that for you

PR #3178, opened 2025-06-20
Added tag factory option for organization module

lnesteroff:
Added tag factory option for organization module. Combines the factory and var approaches for tags so will be backwards compatible. 

---
**Checklist**


lnesteroff:
> This is very nice. Couple comments before this is good to go:
> 
> * the same factory needs to be implemented in the project module
> * we might want to add `context.tag_keys` and `context.tag_values` to `factories_config` and then lookup the parent id for keys and values there, and fallback to the passed in value if the key is missing

Happy to add this to the project module, but isn't this already achieved through the PF? You can incl the tag defs in the PF yamls

ludoo:
But not all project creation goes through pf :)

lnesteroff:
Great point, would you want it in this PR or separate?

PR #3190, opened 2025-06-23
Added option for tag factory in resman

lnesteroff:
Added option to use tag factory in resman.

---
**Checklist**


lnesteroff:
Any thoughts on setting a default dir for tags factory in var.factories_config.tags? I've currently got it set to data/tags to align with the style of the other factories, but we don't have any use of the tags factory by default.

ludoo:
> Any thoughts on setting a default dir for tags factory in var.factories_config.tags? I've currently got it set to data/tags to align with the style of the other factories, but we don't have any use of the tags factory by default.

I would merge this PR to enable support for tags, but start a discussion on whether to transition our code defined tags to the factory.

Pros: we would simplify code and remove the need to expose tag names via variables.
Cons: we depend on specific tags to exist in several places.

ludoo (approved):

PR #3197, opened 2025-06-27
Create (or import) subnets with empty description

lnesteroff:
Create (or import) subnets with empty description. Google forces recreate on subnet description change (for some reason). This change lines up how the alternative types of subnets (proxy only/PSC) handles descriptions with how the regular subnet blocks handles descriptions. This is required when importing existing subnets into the subnet factory that don't have a description without needed to recreate the whole subnet.

---
**Checklist**

### DIFF
--- a/modules/net-vpc/subnets.tf
+++ b/modules/net-vpc/subnets.tf
@@ -150,6 +150,7 @@ resource "google_compute_subnetwork" "subnetwork" {
   ip_cidr_range                    = try(each.value.ipv6.ipv6_only, false) ? null : each.value.ip_cidr_range
   allow_subnet_cidr_routes_overlap = each.value.allow_subnet_cidr_routes_overlap
   description = (
+    # Set description to an empty string (eg "") to create subnet without a description.
     each.value.description == null
     ? "Terraform-managed."
     : each.value.description
@@ -218,6 +219,7 @@ resource "google_compute_subnetwork" "private_nat" {
   region        = each.value.region
   ip_cidr_range = each.value.ip_cidr_range
   description = (
+    # Set description to an empty string (eg "") to create subnet without a description.
     each.value.description == null
     ? "Terraform-managed private NAT subnet."
     : each.value.description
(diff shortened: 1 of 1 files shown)


wiktorn (approved):

juliocc:
I see. Thanks for the explanation.

We have coalesce in a lot of places. Seems like we'll have to go around changing them.

lnesteroff:
> > Coalesce passes first non null or empty. So if we wanted to create (or import in my case currently) a proxy_only subnet with NO description at all, and in the factory we have: description: "" Coalesce would mean that description is "Terraform-managed proxy-only subnet for Regional HTTPS, Internal HTTPS or Cross-Regional HTTPS Internal LB."
> 
> Just a follow up remark. As I understand the intention given the description in the PR, I probably won't remember that, when I'll be looking at the code. Maybe it's worth adding a code in place, for our future selves?
> 
> (or add a test around it)

When you say adding a code in place, what do you mean?

lnesteroff:
> I see. Thanks for the explanation.
> 
> We have coalesce in a lot of places. Seems like we'll have to go around changing them.

TBH most of the time you'd want it to at least say it's managed by Terraform. It's just for whatever reason, we force destroy and recreate on changes to description of subnets lol

juliocc (approved):

modules/net-vpc/subnets.tf
```diff
@@ -201,9 +202,10 @@ resource "google_compute_subnetwork" "proxy_only" {
   name          = each.value.name
   region        = each.value.region
   ip_cidr_range = each.value.ip_cidr_range
-  description = coalesce(
-    each.value.description,
-    "Terraform-managed proxy-only subnet for Regional HTTPS, Internal HTTPS or Cross-Regional HTTPS Internal LB."
+  description = (
+    each.value.description == null
```
wiktorn, line 207:
```suggestion
    # Set description to an empty string (eg "") to create subnet without a description.
    each.value.description == null
```

PR #3205, opened 2025-06-30
exposing the network_id from the net-vpc module for use with tags

lnesteroff:
For binding secure tags to network resources we require the resource number (https://cloud.google.com/vpc/docs/create-manage-tags-vpc-resources#attach_tags). This change exposes the network_id on the net-vpc outputs. Given I had to create a datasources, I also converted id/name/self_link to also use the datasource option instead of template.

---
**Checklist**

### DIFF
--- a/modules/net-vpc/main.tf
+++ b/modules/net-vpc/main.tf
@@ -18,22 +18,16 @@ locals {
   network = (
     var.vpc_create
     ? {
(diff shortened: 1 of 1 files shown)


ludoo (changes requested):
We limit datasource use as much as possible, and not only because it fails in tests. Have a look at the project module for how to properly implement this.

ludoo (approved):

Issue #3208, opened 2025-07-02
[project-factory] Service Agent Substitution only works with "iam"


V0idC0de:
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

Substitution of Service Agent `iam_email` values works only with the `iam:` attribute of a project, not `iam_bindings:` and `iam_bindings_additive:`.

### Release version

v41.0.0

### Terraform version

1.12.2

### Steps to reproduce

Substitution works with the following project configuration:
```yaml
services:
  - dialogflow.googleapis.com
iam:
  roles/cloudfunctions.invoker:
    - dialogflow
```

however, the following substitution fails with the following log:
```yaml
services:
  - dialogflow.googleapis.com
iam_bindings_additive:
  dialogflow_sa_invoker:
    role: roles/cloudfunctions.invoker
    member: dialogflow
```

```
Error: Error in function call

  on .terraform/modules/factory/modules/project-factory/main.tf line 197, in module "projects-iam":
--- snip for readability ---
    ├────────────────
    │ each.key is "project-key-redacted"
    │ local.context.iam_principals is object with no attributes
    │ local.service_agents_email is object with 47 attributes
    │ module.projects is object with 1 attribute "project-key-redacted"
    │ module.service-accounts is object with no attributes
    │ v.member is "dialogflow"
Call to function "try" failed: no expression succeeded:
--- snip for readability ---
```

### Expected behavior

Substitution is performed consistently, regardless of `iam`, `iam_bindings`, or `iam_bindings_additive` being used.

### Actual behavior

Substitution works for `iam`, but fails for the others.

### Additional context

This is probably caused by the way, the Service Identities are stored in the `locals`.
Here is the definition of the Service Identities' E-Mails: https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/commit/7ce4381ac3536df1a6f969e8a51c80a512f040fa#diff-5299df52ff6618d346ba21bdfd78a430b0468246f23b05c3875f2bf13056cb31R45-R47
This stores them as map, using keys like `<project-key>/<service-agent-name>` and values containing their `iam_email`.

[...]

PR #3210, opened 2025-07-02
Fix for service agent substitutions in project factory additive bindings

V0idC0de:
Fixes broken substitution of Service Agents when calling module `project-iam`. Setting `iam_bindings` and `iam_bindings_additive` now substitutes like `iam`, where it already works.

Addresses the substitution issue described here: https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/3208

---
**Checklist**

### DIFF
--- a/modules/project-factory/main.tf
+++ b/modules/project-factory/main.tf
@@ -178,7 +178,8 @@ module "projects-iam" {
           # other automation service account (project/automation/rw)
           local.context.iam_principals[vv],
           # project's service identities
-          local.service_agents_email[each.key][vv],
+          local.service_agents_email["${each.key}/${vv}"],
+          local.service_agents_email[vv],
           # passthrough + error handling using tonumber until Terraform gets fail/raise function
           (
             strcontains(vv, ":")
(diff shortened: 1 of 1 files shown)


juliocc (approved):
Nice catch. I've fixed a couple of small details. 

If all tests pass we can go ahead and merge.

ludoo (dismissed):
I'm not sure these changes are right. The local is across all projects, so it's prefixed by project key.

V0idC0de:
> I'm not sure these changes are right. The local is across all projects, so it's prefixed by project key.

Yes, it is. The `local.service_agents_email` uses a single composed key to map to the `iam_email`. That key consist of `project_key` and the Service Agent's name (as in the `service-agents.yaml`) separated by a `/`.

The substitution first prepends `${each.key}/` to the `member`, assuming that you're referring to the current project (since nothing else was specified). If that fails, it uses the `member` literally, which then has to contain the `project_key` (useful, if you're referring to another project's Service Agent).

So the original code's intention is correct, but the prefix (project-key) and Service Agent name are stored in the very same key as combination, instead of a nested map (which is what the previous code tries to look-up).

ludoo:
Yep chatted with Julio it's all good, should not check code on my phone waiting for the bus 😂 Thanks for raising this.!

ludoo (approved):

PR #3218, opened 2025-07-06
Fixed gke-hub module to support regional deployment

Art1k:
---
**Checklist**

### DIFF
--- a/modules/gke-hub/main.tf
+++ b/modules/gke-hub/main.tf
@@ -38,6 +38,7 @@ resource "google_gke_hub_membership" "default" {
   provider      = google-beta
   for_each      = var.clusters
   project       = var.project_id
+  location      = var.location != null ? var.location : "global"
   membership_id = each.key
   endpoint {
     gke_cluster {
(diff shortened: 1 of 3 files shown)


modules/gke-hub/main.tf
```diff
@@ -38,6 +38,7 @@ resource "google_gke_hub_membership" "default" {
   provider      = google-beta
   for_each      = var.clusters
   project       = var.project_id
+  location      = var.location != null ? var.location : "global"
```
juliocc: Don't try to guess. If the user wants global, let them pass it it.
```suggestion
  location      = var.location
```

modules/gke-hub/main.tf
```diff
@@ -68,28 +69,38 @@ resource "google_gke_hub_feature" "default" {
...
     }
   }
+  dynamic "fleet_default_member_config" {
+    for_each = each.key == "servicemesh" && each.value != null ? { 1 = 1 } : {}
+    content {
+      mesh {
+        management = "MANAGEMENT_AUTOMATIC"
+      }
+    }
+  }
```
juliocc: Should this be part of this PR? IMO this should be its own variable to allow the user to properly configure `fleet_default_member_config`

juliocc (changes requested):

modules/gke-hub/main.tf
Art1k: This was causing issues, when you sync GKE into fleet, you keep getting changes or if this not set and cluster has joined the fleet => TF always tries to remove it, i would do something like 

```
dynamic "fleet_default_member_config" {
    for_each = var.fleet_default_member_config != null ? { 1 = 1 } : {}
    content {
      mesh {
        management = var.fleet_default_member_config.management
      }
    }
  }
```

modules/gke-hub/variables.tf
```diff
@@ -63,6 +63,15 @@ variable "configmanagement_templates" {
...
 
+variable "fleet_default_member_config" {
+  description = "Fleet default member config."
+  type = object({
+    management = optional(string, "MANAGEMENT_AUTOMATIC")
+  })
+  default  = null
+  nullable = true
+}
+
```
juliocc: This is not enough. If you're going to expose `fleet_default_member_config` you need to provide the same level of flexibility as the matching attribute in the provider.

juliocc (approved):

PR #3219, opened 2025-07-07
Allowing multiple on-prem domains

lnesteroff:
This allows the creation of multiple forwarders to onprem resolvers. It also allows overwriting the resolvers on specific domains if they need to be resolved by specific resolvers.

I understand this needs to be applied to the other network stages, and probably will need to update the docs a bit, but just wanted to sanity check the approach first.

---
**Checklist**


fast/stages/2-networking-a-simple/variables.tf
```diff
@@ -42,12 +42,19 @@ variable "alert_config" {
 variable "dns" {
   description = "DNS configuration."
   type = object({
-    gcp_domain    = optional(string, "gcp.example.com")
-    onprem_domain = optional(string, "onprem.example.com")
-    resolvers     = optional(list(string), [])
+    gcp_domain = optional(string, "gcp.example.com")
+    onprem_domain = optional(list(object({
```
sruffilli: change to `onprem_domains` (plural)

fast/stages/2-networking-c-separate-envs/variables.tf
```diff
@@ -42,13 +42,34 @@ variable "alert_config" {
 variable "dns" {
   description = "DNS configuration."
   type = object({
-    gcp_domain     = optional(string, "gcp.example.com")
-    onprem_domain  = optional(string, "onprem.example.com")
+    gcp_domain = optional(string, "gcp.example.com")
+    onprem_domain = optional(list(object({
```
sruffilli: change to `onprem_domains` (plural)

sruffilli (changes requested):

fast/stages/2-networking-a-simple/variables.tf
lnesteroff: Good call-out, thanks!

sruffilli (approved):

PR #3222, opened 2025-07-08
Extend support for universe

karpok78:
---
**Checklist**


modules/compute-vm/outputs.tf
```diff
@@ -86,10 +86,12 @@ output "service_account_iam_email" {
...
+  value = try(coalesce(google_compute_region_instance_template.default[0],
+  google_compute_region_instance_template.default[0]), null)
 }
 
 output "template_name" {
   description = "Template name."
-  value       = try(google_compute_instance_template.default[0].name, null)
+  value = try(coalesce(google_compute_region_instance_template.default[0].name,
+  google_compute_region_instance_template.default[0].name), null)
 }
```
juliocc: I think `coalesce(x,x)` is a nop. Can you explain this change?

modules/compute-vm/main.tf
```diff
@@ -35,8 +35,10 @@ locals {
     ? "TERMINATE"
     : "MIGRATE"
   )
-  region = join("-", slice(split("-", var.zone), 0, 2))
-  gpu    = var.gpu != null
+  region = join("-", slice(split("-", var.zone), 0,
+    startswith(var.zone, "u-") ? 3 : 2
+  ))
```
juliocc, line 40: All these magic numbers make this really difficult to read. What is the format for regions in other universes?

modules/project-factory/main.tf
```diff
@@ -126,7 +127,7 @@ module "projects" {
 module "projects-iam" {
   source   = "../project"
   for_each = local.projects
-  name     = module.projects[each.key].project_id
+  name     = module.projects[each.key].name
```
juliocc, line 130: why switch `project_id` with `name` here?

modules/compute-vm/outputs.tf
karpok78: Currently the model does not output regional template data.
Coalosce is used here  to export the correct data depending on which kind of template is deployed using the module

modules/compute-vm/main.tf
karpok78: Other region universe use the same format as native ones, but prefixed with u-.

modules/project-factory/main.tf
ludoo: As Julio noted, please don't change project factory code without providing an extending rationale and tests.

karpok78: This allows the prefix parameter of the universe variable to be leveraged.
Using project_id results in a doubling the prefix for certain iam bindings.

ludoo: We can'ut use the name, as it can be different from the project id. We need to address this in a different way.

modules/compute-vm/outputs.tf
juliocc: My point is that you're doing 
```
coalesce(google_compute_region_instance_template.default[0].name,
         google_compute_region_instance_template.default[0].name)
```

This is not doing anything significant

modules/project-factory/main.tf
karpok78: You're right.
Here I was just aligning parameters for module projects and projects-aim for consistency :
`module "projects" {
  source          = "../project"
  for_each        = local.projects
  billing_account = each.value.billing_account
  deletion_policy = each.value.deletion_policy
  name            = each.value.name
  `
but
  `module "projects-iam" {
  source   = "../project"
  for_each = local.projects
  name     = module.projects[each.key].project_id
  project_reuse = {
    use_data_source = false
    attributes = {
      name             = module.projects[each.key].name
      number           = module.projects[each.key].number
      services_enabled = module.projects[each.key].services
    }
  }`
  
  Actually the issue lies in the projects module that derives the project_id from the project name.
  I understand this was not the intended behavior ? Is that it ?

modules/compute-vm/outputs.tf
karpok78: My bad,
I copy paste too fast. I added a new commit to coalesce google_compute_instance_template and google_compute_region_instance_template

modules/project-factory/main.tf
ludoo: No, it derives the project id from the combination of prefix and name. You always need to refer to project_id, not only because it's the right attribute, but also because a lot of dependencies are on it.

If project id is broken for universe, we need to fix that in the project module not try and do things that will break everyone else. :)

ludoo:
Closing this, as Julio suggested we need to tackle this one issue at a time. Feel free to reopen, or better yet open an issue with an actual code example against the project module that shows how project id is different from what you would expect.

PR #3224, opened 2025-07-08
Support regional instance templates in compute-vm module outputs

### LINKED ISSUE #3225: compute-vm module outputs are null for regional instance templates
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

The compute-vm module was updated in v40.0.0 to support regional compute instance templates. While the underlying regional resources are created correctly when using this feature, the module's outputs (e.g., instance_id, instance_name, etc.) are returning null.

### Release version

02a9bdfe2

### Terraform version

v1.12.0

### Steps to reproduce

1. Configure the compute-vm module to create a VM from a regional instance template.
2. Run terraform apply. The resources will be created successfully in Google Cloud.
3. Check the Terraform outputs for the module.

### Expected behavior

The module outputs (instance_id, instance_name, etc.) should be populated with the attributes of the newly created instance template, just as they are when using a global instance template.

### Actual behavior

The module outputs are null, despite the successful creation of the instance template in Google Cloud.

### Additional context

_No response_

javiroger:
This PR fixes the null outputs when creating regional instance templates.

Fixes #3225

---
**Checklist**

### DIFF
--- a/modules/compute-vm/template.tf
+++ b/modules/compute-vm/template.tf
@@ -332,13 +332,6 @@ resource "google_compute_region_instance_template" "default" {
     }
   }
 
-  dynamic "network_interface" {
-    for_each = var.network_attached_interfaces
-    content {
-      network_attachment = network_interface.value
-    }
-  }
-
   scheduling {
     automatic_restart           = !var.options.spot
     instance_termination_action = local.termination_action
--- a/modules/compute-vm/README.md
+++ b/modules/compute-vm/README.md
@@ -991,7 +991,7 @@ module "sole-tenancy" {
 | [service_account_email](outputs.tf#L73) | Service account email. |  |
 | [service_account_iam_email](outputs.tf#L78) | Service account email. |  |
 | [template](outputs.tf#L87) | Template resource. |  |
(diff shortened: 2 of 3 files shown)


modules/compute-vm/template.tf
```diff
@@ -332,13 +332,6 @@ resource "google_compute_region_instance_template" "default" {
     }
   }
 
-  dynamic "network_interface" {
```
ludoo, line 335: Why has this been removed?

javiroger: Hi @ludoo, thanks for the quick reply and sorry for the confusion, as the PR description was created automatically by Gemini CLI 🙂. I've updated it. The dynamic “network_interface” has been removed because it caused the following error: Unexpected attribute: An attribute named "network_attachment" is not expected here. Checking the [docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_region_instance_template#nested_network_interface), it appears that it is not supported as in the global instance templates.

PR #3235, opened 2025-07-14
Option to use remote state instead of tfvars

lnesteroff:
Option to use remote state instead of tfvars. This is required for automation tooling such as HCP Terraform/TFC/TFE which is not able to pull tfvars at run time. 

Just wanted to discuss the approach and if there are any obvious issues before polishing and applying to the other stages.

Most of the logic is in fast/stages/1-resman/data-sources.tf. I would likely want to move some of the locals into main.tf, but kept it all together for the initial review.

---
**Checklist**


ludoo:
I personally would not want the extra complexity and indirection in FAST. :)

PR #3246, opened 2025-07-21
feat: Add Service Agent substitution for Buckets and `iam_by_principal` in `project-factory`

V0idC0de:
Similar to https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/3210, this PR adds the feature to use Service Agent aliases in IAM Bindings for Buckets, created in the `project-factory` module.

Projects can already do this but I just realized that this is not the case for Buckets.
For example, this is useful when storing ZIP files with Cloud Function code in a Bucket. For deploying these ZIPs via the `cloud_function_V2` module, it is required, that the `gcf-admin-robot` Service Agent can read this Bucket.

But once assigned via a `member` value of `project_name/gcf-admin-robot` or `gcf-admin-robot` (if Bucket is in the same Project as the Function to be deployed), the `try()` fails, because none of the queried Service Account lists have that alias. Service Agents are not considered here yet, which is different from how Project IAM is handled.

As there should be no circular dependency between Buckets and Service Accounts, this seems like an easy fix. It worked in my local version, where it tested the code.

---
**Checklist**


V0idC0de:
Rebased my branch onto your current `master`.
Also, I discovered, that the substitution for IAM permissions in Buckets will consistently fail in many cases, since the logic is using a non-existent key to lookup the Projects.
The Bucket's `each.key` is a combination of `project_key` and the Bucket's `name`, as seen [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/3246/files#diff-5299df52ff6618d346ba21bdfd78a430b0468246f23b05c3875f2bf13056cb31R329).
Now, this exact key is used to lookup Projects, which always fails, since the Projects obviously don't contain the Bucket's name (see [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/3246/files#diff-5299df52ff6618d346ba21bdfd78a430b0468246f23b05c3875f2bf13056cb31L375)).

Also, other lookups use `each.value.project` (which doesn't exist, as seen in the creation of the `local.buckets` object [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/ffb11c98df8cda2f8dc9b69ab067121015eb8287/modules/project-factory/factory-projects.tf#L71)) instead of `each.value.project_key`.

The latest commit resolves that, by consistently using either `each.value.project_key` instead of `each.key` or `each.key.project`.

I stumbled upon this, since it prevented any permissions from being set via literal `member` values and any substitutions that didn't explicitly specified the `project` key (or happened to be one of the occasions where it was already correctly implemented.

V0idC0de:
Another comment, since I was also digging into substitution for Service Account under `automation` and `iam_sa_roles` in general.

I noticed that this substitution order is essentially always the same. Since Projects and Service Account are now created without any permissions and then permissions are assigned via a second `...-iam` module call, I'd suggest looking into a `local` variable, building this substitution map once for Service Accounts (including those created under `automation`) and then using that everywhere. Seems like a natural consolidation, after you added the awesome substitution feature everywhere from time to time.

`automation` is still a unicorn though. It doesn't have its own `...-iam` module call and cannot substitute its members that well. Any member must be known explicitly in advance or be part of the context. However, it would be really useful for the pattern, that FAST employs itself when building Service Accounts for GitHub integrations, where one Account has `Service Account Token Creator` on the other, to impersonate it.
This is currently not really possible, since the substitution doesn't work for these `automation` Service Accounts. It does work for regular Service Accounts within the same Project, though. I already tried that.

Once I find some time, I'll look into how this could be done, since it changes and consolidates some code all over the place. Maybe you also have something in that direction already or are opposed to that for some reason :)

PR #3249, opened 2025-07-23
Feat: Add branch protection object to the repositories variable

Ali-Aburub:
---
**Checklist**


modules/secure-source-manager-instance/main.tf
```diff
@@ -14,6 +14,22 @@
...
+      repository                = k1
+      branch_rule_id            = k2
+      include_pattern           = v2.include_pattern
+      minimum_approvals_count   = try(v2.minimum_approvals_count, null)
+      minimum_reviews_count     = try(v2.minimum_reviews_count, null)
+      require_comments_resolved = try(v2.require_comments_resolved, null)
+      require_linear_history    = try(v2.require_linear_history, null)
+      require_pull_request      = try(v2.require_pull_request, null)
+      disabled                  = try(v2.disabled, null)
+      allow_stale_reviews       = try(v2.allow_stale_reviews, null)
```
juliocc: I think all these try are unnecesary. Can you try (pun intended) removing them?

modules/secure-source-manager-instance/main.tf
```diff
@@ -14,6 +14,22 @@
  * limitations under the License.
  */
 
+locals {
+  branch_rules = merge([for k1, v1 in var.repositories : { for k2, v2 in v1.branch_rules :
```
juliocc: Split this in a few lines. Something like 

```terraform

  branch_rules = merge([
    for k1, v1 in var.repositories : { 
       for k2, v2 in v1.branch_rules :
```

juliocc (dismissed):
Thanks for this PR. Very useful.

Could you add at least one (tested) example in the README using the new feature?

modules/secure-source-manager-instance/main.tf
Ali-Aburub: I will TRY that out :)

wiktorn:
Thanks a lot @Ali-Aburub

PR #3255, opened 2025-07-31
FAST light implementation

ludoo:
This is a prototype coming out of discussions between @juliocc and I, that completely redesigns our context replacement system, and creates a much lighter and more flexible alternative stage for FAST that combines bootstrap, resource management and project factory in a single stage.

Blueprints have also been deprecated here, as we have been discussing for at least a year, since lower level implementations should move to project templates.

This is still a work in progress, main items left

- [x] workload identity pools/providers support
- [x] ci/cd support
- [x] tfvars output file from the new bootstrap stage
- [x] allow splitting tag resources and IAM creation in modules
- [x] test compatibility of current FAST stages
  - [x] security
  - [x] networking
  - [x] project factory
- [x] documentation and diagrams
- [x] "legacy FAST" compatible dataset
- [ ] "light FAST" dataset (maybe)
- [ ] alternative dataset @juliocc 

The included data set is used to test this code in an actual organization, aims at eventual compatibility with current FAST, and will be refactored into a generalized implementation before merging.


wiktorn:
Can we split blueprint deprecation into a separate (and quick) PR? Reviewing this will be hard with Github interface :-/

PR #3275, opened 2025-08-13
Provide node_pool_auto_config only when node auto provisioning is enabled for GKE standard cluster

kumadee:
For a GKE standard cluster with node auto provisioning disabled, the node_pool_auto_config settings are irrelevant and seeing the changes in terraform plan can be confusing.

With this PR, if the node auto provisioning is enabled then node_pool_auto_config settings will be applied otherwise not.

---
**Checklist**

### DIFF
--- a/modules/gke-cluster-standard/main.tf
+++ b/modules/gke-cluster-standard/main.tf
@@ -88,16 +88,19 @@ resource "google_container_cluster" "cluster" {
       }
     }
   }
-  node_pool_auto_config {
-    network_tags {
-      tags = var.node_pool_auto_config.network_tags
-    }
-    resource_manager_tags = var.node_pool_auto_config.resource_manager_tags
-    node_kubelet_config {
-      insecure_kubelet_readonly_port_enabled = upper(var.node_pool_auto_config.kubelet_readonly_port_enabled)
-    }
-    linux_node_config {
-      cgroup_mode = var.node_pool_auto_config.cgroup_mode
+  dynamic "node_pool_auto_config" {
+    for_each = (local.cas != null && local.cas.enabled) ? [""] : []
+    content {
+      network_tags {
+        tags = var.node_pool_auto_config.network_tags
+      }
+      resource_manager_tags = var.node_pool_auto_config.resource_manager_tags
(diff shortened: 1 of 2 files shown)


ludoo (commented):
Thanks for this PR, can you apply the same changes to the autopilot module?

kumadee:
> Thanks for this PR, can you apply the same changes to the autopilot module?

Currently the autopilot module always enables (implicitly done if `cluster_autoscaling.enabled` is not defined) node auto provisioning. From that perspective this change is not applicable.

I could add this change only if we want to make node auto provisioning optional for GKE autopilot. In such a case, the end user also has to ensure that they specify their own node pools. We would also have to do additional refactoring of the autopilot module for e.g. it is missing `node_pool_defaults` which might become relevant for user defined node pools.

ludoo (approved):

kumadee:
Looks like the test failed on CI due to usage of terraform < 1.12.0. I have a higher version on my local.
Logical binary operation short-circuiting was introduced with https://github.com/hashicorp/terraform/pull/36224 in terraform v1.12.0.

Nonetheless, I will handle the issue for lower versions of terraform.

modules/gke-cluster-standard/main.tf
```diff
@@ -88,16 +88,19 @@ resource "google_container_cluster" "cluster" {
...
-      tags = var.node_pool_auto_config.network_tags
-    }
-    resource_manager_tags = var.node_pool_auto_config.resource_manager_tags
-    node_kubelet_config {
-      insecure_kubelet_readonly_port_enabled = upper(var.node_pool_auto_config.kubelet_readonly_port_enabled)
-    }
-    linux_node_config {
-      cgroup_mode = var.node_pool_auto_config.cgroup_mode
+  dynamic "node_pool_auto_config" {
+    for_each = (local.cas != null && try(local.cas.enabled, false)) ? [""] : []
```
ludoo: why not `try(local.cas.enabled, null) == true`?

kumadee: refactored as suggested

ludoo (approved):

PR #3333, opened 2025-09-17
Add conditional grants on networking to pf service account in FAST stage 0

fenyvesi-levi:
---
**Checklist**

The project-factory stage ran into an error when trying to grant network user role for principals

### DIFF
--- a/fast/stages/0-org-setup/data/folders/networking/.config.yaml
+++ b/fast/stages/0-org-setup/data/folders/networking/.config.yaml
@@ -29,7 +29,6 @@ iam_by_principals:
     - roles/viewer
   $iam_principals:service_accounts/iac-0/iac-pf-rw:
     - $custom_roles:service_project_network_admin
-    - roles/resourcemanager.projectIamAdmin
   $iam_principals:service_accounts/iac-0/iac-pf-ro:
     - roles/compute.viewer
     - $custom_roles:project_iam_viewer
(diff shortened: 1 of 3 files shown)


fast/stages/0-org-setup/data/folders/networking/.config.yaml
```diff
@@ -29,8 +29,10 @@ iam_by_principals:
     - roles/viewer
   $iam_principals:service_accounts/iac-0/iac-pf-rw:
     - $custom_roles:service_project_network_admin
+    - roles/resourcemanager.projectIamAdmin
```
ludoo: We typically don't do this, but implement it via a delegated role grant that only allows specific roles.

ludoo: [this](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/v43.0.0/fast/stages/1-resman/data/stage-2/networking.yaml#L40)

fenyvesi-levi: Thanks @ludoo ! Reconfigured the role grant so that the pf sa can only grant 4 roles (same as in 1-resman-legacy).
Also changed the tfvars output to sensitive if that's okay, so it wouldn't output values like billing account ID.

fast/stages/0-org-setup/outputs.tf
```diff
@@ -32,4 +32,5 @@ output "projects" {
 output "tfvars" {
   description = "Stage tfvars."
   value       = local.of_tfvars
+  sensitive   = true
```
ludoo, line 35: I don't think there's a real risk outputting the billing account id. But this won't hurt so it's ok.

ludoo:
At some point a similar binding will need to be brought back on the security stage for the Key Encrypt/Decrypt role.

PR #3335, opened 2025-09-17
docs(blueprints): add deprecation notice

kaue:
adds a README.md file to a new, empty blueprints directory to inform users that the Terraform blueprints have been deprecated as of the v44.0.0 release


blueprints/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+**This directory is deprecated as of release `v44.0.0`.**
+
+The blueprints that were previously in this directory have been deprecated. Some of the old blueprints will be refactored as either FAST project templates or module-level recipes.
+
+If there is a specific blueprint that you would like to see refactored, please open a feature request.
+
+## Accessing the Deprecated Blueprints
+
+The last version of the blueprints is available in the `v43.0.0` tag. You can still use them by referencing this tag in your Terraform configuration.
```
ludoo: I would omit the code below and just point to the download page for that release. Blueprints are usually meant to be deployed directly and not via a module call.

PR #3336, opened 2025-09-18
Allow setting resource policies in compute-vm module

ludoo:
**Breaking Changes**

```upgrade-note
`modules/compute-vm`: the instance schedule variable has changed type.
```


modules/compute-vm/resource-policies.tf
```diff
@@ -55,27 +45,27 @@ locals {
...
   instance_schedule_policy {
-    expiration_time = local.ischedule.expiration_time
-    start_time      = local.ischedule.start_time
-    time_zone       = local.ischedule.timezone
+    expiration_time = var.instance_schedule.expiration_time
+    start_time      = var.instance_schedule.start_time
+    time_zone       = var.instance_schedule.timezone
     dynamic "vm_start_schedule" {
-      for_each = local.ischedule.vm_start != null ? [""] : []
+      for_each = var.instance_schedule.vm_start != null ? [""] : []
```
wiktorn, line 60: I've noticed @juliocc  is using in such contexts:
```suggestion
      for_each = var.instance_schedule.vm_start[*]
```
(See [splat expressions](https://developer.hashicorp.com/terraform/language/expressions/splat#single-values-as-lists)). It is not obvious at the beginning, but once learned might be a nice pattern to use. WDYT?

ludoo: Ah interesting, I never saw the splat used in for each. :) If we decide that's our standard, there's a ton of places where we need to change it.

ludoo: @wiktorn what happens is the vsalue is null? would the splat still work?

wiktorn: >  If the value is null then the splat expression will return an empty tuple.

ludoo: No it does not work

image

image

ludoo: It does not seem to do that.

PR #3345, opened 2025-09-21
Rationalize prefix handling for project factory automation resources

ludoo:
This PR rationalizes handling of prefix in project factory automation resources by

- skipping the global prefix in service account names, to simplify length constraints and adopt the convention used in the FAST org setup stage
- allowing to override prefix in storage buckets, so that an existing bucket can be referenced by "portable" key when defining managed folders and bucket creation is not wanted

A new subsection has been added to the project factory README to explain prefix behaviour for automation resources.

**Breaking Changes**

```upgrade-note
`modules/project-factory`: the format for automation service account names has changed.
```

### DIFF
--- a/fast/stages/0-org-setup/schemas/project.schema.json
+++ b/fast/stages/0-org-setup/schemas/project.schema.json
@@ -530,6 +530,10 @@
         "name": {
           "type": "string"
         },
+        "create": {
+          "type": "boolean",
+          "default": true
+        },
         "description": {
           "type": "string"
         },
--- a/modules/project-factory/README.md
+++ b/modules/project-factory/README.md
@@ -31,7 +31,9 @@ The code is meant to be executed by a high level service account with powerful p
   - [Factory-wide project defaults, merges, optionals](#factory-wide-project-defaults-merges-optionals)
   - [Project templates](#project-templates)
   - [Service accounts and buckets](#service-accounts-and-buckets)
-  - [Automation project and resources](#automation-project-and-resources)
+  - [Automation resources](#automation-resources)
+    - [Prefix handling](#prefix-handling)
+    - [Complete automation example](#complete-automation-example)
 - [Billing budgets](#billing-budgets)
 - [Context-based interpolation](#context-based-interpolation)
   - [Folder context ids](#folder-context-ids)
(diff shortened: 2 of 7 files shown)


fast/stages/0-org-setup/schemas/project.schema.json
```diff
@@ -530,6 +530,10 @@
         "name": {
           "type": "string"
         },
+        "create": {
+          "type": "boolean",
+          "default": true
```
wiktorn, line 535: I'd not add defaults in json schema, as it is additional place to maintain, and it is not adding any value, doesn't it?

modules/project-factory/README.md
```diff
@@ -118,17 +120,60 @@ buckets:
...
 - an optional `service_accounts` list where each element defines a service account in the controlling project
-- an optional `bucket` which defines a bucket in the controlling project, and the map of roles/principals in the corresponding value assigned on the created bucket; principals can refer to the created service accounts by key
+- an optional `bucket` which defines a bucket and/org managed folders in the controlling project; bucket names cannot use interpolation so where bucket creation is not needed, they need to be explicit
+
+#### Prefix handling
+
+To easily distinguish automation resources in the controlling project, service account and bucket names use a prefix that embeds the "local" project name to the default prefix. Due to the difference in maximum length and name uniqueness, service accounts and buckets treat the prefix differently.
+
+For service accounts the global prefix is ignored, and the "local" project name is used as a prefix. For example, a project defined in a `prod-app-example-0.yaml` file where the prefix is `foo` will have the `rw` automation service account resulting in the `prod-app-example-0-rw` name.
```
wiktorn, line 139: Adding prefixes to the projects is super-useful when SA are provisioned into a dedicated project. But for cases, when CICD is done from the application project, I'd love to see an option to disable this, and just have the same name across all projects.

fast/stages/0-org-setup/schemas/project.schema.json
ludoo: It does have a value: autocomplete. And it's a signal of what the default value is.

modules/project-factory/README.md
ludoo: I don't follow: how would you create projects without prefixes?

fast/stages/0-org-setup/schemas/project.schema.json
ludoo: I agree we should be consistent though, so let's maybe decided what we want to do, then align all schemas?

modules/project-factory/README.md
ludoo: Maybe I get it: when CI/CD is done from the app project, I don't see a lot of benefit using the automation block which is specifically intended for when a controlling project exists. In those cases service accounts / buckets are better created directly via the respective blocks, where there's no prefix injection.

ludoo: After this merges I will add support to the stage (not the module) for output files very much like the org setup stage has. CI/CD will be slightly more complex, we need to iterate on a few different approaches for that.

PR #3346, opened 2025-09-21
Added Cloud Build v2 connection module

apichick:
---
**Checklist**


ludoo (approved):

modules/cloud-build-v2-connection/variables.tf
```diff
@@ -0,0 +1,181 @@
...
+
+variable "prefix" {
+  description = "Prefix."
+  type        = string
+  default     = null
+  validation {
+    condition     = var.prefix != ""
+    error_message = "Prefix cannot be empty, please use null instead."
+  }
+}
```
wiktorn: Is the requirement that connection names need to be globally unique? Or do they need to be only unique within the project? If the latter is the case, then I think we don't need the `prefix`

modules/cloud-build-v2-connection/main.tf
```diff
@@ -0,0 +1,173 @@
...
+        invert_regex = each.value.push.invert_regex
+        tag          = each.value.push.tag
+      }
+    }
+    dynamic "pull_request" {
+      for_each = try(each.value.pull_request, null) == null ? [] : [""]
+      content {
+        branch          = each.value.pull_request.branch
+        invert_regex    = each.value.pull_request.invert_regex
+        comment_control = each.value.push.tag
```
wiktorn:
```suggestion
        comment_control = each.value.push.comment_control
```

modules/cloud-build-v2-connection/variables.tf
apichick: not really, I can remove it then.

modules/cloud-build-v2-connection/main.tf
```diff
@@ -0,0 +1,173 @@
...
+  location          = var.location
+  parent_connection = local.name
+  remote_uri        = each.value.remote_uri
+  annotations       = each.value.annotations
+}
+
+resource "google_cloudbuild_trigger" "triggers" {
+  for_each    = local.triggers
+  location    = var.location
+  name        = each.value.trigger_name
```
wiktorn: With this structure, we require users to provide unique names across all defined triggers in different repositories, but we could use here `each.key` and be sure that the name is unique within the project. WDYT?

apichick: sure, actually that's the correct implementation.

modules/cloud-build-v2-connection/variables.tf
apichick: removed, done.

modules/cloud-build-v2-connection/main.tf
apichick: fixed, done

modules/cloud-build-v2-connection/main.tf
apichick: changed, done

Issue #3355, opened 2025-09-25
Project Notifications channels, Output refers to sensitive values


LegendSebastiano-L:
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

I'm adding some Pagerduty notifications channels in the [project-factory](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/project-factory) module, and I have this error:

```
* Failed to execute "tofu plan" in ./.terragrunt-cache/XN-8mEXSCq1in1iFqPgo-euJQUE/J_TmC8qWu6sMX-WGFE7iRDBL0Og/modules/project-factory
  ╷
  │ Error: Output refers to sensitive values
  │ 
  │   on outputs.tf line 29:
  │   29: output "projects" {
  │ 
  │ To reduce the risk of accidentally exporting sensitive data that was
  │ intended to be only internal, OpenTofu requires that any root module output
  │ containing sensitive data be explicitly marked as sensitive, to confirm
  │ your intent.
  │ 
  │ If you do intend to export this data, annotate the output value as
  │ sensitive by adding the following argument:
  │     sensitive = true
```

### Release version

v41.0.0

### Terraform version

v1.12.2

### Steps to reproduce

create a `terragrunt.hcl` with the following content
```
include "root" {
  path = find_in_parent_folders("gcp.hcl")
}

terraform {
  source = "https://github.com/GoogleCloudPlatform/cloud-foundation-fabric.git//modules/project-factory?ref=v41.0.0"
}

inputs = {
  data_defaults = {
    billing_account  = ""
    storage_location = "US"
  }
  data_merges = {}
  data_overrides = {
    logging_data_access = {
      allServices = {
        DATA_READ  = {}
        DATA_WRITE = {}
      }
    }
  }
  factories_config = {
    budgets = {
      billing_account       = ""
      budgets_data_path     = "data/budgets"
      notification_channels = {}
    }
    projects_data_path = "data/projects"
    folders_data_path  = "data/hierarchy"
    context = {}
  }
}
```

and in `./data/hierarchy/tech/project.yaml` the following
```
name: XXX
prefix: "XXX"
deletion_policy: "PREVENT"
auto_create_network: false
parent: folders/XXX

notification_channels:
  pagerduty-api-critical:
    display_name: pagerduty-api-critical
    type: pagerduty
    sensitive_labels:
[...]

PR #3356, opened 2025-09-25
Fix issue/3355: 'Output refers to sensitive values' in project-factory

dlouvier:
Fix issue when outputs contains sensitive values in the project-factory module, `tf plan` fails.

https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/3355

---
**Checklist**

### DIFF
--- a/modules/project-factory/outputs.tf
+++ b/modules/project-factory/outputs.tf
@@ -107,6 +107,7 @@ output "project_numbers" {
 output "projects" {
   description = "Project attributes."
   value       = local.outputs_projects
+  sensitive   = true
 }
 
 output "service_account_emails" {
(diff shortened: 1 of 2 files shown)


modules/project-factory/outputs.tf
```diff
@@ -107,6 +107,7 @@ output "project_numbers" {
 output "projects" {
   description = "Project attributes."
   value       = local.outputs_projects
+  sensitive   = true
```
ludoo, line 110: this will have wide ranging implications for all callers, let me actually run a few tests: I have not hit this during regular usage, and if we have to mark this sensitive we still need to provide as much information as possible in an additional non-sensitive output

juliocc: Alternatively, we could filter out the sensitive fields from `local.output_projects`

dlouvier: Sure, no problem @ludoo 

I also encounter this problem in other module ([net-vpn-ha](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/net-vpn-ha/outputs.tf#L102)) but I just create this change for a single module for now.

Let me know when you complete your tests, if that's ok to create a change request for the other module.

dlouvier: Or if you prefer, I can create and test the change for net-vpn-ha in case this module is more isolated and easy to test.

PR #3365, opened 2025-10-01
Adding support for managed connection pooling in CloudSQL

ramja-google:
Added support for [managed connection pooling](https://cloud.google.com/sql/docs/postgres/configure-mcp#terraform)  in postgres cloud sql enterprise plus instances. This includes changing the main.tf to add the logic and introducing new a variable in  variable.tf file to enable this feature. The various optional flags that accompany the feature are accepted in a map of string.

---
**Checklist**

- [ X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X ] Ran `terraform fmt` on all modified files
- [X ] Made sure all relevant tests pass


modules/cloudsql-instance/main.tf
```diff
@@ -127,6 +127,22 @@ resource "google_sql_database_instance" "primary" {
       }
     }
 
+    dynamic "connection_pool_config" {
+      # Only enable for PostgreSQL on Enterprise Plus edition.
+      for_each = var.managed_connection_pooling_config.enabled && local.is_postgres && var.edition == "ENTERPRISE_PLUS" ? [1] : []
```
juliocc: ensure this doesn't fail if `var.managed_connection_pooling_config` is unset

modules/cloudsql-instance/main.tf
```diff
@@ -127,6 +127,22 @@ resource "google_sql_database_instance" "primary" {
...
+      content {
+        connection_pooling_enabled = true
+        dynamic "flags" {
+          for_each = var.managed_connection_pooling_config.flags != null ? var.managed_connection_pooling_config.flags : {}
```
juliocc: with the proposed type below, the null check is redundant.

modules/cloudsql-instance/variables.tf
```diff
@@ -176,6 +176,19 @@ variable "maintenance_config" {
...
+  description = "Configuration for Managed Connection Pooling. NOTE: This feature is only available for PostgreSQL on Enterprise Plus edition instances."
+  type = object({
+    enabled = bool
+    flags   = map(string)
+  })
+  default = {
+    enabled = false
+    flags   = {}
+  }
+  nullable = false
```
juliocc, line 186: 1. this should include a validation to ensure is only set for the proper edition.
2. a better type for this variable would be
```terraform
variable "managed_connection_pooling_config" {
  type = object({
    enabled = optional(bool, false)
    flags   = optional(map(string), {})
  })
  default = {}
  nullalble = false
}
```

ghost: Thanks for the suggestion. I updated the variable.

modules/cloudsql-instance/main.tf
ghost: Agreed and actioned

modules/cloudsql-instance/main.tf
ghost: It's set as optional with default values..

PR #3373, opened 2025-10-03
Add support for output files to FAST project factory stage

ludoo:
As per recent discussions, this implements a way for the project factory stage to define output files (providers, tfvars) **without any support from the underlying module**. This is preliminary work to add CI/CD support to the stage in the future.

A description of the feature is available in the modified README.


fast/stages/2-project-factory/README.md
```diff
@@ -351,13 +354,121 @@ automation:
         - group:devops@example.org
 ```
 
+### Generated provider and Terraform variables for projects
+
+This stage can optionally be configured to generate provider and tfvars files ("output files") for projects. These files can then be distributed to project owners to help them boostrap automation, and will be used in future releases to configure project-level CI/CD from this factory.
+
+Ooutput file generation is configured in the defaults file, and supports two usage modes:
```
juliocc:
```suggestion
Output file generation is configured in the defaults file, and supports two usage modes:
```

fast/stages/2-project-factory/README.md
```diff
@@ -351,13 +354,121 @@ automation:
...
+      # at least one of the ro or rw matches needs to be defined
+      ro: automation/ro
+      rw: automation/rw
+    # which bucket is used for the provider backend
+    storage_bucket: $storage_buckets:iac-0/iac-shared-state
+    # create managed folders in the bucket by default and set IAM on them
+    # storage_folders_create: true
+```
+
+The above snippet will create one or two provider files, and one tfvars file for each project where automation service accounts matching the pattern are defined.
```
juliocc: both need to match? or either?

ludoo: Thanks, reworded.

ludoo:
```
The above snippet will create zero, one, or two provider files depending on how many service accounts match for each indivdual project. One tfvars file will also be created for each project with at least one provider file.
```

PR #3377, opened 2025-10-05
feat(bigquery-dataset): add optional schema support for views

weather2602:
**Description**

This PR adds support for passing an optional schema attribute to BigQuery views in the bigquery-dataset module.

- Updated variables.tf to include schema as an optional attribute in the views variable.
- Updated main.tf to propagate schema = try(jsonencode(each.value.schema), null) when creating google_bigquery_table resources of type view.
- Regenerated documentation with tools/tfdoc.py.

This enables users to specify field definitions and descriptions for views in native HCL (lists/objects), similar to tables, while maintaining backward compatibility. Internally, the schema is jsonencoded to satisfy the provider requirement.

Checklist
- [X] Made sure all relevant tests pass (ran module-specific pytest examples for `bigquery-dataset` locally with a sandbox GCP project)

Breaking Changes
None — the new argument is optional, so existing configurations continue to work without modification.


ludoo (commented):
Can you add an example to test this?

ludoo (approved):

PR #3378, opened 2025-10-05
Allow forcing jit service agents generation for universe in project and project factory modules

ludoo:
TPC does not automatically trigger service agent creation for compute, and potentially other APIs. This change allow flagging specific service agents from the universe variable, so that they are created even though they are not jit in global GCP.

A block for easier debugging of service agent locals has also been added and commented out to the project module.


modules/project/service-agents.tf
```diff
@@ -16,6 +16,17 @@
...
+
+# resource "terraform_data" "precondition" {
+#   lifecycle {
+#     precondition {
+#       condition     = local.primary_service_agents == null
+#       error_message = yamlencode(local.primary_service_agents)
+#     }
+#   }
+# }
+
```
wiktorn, line 29: Do we want to keep this?

wiktorn (approved):

modules/project/service-agents.tf
ludoo: Yes, until we refactor these locals which are **very** opaque. I explicitly mention adding this in the PR description. :)

PR #3379, opened 2025-10-05
Allow FAST stage 0 provider template to work with universe

ludoo:
This PR adds support for TPC to fast stage 0 provider generation by

- setting the right storage API URL in the backend
- disabling impersonation in the backend (not currently supported)
- setting the universe domain in the provider blocks


fast/stages/0-org-setup/output-files.tf
```diff
@@ -51,6 +51,11 @@ locals {
         id          = local.organization_id
       }
       prefix = local.defaults.prefix
+      universe = try(
+        local.project_defaults.overrides.universe,
+        local.project_defaults.defaults.universe,
+        null
+      )
```
wiktorn, line 58:
```suggestion
      universe = local.of_universe_domain
```

ludoo: universe != universe domain

PR #3380, opened 2025-10-05
Lightly refactor service agents locals in project module

ludoo:
This lightly refactors service agent locals in project module, to improve legibility and replace references from other files to "private" locals.

### DIFF
--- a/modules/project/service-agents.tf
+++ b/modules/project/service-agents.tf
@@ -28,106 +28,115 @@
 # }
 
 locals {
-  services = [
-    for s in distinct(concat(
-      local.available_services, try(var.project_reuse.attributes.services_enabled, [])
-    )) : s if !contains(local._universe_unavailable_si, s)
-  ]
-  _service_agents_data = yamldecode(file("${path.module}/service-agents.yaml"))
-  # map of api => list of agents
-  _service_agents_by_api = {
-    for agent in local._service_agents_data :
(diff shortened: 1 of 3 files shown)


modules/project/service-agents.tf
```diff
@@ -142,16 +141,6 @@ data "google_bigquery_default_service_account" "bq_sa" {
...
-moved {
-  from = google_project_service_identity.jit_si
-  to   = google_project_service_identity.default
-}
-
-moved {
-  from = google_project_service_identity.servicenetworking[0]
-  to   = google_project_service_identity.default["servicenetworking.googleapis.com"]
-}
-
```
juliocc, line 154: I'd keep the move blocks

juliocc (approved):

modules/project/service-agents.tf
ludoo: They are pretty old, you sure we want to keep them? They prevent defining higher level moved blocks on the same resources (eg when migrating FAST)

juliocc: Yes. IMO all moved blocks should be kept indefinitely. 

But your call if they're causing you pain

ludoo: No biggie, I tend to remove them after a few releases, as I had issues in the past. Let's keep these a bit more.

ludoo: done

PR #3382, opened 2025-10-05
Support universe in fast security stage

ludoo:


ludoo:
Argh, some unrelated changes went in

PR #3412, opened 2025-10-14
Add support for context to DNS modules

ludoo:
More requirements for the new net stage.


modules/dns-response-policy/variables.tf
```diff
@@ -21,6 +21,16 @@ variable "clusters" {
   nullable    = false
 }
 
+variable "context" {
+  description = "Context-specific interpolations."
+  type = object({
+    networks    = optional(map(string), {})
+    project_ids = optional(map(string), {})
```
juliocc, line 28: not for now but in the future we can have a `hosts` namespace (or something similar)

ludoo: Yup, I'm adding tests also to make these ready for more complicated contexts. Right now I'm adding what Simo needs, and any obvious namespaces the module can leverage (IAM, location, etc.).

ludoo: Resolving this to merge, but let's absolutely remember to make a deep pass once Simo is done to add any other contexts that fit to modules.

PR #3414, opened 2025-10-15
Add support for context to net-cloudnat, net-firewall-policy modules

ludoo:
Deps for the new network stage.

### DIFF
--- a/modules/secure-source-manager-instance/README.md
+++ b/modules/secure-source-manager-instance/README.md
@@ -11,6 +11,7 @@ This module allows to create a Secure Source Manager instance and repositories i
   - [Private instance](#private-instance)
   - [IAM](#iam)
   - [Branch Protection Rules](#branch-protection-rules)
+  - [Initial Configuration](#initial-configuration)
 - [Variables](#variables)
 - [Outputs](#outputs)
 <!-- END TOC -->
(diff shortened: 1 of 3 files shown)


modules/net-firewall-policy/hierarchical.tf
```diff
@@ -33,23 +33,43 @@ resource "google_compute_firewall_policy_rule" "hierarchical" {
...
+      lookup(local.ctx.networks, n, n)
+    ]
+  )
+  target_service_accounts = (
+    local.rules[each.key].target_service_accounts == null ? null : [
+      for n in local.rules[each.key].target_service_accounts :
+      lookup(local.ctx.iam_principals, n, n)
+    ]
+  )
+  tls_inspect = local.rules[each.key].tls_inspect
```
wiktorn, line 55:
```suggestion
  for_each        = local.use_hierarchical ? local.rules : {}
  firewall_policy = google_compute_firewall_policy.hierarchical[0].name
  action          = each.value.action
  description     = each.value.description
  direction       = each.value.direction
  disabled        = each.value.disabled
  enable_logging  = each.value.enable_logging
  priority        = each.value.priority
  target_resources = (
    each.value.target_resources == null ? null : [
      for n in each.value.target_resources :
      lookup(local.ctx.networks, n, n)
    ]
  )
  target_service_accounts = (
    each.value.target_service_accounts == null ? null : [
      for n in each.value.target_service_accounts :
      lookup(local.ctx.iam_principals, n, n)
    ]
  )
  tls_inspect = each.value.tls_inspect
```

ludoo: commit suggestion to deleted lines is not supported...

good call, let me do it by hand

ludoo: well no, we explicitly note why we don't do it; if you want to re-validate the old assumption, let's open a separate PR

image

PR #3428, opened 2025-10-16
gitignore update

juliocc:


.gitignore
```diff
@@ -1,16 +1,15 @@
 venv/*
-*/venv/*
```
ludoo, line 2: we should keep venv

juliocc: It's redundant with the line just above

juliocc: brought it back just in case :)

PR #3429, opened 2025-10-16
Add Agent Engine module.

LucaPrete:
Add new Agent Engine module.

---
**Checklist**

**Breaking Changes**

```upgrade-note
`terraform-provider-google`:  Bump provider to 7.6.0, to allow use of `google_vertex_ai_reasoning_engine` in modules/agent-engine
```

### DIFF
--- a/modules/agent-engine/README.md
+++ b/modules/agent-engine/README.md
@@ -71,11 +71,14 @@ This example assumes you are providing the [source packages](#packaging-dependen
 
 ```hcl
 module "agent_engine" {
-  source          = "./fabric/modules/agent-engine"
-  name            = "my-agent"
-  project_id      = var.project_id
-  agent_framework = "google-adk"
-  region          = var.region
+  source     = "./fabric/modules/agent-engine"
+  name       = "my-agent"
+  project_id = var.project_id
+  region     = var.region
+
+  agent_engine_config = {
+    agent_framework = "google-adk"
+  }
 
   source_files = {
     path = "assets/src/"
@@ -91,9 +94,12 @@ module "agent_engine" {
   source          = "./fabric/modules/agent-engine"
   name            = "my-agent"
   project_id      = var.project_id
-  agent_framework = "google-adk"
   region          = var.region
   generate_pickle = false
+
+  agent_engine_config = {
+    agent_framework = "google-adk"
+  }
 
   source_files = {
     path = "assets/src/"
@@ -112,11 +118,14 @@ At the moment, you'll need at least to grant to it the `roles/viewer` role.
 
 ```hcl
 module "agent_engine" {
-  source          = "./fabric/modules/agent-engine"
-  name            = "my-agent"
-  project_id      = var.project_id
-  agent_framework = "google-adk"
-  region          = var.region
+  source     = "./fabric/modules/agent-engine"
+  name       = "my-agent"
+  project_id = var.project_id
+  region     = var.region
+
+  agent_engine_config = {
+    agent_framework = "google-adk"
+  }
 
   service_account_config = {
     create = false
@@ -133,11 +142,14 @@ Alternatively, you can use an existing service account.
 
 ```hcl
 module "agent_engine" {
-  source          = "./fabric/modules/agent-engine"
-  name            = "my-agent"
-  project_id      = var.project_id
-  agent_framework = "google-adk"
-  region          = var.region
+  source     = "./fabric/modules/agent-engine"
+  name       = "my-agent"
+  project_id = var.project_id
+  region     = var.region
+
+  agent_engine_config = {
+    agent_framework = "google-adk"
+  }
 
   service_account_config = {
     create = false
@@ -159,12 +171,15 @@ To use KMS keys you'll need to grant the AI Platform Service Agent (`service-YOU
 
 ```hcl
 module "agent_engine" {
-  source          = "./fabric/modules/agent-engine"
-  name            = "my-agent"
-  project_id      = var.project_id
-  agent_framework = "google-adk"
-  region          = var.region
-  encryption_key  = "projects/${var.project_id}/locations/${var.region}/keyRings/my-keyring/cryptoKeys/my-key"
+  source         = "./fabric/modules/agent-engine"
+  name           = "my-agent"
+  project_id     = var.project_id
+  region         = var.region
+  encryption_key = "projects/${var.project_id}/locations/${var.region}/keyRings/my-keyring/cryptoKeys/my-key"
+
+  agent_engine_config = {
+    agent_framework = "google-adk"
+  }
 
   source_files = {
     path = "assets/src/"
@@ -179,19 +194,21 @@ You can define environment variables and load existing secrets as environment va
 
 ```hcl
 module "agent_engine" {
-  source          = "./fabric/modules/agent-engine"
-  name            = "my-agent"
-  project_id      = var.project_id
-  agent_framework = "google-adk"
-  region          = var.region
+  source     = "./fabric/modules/agent-engine"
+  name       = "my-agent"
+  project_id = var.project_id
+  region     = var.region
 
-  environment_variables = {
-    FOO = "my-foo-variable"
-  }
+  agent_engine_config = {
+    agent_framework = "google-adk"
 
-  secret_environment_variables = {
-    BAR = {
-      secret_id = "projects/YOUR_PROJECT_NUMBER/secrets/my-bar-secret"
+    environment_variables = {
+      FOO = "my-foo-variable"
+    }
+    secret_environment_variables = {
+      BAR = {
+        secret_id = "projects/YOUR_PROJECT_NUMBER/secrets/my-bar-secret"
+      }
     }
   }
 


modules/agent_engine/main.tf
```diff
@@ -0,0 +1,208 @@
...
+  depends_on = [
+    null_resource.default
+  ]
+}
+
+resource "google_storage_bucket_object" "pickle" {
+  count  = var.generate_pickle ? 0 : 1
+  name   = "${filesha1("${var.source_files.path}/${var.source_files.pickle_out}")}.pkl"
+  bucket = local.bucket_name
+  source = "${var.source_files.path}/${var.source_files.pickle_out}"
```
wiktorn: Please add also `source_md5hash` - to calculate it properly you may need to add `local_file` resource that depends on your `null_resource`.

This is the only way to make sure, that if you update this, it won't crash the provider.

modules/agent_engine/main.tf
```diff
@@ -0,0 +1,208 @@
...
+}
+
+resource "google_storage_bucket_object" "requirements" {
+  name   = "${filesha1("${var.source_files.path}/${var.source_files.requirements}")}.txt"
+  bucket = local.bucket_name
+  source = "${var.source_files.path}/${var.source_files.requirements}"
```
wiktorn: Add here `source_md5hash` as well, though here `filemd5()` is enough, as this file is not changing during plan / apply.

modules/agent_engine/main.tf
LucaPrete: you meant this wiktor? https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_object#detect_md5hash-1

LucaPrete: so you're saying I can keep the normal name (i.e. pickle.pkl) and use detect_md5hash-1 = true ...right?

modules/agent_engine/main.tf
LucaPrete: Actually the file may change throughout updates, that's why I use filesha1 also here...so it would make sense to use instead also here detect_md5hash.

modules/agent_engine/main.tf
wiktorn: I mean the warning that's there, and [source_md5](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_object#source_md5hash-1) field.

No, I mean creating:
```hcl
resource "local_file" pickle {
 path = "to/pickle"
}
...

resource "google_storage_bucket_object" "..." {
  ...
  source_md5hash = local_file.pickle.content_md5
  ...
}
```

modules/agent-engine/main.tf
```diff
@@ -0,0 +1,212 @@
...
+  count          = var.generate_pickle ? 1 : 0
+  name           = "pickle.pkl"
+  bucket         = local.bucket_name
+  source         = "${var.source_files.path}/${var.source_files.pickle_out}"
+  source_md5hash = filemd5("${var.source_files.path}/${var.source_files.pickle_out}")
+
+  depends_on = [
+    null_resource.default
+  ]
+}
```
wiktorn, line 179:
```suggestion
resource "local_file" "pickle_out" {
  filename = "${var.source_files.path}/${var.source_files.pickle_out}"
  depends_on = [ null_resource.default ]
}

resource "google_storage_bucket_object" "pickle_from_src" {
  count          = var.generate_pickle ? 1 : 0
  name           = "pickle.pkl"
  bucket         = local.bucket_name
  source         = "${var.source_files.path}/${var.source_files.pickle_out}"
  source_md5hash = local_file.pickle_out.content_md5
}
```

You need to create a separate `local_file`, otherwise terraform will call `filemd5()` during plan, and during apply this value will change, as you will regenerate the file. This is how you get inconsistent plan provider error (which in this case, is actually user error).

tests/modules/agent_engine/assets/src/pickle.pkl
juliocc, line 1: I wouldn't distribute a pickle file. Can you do without this file?

juliocc (changes requested):
A few comments:
- the module needs a versions.tf/tofu
- make sure the python code follows the yapf conventions
- module names use dashes (you're using underscore)
- can you avoid distributing pickle and tar.gz files?

modules/agent-engine/main.tf
LucaPrete: @wiktorn did all the tests. The actual version works consistently both across plans and applies.

tests/modules/agent_engine/assets/src/pickle.pkl
LucaPrete: the module needs a versions.tf/tofu - **Done**
make sure the python code follows the yapf conventions **Done I think yapf wasn't complaining even before**
module names use dashes (you're using underscore) **Done**
can you avoid distributing pickle and tar.gz files?

No. The pickle file is essential to make reason engine works. The one in the repo (only in the test folder) is there to execute the tests. If you prefer, given during tests we only execute a plan, we may be able to substitute it with a random text file. This is true for the tar.gz as well. I'm not getting the concern. You don't want these files to be opaque to the users that downlaod the repo?

modules/agent-engine/main.tf
```diff
@@ -0,0 +1,212 @@
...
+    local.ctx.project_ids, var.project_id, var.project_id
+  )
+  service_account_email = (
+    var.service_account_config.create
+    ? google_service_account.default[0].email
+    : local._service_account_external_email
+  )
+  service_account_roles = [
+    for role in var.service_account_config.roles
+    : lookup(var.context.custom_roles, role, role)
```
wiktorn:
```suggestion
    : lookup(local.ctx.custom_roles, role, role)
```
?

modules/agent-engine/README.md
```diff
@@ -0,0 +1,233 @@
...
+  name            = "my-agent"
+  project_id      = var.project_id
+  agent_framework = "google-adk"
+  region          = var.region
+
+  source_files = {
+    path = "assets/src/"
+  }
+}
+# tftest inventory=minimal.yaml skip-tofu
```
LucaPrete: I initially disabled tofu tests because I thought the problem we're having with tests was related to it.
Once we fix terraform tests I'll re-enable it and see how it goes. Likely will need to stay enabled.

modules/agent-engine/README.md
```diff
@@ -0,0 +1,234 @@
...
```
wiktorn:
```suggestion
# tftest inventory=minimal.yaml
```

Why are you skipping tests for tofu? The tests are working.

modules/agent-engine/versions.tf
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+
+# Fabric release: v45.0.0
+
+terraform {
+  required_version = ">= 1.12.2"
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 7.0.1, < 8.0.0" # tftest
```
wiktorn: `google_vertex_ai_reasoning_engine` was introduced in 7.0.6, so you need to bump the version

modules/agent-engine/README.md
LucaPrete: Read the comment I left already :)

modules/agent-engine/versions.tf
LucaPrete: @wiktorn I don't understand.. shouldn't this download the latest provider between 7.0.1 and 8.0.0, so bump the version automatically? Is this something specific to out caching layer?

wiktorn: It allows any version between 7.0.1 and 8.0.0. It will download latest, if there are no other restrictions, but it may also download 7.0.1. (If within your root module there are some other restrictions).

So, to make sure, that this works no matter what, we need to name here *minimum* requirements. And minimum for this module is 7.0.6.

modules/agent-engine/outputs.tf
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+output "id" {
+  description = "Fully qualified Agent Engine id."
+  value       = google_vertex_ai_reasoning_engine.default.id
+}
```
wiktorn, line 20: Maybe it's also worth adding service account, if it is created?
```hcl
output "service_account" {
  description = "Service account resource."
  value       = try(google_service_account.service_account[0], null)
}

output "service_account_email" {
  description = "Service account email."
  value       = local.service_account_email
}

output "service_account_iam_email" {
  description = "Service account email."
  value = join("", [
    "serviceAccount:",
    local.service_account_email == null ? "" : local.service_account_email
  ])
}
```
?

modules/agent-engine/versions.tf
LucaPrete: got it. Will do.

modules/agent-engine/versions.tf
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+
+# Fabric release: v45.0.0
+
+terraform {
+  required_version = ">= 1.12.2"
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 7.0.6, < 8.0.0" # tftest
```
wiktorn:
```suggestion
      version = ">= 7.6.0, < 8.0.0" # tftest
```
my bad!

wiktorn: This also should result in lintng failure, as you need to bump across all modules.. Let me check why it is not failing :-)

juliocc (approved):

tests/modules/agent_engine/assets/src/pickle.pkl
LucaPrete: Solved with jullio offline.

PR #3435, opened 2025-10-17
Factory based FAST Networking stage

sruffilli:
New factory based networking stage, shipping with a single dataset (peering) to keep the PR size somewhat manageable.


juliocc (approved):
LGTM. I tested it and everything looks good.

Just one small comment. I think there's a new empty file: `adrs/fast/add-stages.md`

sruffilli:
> Just one small comment. I think there's a new empty file: `adrs/fast/add-stages.md`

WAI - I simply added the new stage to it since it references it.

PR #3438, opened 2025-10-17
Add PAM support

juliocc:


modules/folder/schemas/pam-entitlements.schema.json
```diff
@@ -0,0 +1,109 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "patternProperties": {
+        "^[a-z][a-z0-9-]{0,61}[a-z0-9]$": {
+            "type": "object",
+            "properties": {
```
ludoo, line 7: I would add `  "additionalProperties": false` to every object ;)

juliocc: Done!

PR #3450, opened 2025-10-22
Allow skipping data source in service account module

ludoo:
This aligns the service account module interface to the new `reuse` interface already adopted in the project module, allowing to optionally skip the data source when resource creation is not required.

All service account attributes are already statically defined in the module, with the exception of `unique_id` which is only needed for tag bindings. The choice with this change is to give users the option of force skipping the data source even when unique id is not available, and the only drawback is that tag bindings are not applied in those cases.

**Breaking Changes**

```upgrade-note
`modules/iam-service-account`:  The `service_account_create` variable has been renamed to `service_account_reuse` and its type has changed.
`modules/project-factory` and `fast/stages/0-org-setup`:  Data sources for service accounts are no longer needed.
```

### DIFF
--- a/modules/iam-service-account/main.tf
+++ b/modules/iam-service-account/main.tf
@@ -50,6 +50,7 @@ locals {
   sa_domain = join(".", compact([
     local.project_id_no_universe, local.universe
   ]))
+  # the condition here avoids referring to attributes not known at plan time
   tag_bindings = (
     var.service_account_reuse != null ||
     try(var.service_account_reuse.attributes.unique_id, null) != null


modules/iam-service-account/outputs.tf
```diff
@@ -50,3 +50,11 @@ output "service_account" {
   description = "Service account resource."
   value       = local.service_account
 }
+
+output "unique_id" {
+  description = "Fully qualified service account id."
+  value       = local.service_account.unique_id
+  depends_on = [
+    local.service_account
+  ]
```
wiktorn: From my understanding, this should not be needed, because you are accessing `local.service_account.unique_id`, and this already creates implicit dependency to `local.service_account`

ludoo: yup, I just aligned to all other outputs

ludoo: removed

PR #3452, opened 2025-10-22
Allow null project id in service account module when reusing service account

ludoo:
The service account module already allows passing a full fledged email in the `name` variable when an existing service account is reused. This PR extends that behaviour so as to allow passing a null project id, and deriving it from the email.

This also fixes a logic error in #3450. Further PRs might test the new feature in different universes, but for now this is extra behaviour only designed for GCP.

### DIFF
--- a/modules/iam-service-account/iam.tf
+++ b/modules/iam-service-account/iam.tf
@@ -89,9 +89,11 @@ locals {
 }
 
 resource "google_service_account_iam_binding" "authoritative" {
-  for_each           = local.iam
-  service_account_id = local.service_account.name
-  role               = lookup(local.ctx.custom_roles, each.key, each.key)
+  for_each = local.iam
+  service_account_id = try(
+    local.service_account.name, local.static_id
+  )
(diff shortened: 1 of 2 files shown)


modules/iam-service-account/main.tf
```diff
@@ -26,11 +26,17 @@ locals {
...
     : local.static_iam_email
   )
-  name       = split("@", var.name)[0]
-  prefix     = var.prefix == null ? "" : "${var.prefix}-"
-  project_id = lookup(local.ctx.project_ids, var.project_id, var.project_id)
+  name   = split("@", var.name)[0]
+  prefix = var.prefix == null ? "" : "${var.prefix}-"
+  project_id = (
+    var.project_id == null
+    ? try(regex("^[^@]+@([^.]+)", var.name)[0], null)
```
sruffilli, line 34: PEBKAC - but took me a good 30 seconds to understand the logic. Wouldn't change it, but add a comment for your future self 😆

ludoo: done

PR #3471, opened 2025-10-26
Streamline stage variables and output files for vpc-sc and security stages

ludoo:
This streamlines the VPC SC and security stages so that both now use the `defaults` file for static context population and to define output files, in the same way as the org setup stage.

The defaults file also implements an optional stage name attribute, which can be used when a stage is run with different states (e.g. tenants, per-environment setups, etc.) and which controls output file names.

VPC-SC standard defaults.

image

Security standard defaults.

image


wiktorn (approved):

ludoo:
This is pending, will resume work after release as the changes need to also encompass vpcsc, net, and stage 3s

PR #3512, opened 2025-11-07
Spanner Instance Module minor fixes and improvements

frandarre14:
Working on a project with @mauriciowillig we've noticed that the spanner_instance module was not passing the variable to the Spanner Instance Database resource within, so it was defaulting on creating the database with GOOGLE_CLOUD_SQL dialect and ignoring POSTGRES even if added as a module variable.

We've also noticed that there was no support to input instance settings like database edition, type and default backup schedule type, so we've added those as well.

---
**Checklist**


ludoo (approved):

modules/spanner-instance/main.tf
```diff
@@ -55,12 +55,15 @@ resource "google_spanner_instance" "spanner_instance" {
...
-  processing_units = var.instance.processing_units
+  name                         = var.instance.name
+  display_name                 = coalesce(var.instance.display_name, var.instance.name)
+  num_nodes                    = var.instance.num_nodes
+  labels                       = var.instance.labels
+  force_destroy                = var.instance.force_destroy
+  processing_units             = var.instance.processing_units
+  edition                      = var.instance.instance_settings.edition
+  instance_type                = var.instance.instance_settings.instance_type
+  default_backup_schedule_type = var.instance.instance_settings.default_backup_schedule_type
```
juliocc, line 66: `instance_settings` is optional, these 3 accesses should use `try()`.

modules/spanner-instance/variables.tf
```diff
@@ -89,6 +89,11 @@ variable "iam_bindings_additive" {
 variable "instance" {
   description = "Instance attributes."
   type = object({
+    instance_settings = optional(object({
+      default_backup_schedule_type = optional(string)
+      edition                      = optional(string)
+      instance_type                = optional(string)
+    }))
```
juliocc, line 96: do you really need a a new nested field here? You could make these 3 attributes top level (and you avoid the `try` mentioned in my previous comment)

PR #3516, opened 2025-11-10
Allow configuring data access logs from org/folder/project schemas

ludoo:
This implements support for data access logs configuration to

- the organization in the FAST org setup stage
- folders anywhere the project factory is used

Project support has been changed so a single interface now works across all schemas. Context expansion for principal exclusions has also been implemented in modules.

The existing project-level default and override for data access logs have been removed, as that type of configuration is best done via folder-level configuration. A project template can still be used to achieve much the same as the previous default, if needed.

**Breaking Changes**

```upgrade-note
`modules/project-factory`: the `logging_data_access` project default and override have been removed, if you need project-wide configuration of data access logs use a folder-level configuration or a project template.
```


modules/organization/logging.tf
```diff
@@ -74,8 +74,11 @@ resource "google_organization_iam_audit_config" "default" {
   dynamic "audit_log_config" {
     for_each = { for k, v in each.value : k => v if v != null }
     content {
-      log_type         = audit_log_config.key
-      exempted_members = audit_log_config.value.exempted_members
+      log_type = audit_log_config.key
+      exempted_members = try([
```
juliocc: This try will hide a lot of errors. What about moving it to the for?

```for x in try(xxx, [])```

File: modules/project-factory/variables.tf [hunk 2]
```diff
@@ -139,11 +134,6 @@ variable "data_overrides" {
       logging  = optional(string)
       storage  = optional(string)
     }), {})
-    logging_data_access = optional(map(object({
-      ADMIN_READ = optional(object({ exempted_members = optional(list(string)) })),
-      DATA_READ  = optional(object({ exempted_members = optional(list(string)) })),
-      DATA_WRITE = optional(object({ exempted_members = optional(list(string)) }))
-    })))
```
juliocc, line 146: You're removing the default and override. That's ok?

modules/organization/logging.tf
ludoo: I want to preserve the `null` if possible, but yes we can also do that.

modules/project-factory/variables.tf
ludoo: it's noted in the PR comment. :)

modules/organization/logging.tf
ludoo: Tested and it made no difference, so went with the cleaner solution you suggested.

PR #3517, opened 2025-11-10
Use dns_policy default value instead of null when creating FAST VPCs

vannicktrinquier:
In module net-vpc, `var.dns_policy` default value is null and not {}.
See https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/master/modules/net-vpc

image

The fix is to use`{}` instead of `null`, to avoid to have blank`google_dns_policy` policy created which is not expected behavior. Expected behavior being that if not var.dns_policy is provided, nothing is created.

FYI the net-vpc module is creating google_dns_policy as follow 
```  count                     = var.dns_policy == null ? 0 : 1```

Already discussed with @sruffilli on that


ludoo:
I don't think you should change the default, as it reflects the underlying module's default

image


Can you explain how you would be able to disable the DNS policy with your change?

ludoo (changes requested):
Misread the change. :)

PR #3523, opened 2025-11-11
Adds support for custom learned routes to net-vpn-ha

sruffilli:
- Support for custom learned routes (it only exists at the BGP session level, specifically google_compute_router_peer
- Cleaned up BFD support (removed) in the examples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,32 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] <!-- from: 2025-11-05 08:29:41+00:00 to: None since: v47.0.0 -->
 
+## [48.0.0] - 2025-11-11
+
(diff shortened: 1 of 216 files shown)


modules/net-vpn-ha/main.tf
```diff
@@ -131,17 +131,32 @@ resource "google_compute_router_peer" "bgp_peer" {
...
   }
+
+  dynamic "custom_learned_ip_ranges" {
+    for_each = try(each.value.bgp_peer.custom_learned_ip_ranges.ip_ranges, {})
+    iterator = range
+    content {
+      range = range.value
+    }
+  }
+
```
ludoo: nit: I dislike newlines within a single resource block (feel free to ignore) :)

sruffilli: ✅

PR #3529, opened 2025-11-14
Add optional secondary read pool support to the alloydb module

sshcherbakov:
**Add optional secondary read pool support to the alloydb module.**

It is now possible to configure and create secondary read pools replicating from the secondary instance using the alloydb module. The configuration is same as existing primary read pool configuration but available under the `cross_region_replication`, e.g:

```
 module "my_alloydb" {
  # source         = "github.com/terraform-google-modules/cloud-foundation-fabric//modules/alloydb?ref=v48.0.0"
  source            = "../../modules/alloydb"
  project_id     = var.project_id
  location       = var.location_primary
  cluster_name   = "${var.name}-primary"
  instance_name  = "${var.name}-primary"
  machine_config = {
    cpu_count    = 2
    machine_type = var.machine_type
  }
  initial_user = {
    password = var.initial_password
  }
  network_config = {
    psc_config = { allowed_consumer_projects = [ data.google_project.client_project.number ] }
  }
  database_version = var.database_version
  read_pool = {
    "${var.name}-read-pool" = {
      node_count = var.read_pool_count
    }
  }
  cross_region_replication = {
    enabled = true
    region  = var.location_secondary
    secondary_cluster_name = "${var.name}-secondary"
    secondary_instance_name = "${var.name}-secondary"
    read_pool = {
      "${var.name}-secondary-read-pool" = {
        node_count = var.read_pool_count
      }
    }
  }

  # Set to true in the real production environment
  deletion_protection = false
}

```

---
**Checklist**

### DIFF
--- a/modules/alloydb/README.md
+++ b/modules/alloydb/README.md
@@ -177,7 +177,7 @@ module "alloydb" {
 
   deletion_protection = false
 }
-# tftest modules=1 resources=4 inventory=cross_region_read_pools.yaml e2e
+# tftest inventory=cross_region_read_pools.yaml e2e
 ```
 
 
--- a/tests/modules/alloydb/examples/cross_region_read_pools.yaml
+++ b/tests/modules/alloydb/examples/cross_region_read_pools.yaml
@@ -23,7 +23,10 @@ values:
       recovery_window_days: 14
     database_version: POSTGRES_15
     deletion_policy: DEFAULT
+    deletion_protection: false
     display_name: db
+    effective_labels:
+      goog-terraform-provisioned: 'true'
     encryption_config: []
     etag: null
     initial_user:
@@ -42,6 +45,8 @@ values:
     secondary_config: []
     skip_await_major_version_upgrade: true
     subscription_type: STANDARD
+    terraform_labels:
+      goog-terraform-provisioned: 'true'
     timeouts: null
   module.alloydb.google_alloydb_cluster.secondary[0]:
     annotations: null
@@ -53,7 +58,10 @@ values:
       recovery_window_days: 14
     database_version: POSTGRES_15
     deletion_policy: FORCE
+    deletion_protection: false
     display_name: db-sec
+    effective_labels:
+      goog-terraform-provisioned: 'true'
     encryption_config: []
     etag: null
     initial_user: []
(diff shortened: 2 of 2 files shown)


sshcherbakov:
I have also removed `google-beta` providers in the alloydb module (effectively replacing it with `google` provider), since all needed functionality seems to be already available in the `google` provider.

wiktorn:
Thanks, can you add your example to the README, so it will get tested as well?

modules/alloydb/main.tf
```diff
@@ -183,7 +191,6 @@ resource "google_alloydb_cluster" "primary" {
 }
 
 resource "google_alloydb_instance" "primary" {
-  provider          = google-beta
```
juliocc, line 186: This will force recreation of the resource. Seems really dangerous for a data-related module/resource.

I suggest keeping all the google-beta as they are.

modules/alloydb/main.tf
```diff
@@ -434,14 +440,12 @@ resource "google_alloydb_instance" "secondary" {
...
-# Read pool (instance_type = "READ_POOL") cannot be created for secondary cluster
-# and does not support the following attributes:
+# Read pool (instance_type = "READ_POOL") does not support the following attributes:
 # * availability_type: Because 1 node pool (read_pool_config.node_count) is always zonal, two or more is always regional.
 # * gce_zone
 # * network_config.enable_outbound_public_ip
-resource "google_alloydb_instance" "read_pool" {
-  provider       = google-beta
-  for_each       = local.read_pool
+resource "google_alloydb_instance" "read_pool_primary" {
```
juliocc, line 454: this rename will also cause a recreate. A moved block could help here (but only if you don't change providers)

sshcherbakov: Mm.. I am just checking locally and changing between "google" and "google-beta" providers does't cause recreation of neither instances nor read pools. 
Could you please confirm the plan to leave "google-beta" as is?

modules/alloydb/main.tf
sshcherbakov: Mm.. I am just checking locally and changing between "google" and "google-beta" providers does't cause recreation of neither instances nor read pools.
Could you please confirm the plan to leave "google-beta" as is?

modules/alloydb/main.tf
sshcherbakov: Adding a moved block.

sshcherbakov: Reverted back the provider.

modules/alloydb/main.tf
sshcherbakov: Reverted back the provider.

Issue #3553, opened 2025-11-27
Project Factory - billing_budgets definition inside project.yaml does not seems to be working


ysolt:
### Affected components

- [ ] Fabric modules
- [x] FAST
- [ ] Blueprints
- [ ] Other

### The problem

The fabric-light / project factory v44 upgrade seems to break the original budget configuration which was working well with v42.0.0

We did look through the current implementation but can't find a way to specify budget configuration from the project and not from the billing budget itself. 

Can you please help me here by either providing us a working project.yaml example or validate if the environment passing between project factory stage -> project factory module -> billing account module works as expected.

Thanks a lot!

### Release version

v44.0.0 and upwards

### Terraform version

v1.13.4

### Steps to reproduce

Setup (project factory)

`data/budgets/sandbox.yaml`
```
amount:
  units: 500
display_name: Sandbox project billing budget
threshold_rules:
  - percent: 0.5
  - percent: 0.9
  - percent: 1
filter:
  period:
    calendar: "WEEK"
```

`data/projects/sandbox-01.yaml`
```
# yaml-language-server: $schema=../../schemas/project.schema.json

parent: $folder_ids:sandbox
services:
  - storage.googleapis.com

billing_budgets:
  - $billing_budgets:sandbox
```

### Expected behavior

Automatically generate project id to the respective billing budget

```
  # module.factory.module.billing-budgets[0].google_billing_budget.default["sandbox"] will be updated in-place
  ~ resource "google_billing_budget" "default" {
        id              = "billingAccounts/0xxxx2-4xxxx4-0xxx4/budgets/fff97ff8-6329-4a24-8ab9-723536509945"
        name            = "fff97ff8-6329-4a24-8ab9-723536509945"
        # (3 unchanged attributes hidden)

      ~ budget_filter {
          ~ projects               = [
                "projects/611828501134",
            ]
            # (7 unchanged attributes hidden)
        }

        # (4 unchanged blocks hidden)
```

### Actual behavior

Project id is getting removed from v42.0.0 -> v49.0.0 upgrade

```
  # module.factory.module.billing-budgets[0].google_billing_budget.default["sandbox"] will be updated in-place
  ~ resource "google_billing_budget" "default" {
        id              = "billingAccounts/0xxxx2-4xxxx4-0xxx4/budgets/fff97ff8-6329-4a24-8ab9-723536509945"
        name            = "fff97ff8-6329-4a24-8ab9-723536509945"
[...]

PR #3555, opened 2025-11-28
Use project numbers in billing budget filter

kovagoadam:
Modified `budget_filter` for billing budget
---
`budget_filter` resource does not take `project_id` instead it needs the project number.

Related to #3553 #3554, but that is for project factory and this is for the billing budget itself. 

**Checklist**

### DIFF
--- a/tests/modules/billing_account/context.tfvars
+++ b/tests/modules/billing_account/context.tfvars
@@ -9,7 +9,7 @@ budgets = {
       period = {
         calendar = "MONTH"
       }
-      projects           = ["$project_ids:foo"]
+      projects           = ["$project_numbers:foo"]
       resource_ancestors = ["$folder_ids:bar"]
     }
     threshold_rules = [
@@ -47,6 +47,9 @@ context = {
   }
   project_ids = {
     foo = "test-prj-foo"
+  }
+  project_numbers = {
+    foo = 1234567890
   }
 }
 iam = {
--- a/tests/modules/billing_account/context.yaml
+++ b/tests/modules/billing_account/context.yaml
@@ -73,7 +73,7 @@ values:
       credit_types_treatment: INCLUDE_ALL_CREDITS
       custom_period: []
       projects:
-      - test-prj-foo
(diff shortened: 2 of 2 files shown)


modules/billing-account/budgets.tf
```diff
@@ -22,6 +22,10 @@ locals {
       "$notification_channels:${k}" => v.id
     }
   )
+  project_numbers = {
```
ludoo: `local.ctx` will not contain the internally defined project numbers, you have to merge them in first.

Check https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/project-factory/projects.tf#L51 for comparison

ludoo: I also don't think we should change the semantics of the project numbers context to have a `projects/` prefix: the billing module should interpolate that. We can't expect the caller to know if project numbers should be plain numbers (like in all other modules), or prefixed.

I would do something like this in the billing module instead: `try("projects/${local.ctx.project_numbers[v]}", v)`

modules/billing-account/budgets.tf
```diff
@@ -99,7 +99,10 @@ resource "google_billing_budget" "default" {
     projects = concat(
       [
         for v in each.value.filter.projects :
-        lookup(local.ctx.project_ids, v, v)
+        try(
+          "projects/${local.ctx.project_numbers[v]}",
+          lookup(local.ctx.project_numbers, v, v)
```
ludoo:
```hcl
try("projects/${local.ctx.project_numbers[v]}", v)
```

As `project_numbers` values will never have a `projects/` prefix.

ludoo: You don't need to nest trys :)

ludoo (dismissed):

modules/billing-account/budgets.tf
ludoo: ouch, github tricked me, ignore

ludoo (approved):

PR #3556, opened 2025-11-30
Add controls related to CMEK encryption and support to FAST

vannicktrinquier:


fast/stages/2-project-factory/variables-fast.tf
```diff
@@ -97,6 +97,14 @@ variable "project_ids" {
   default     = {}
 }
 
+variable "security_project_ids" {
```
ludoo: I don't like this too much, but I know there's a problem passing the same type of information from multiple stages that we eventually need to address.

But doing it like this will only make the problem worse: what if you need project numbers at some point? etc. etc.

WDYT about bundling security outputs in a `fast_security` variable with `project_ids`, `project_numbers`, `kms_keys`, etc?

```hcl
variable "fast_security" {
  type = object({
    kms_key_ids = optional(map(string), {})
    project_ids = optional(map(string), {})
    project_numbers = optional(map(string), {})
    # etc
```

ludoo: But looking at your code, why do you even need to export project ids from the security stage?

I would simply remove this.

fast/stages/2-project-factory/main.tf
```diff
@@ -88,7 +88,7 @@ module "factory" {
     locations             = local.context.locations
     notification_channels = local.context.notification_channels
     project_ids = merge(
-      var.project_ids, var.host_project_ids, local.context.project_ids
+      var.project_ids, var.host_project_ids, var.security_project_ids, local.context.project_ids
```
ludoo: Why are you adding security projects?

modules/project-factory/projects.tf
```diff
@@ -182,3 +180,27 @@ module "projects-iam" {
   shared_vpc_service_config = each.value.shared_vpc_service_config
   universe                  = each.value.universe
 }
+
+# observability require IAM and service agent configured when using CMEK encryption
+module "projects-observability" {
+  source   = "../project"
```
ludoo: We should find a way of fitting this into the existing two project modules. Please let's discuss why you need to do this.

ludoo (changes requested):
This PR is too large and tackles too many things, some of which need a design discussion.

Can you please split it so:

- one PR adds support for KMS keys contexts to modules
- one PR implements any change you need in the stage 0 dataset

Then we discuss why you need to add projects to the security stage, and why you're adding an additional project module call to the project factory. None of those look ok from my PoV.

modules/project-factory/projects.tf
vannicktrinquier: We got a circular dependency between cmek permissions, log bucket creation using cmek and log based metrics using context, and context based on log bucket creation. 

Something like:
1. log based metrics require log buckets and use log bucket context
2. log bucket context is build as follow (need `module.log-buckets`)
```
  ctx_log_buckets = merge(local.ctx.log_buckets, local.log_buckets)
  log_buckets = {
    for key, log_bucket in module.log-buckets : key => log_bucket.id
  }
```
3. log bucket require cmek encryption key and iam stuff done for cmek so it depends_on `module.projects-iam` to get the expected permissions to role can be assigned to service agent. If log bucket is created before module.projects-iam, it will fails because cmek permission
4. `module.project-iam` depends on log bucket context .... (as done initially)

So this was to break the circular dependency on log bucket which need permissions on cmek before being created. So log based metrics are created once bucket are created and once cmek is setup with iam

Let me try to see during the week and deep dive if i can avoid it. Open to ideas if any

fast/stages/2-project-factory/variables-fast.tf
vannicktrinquier: I dont need it anymore so i will remove it. 

It was needed in my first draft ... but finally i find cleaner way by moving creation of security folder to stage 0 (which allow me to set the org policy using folder id without needing the project created).

fast/stages/2-project-factory/main.tf
vannicktrinquier: Not needed anymore, i will remove it

vannicktrinquier: Removed (not needed anymore)

fast/stages/2-project-factory/variables-fast.tf
vannicktrinquier: Removed (not needed anymore)

fast/stages/0-org-setup/datasets/hardened/organization/.config.yaml
```diff
@@ -81,19 +81,35 @@ iam_by_principals:
     - $custom_roles:organization_admin_viewer
     - $custom_roles:tag_viewer
   $iam_principals:service_accounts/iac-0/iac-networking-rw:
+    # uncomment for cooperative VPC-SC configurations
```
vannicktrinquier, line 84: Sync from classic datasets that were not retrofit previously to hardened

fast/stages/0-org-setup/variables.tf
```diff
@@ -43,6 +43,7 @@ variable "factories_config" {
     cicd_workflows    = optional(string)
     defaults          = optional(string, "datasets/classic/defaults.yaml")
     folders           = optional(string, "datasets/classic/folders")
+    observability     = optional(string)
```
vannicktrinquier: New factories config which contains the list of monitoring alerts and log based metrics. The `defaults.yaml` above is used to know where to deploy the controls.

fast/stages/2-project-factory/data/projects/dev-app-a-0.yaml
```diff
@@ -15,6 +15,8 @@
 # yaml-language-server: $schema=../../schemas/project.schema.json
 
 parent: $folder_ids:team-a/dev
+services:
+  - compute.googleapis.com
 shared_vpc_service_config:
```
vannicktrinquier, line 20: Added as if not present, the service project cannot be added to shared VPC. (this was not the case few weeks ago but tested multiple times, and needed now to not have issue)

fast/stages/0-org-setup/datasets/classic/folders/security/dev/.config.yaml
vannicktrinquier, line 1: Create of folder moves to stage 0. This make every easier for handling organization policies from the where the KMS keys have to be created by using `under:<DEV_SEC_PROJECT` which is know at stage 0 already now

fast/stages/0-org-setup/datasets/hardened/observability/auditConfigChanges.yaml
vannicktrinquier, line 1: Moved to new path

fast/stages/3-data-platform-dev/data-domains-automation.tf
```diff
@@ -16,19 +16,34 @@
 
 # tfdoc:file:description Data product automation resources.
 
-module "dd-automation-bucket" {
-  source = "../../../modules/gcs"
-  for_each = {
+locals {
+  dd_automation = {
     for k, v in local.data_domains :
```
vannicktrinquier, line 21: Add support in stage 3 data platform for using CMEK keys every where (was missing initially)

fast/stages/3-gke-dev/main.tf
```diff
@@ -26,6 +26,23 @@ locals {
...
+  _cmek_keys_pubsub = toset(compact(flatten([
+    [for k, v in var.clusters : try(v.enable_features.upgrade_notifications.kms_key_name, null)],
+  ])))
+  service_encryption_key_ids = {
+    for k, v in {
+      "container.googleapis.com" = local._cmek_keys_container
+      "pubsub.googleapis.com"    = local._cmek_keys_pubsub
+    } : k => v if length(v) > 0
+  }
 }
```
vannicktrinquier, line 46: Add support to stage 3 GKE to use CMEK everywhere (gke cluster, nodepool, etcd encryption, pubsub ...)

modules/project-factory/projects.tf
```diff
@@ -150,21 +146,19 @@ module "projects-iam" {
...
-    log_buckets = local.ctx_log_buckets
     project_ids = merge(
       local.ctx.project_ids,
       { for k, v in module.projects : k => v.project_id }
     )
   })
   factories_config = {
     # we do anything that can refer to IAM and custom roles in this call
-    observability    = try(each.value.factories_config.observability, null)
     pam_entitlements = try(each.value.factories_config.pam_entitlements, null)
```
vannicktrinquier, line 162: Moved from module to stage 0 as per discussion.
This avoid to have change to project-factory module (adding new call to project module).

modules/project/service-agents.tf
```diff
@@ -139,6 +139,12 @@ data "google_bigquery_default_service_account" "bq_sa" {
   depends_on = [google_project_service.project_services]
 }
 
+data "google_logging_project_settings" "logging_sa" {
+  count      = contains(var.services, "logging.googleapis.com") ? 1 : 0
+  project    = local.project.project_id
+  depends_on = [google_project_service.project_services]
+}
+
 moved {
```
vannicktrinquier, line 148: As done for cloud storage or bigquery, logging is a legacy case where service agent are created only after calling data. google_logging_project_settings

This is needed to ensure logging bucket have the agent created before we assign encrypter/decrypter permission

tests/fast/stages/s0_org_setup/data-hardened/defaults.yaml
```diff
@@ -0,0 +1,68 @@
...
+    id: 1234567890
+    customer_id: abcd123456
+projects:
+  defaults:
+    prefix: ft0
+    locations:
+      bigquery: $locations:primary
+      logging: $locations:primary
+      storage: $locations:primary
+    observability:
```
vannicktrinquier, line 32: Adding test for observability

fast/stages/0-org-setup/datasets/classic/folders/security/dev/.config.yaml
ludoo: I would explore the possibility of using conditionals in org policies tied to the environment tags. Environment folders are often removed, and this should keep working even without.

vannicktrinquier: Actually i am using them already, implemented as follow in `organization/org-policies/cloudkms.yaml`
Is it what u mean ?

```
gcp.restrictCmekCryptoKeyProjects:
  rules:
    - allow:
        values:
          - 'under:${folder_ids["security/dev"]}'
      condition:
        title: Restrict CMEK Projects for development projects
        expression: |
          resource.matchTag('${organization.id}/environment', 'development')
    - allow:
        values:
          - 'under:${folder_ids["security/prod"]}'
      condition:
        title: Restrict CMEK Projects for non-development projects
        expression: |
          !resource.matchTag('${organization.id}/environment', 'development')
```

Here depending of the environment, the policy will ensure that project in development are using cmek keys from security projects in dev hierarchy. (and same for prod)

ludoo: Perfect. Then TBH I would even remove the dev/prod folders and assign those tags at the project level. :) But we can do it later nw.

vannicktrinquier: Noted, if this ok for u, we can keep as above for this PR. 
And in future PR, i can clarify with u and update later if needed

fast/stages/2-security/factory-projects.tf
```diff
@@ -25,7 +25,6 @@ module "factory" {
   )
   context = local.ctx
   factories_config = {
-    folders  = var.factories_config.folders
```
ludoo, line 28: This will break existing installations, and prevent some legitimate usage. Let's keep this as is.

ludoo: (resolve once you have reverted this change)

modules/project-factory/projects.tf
```diff
@@ -55,10 +55,6 @@ locals {
   project_numbers = {
     for k, v in module.projects : k => v.number
   }
-  ctx_log_buckets = merge(local.ctx.log_buckets, local.log_buckets)
```
ludoo, line 58: This will prevent using a log bucket in the project for project-level sinks. I understand why you're doing it, but let's remember to revisit this in the future. Maybe the only option we really have for logging / monitoring is to add a third module invocation. Just resolve this comment for now, but let's discuss once this PR has been merged.

ludoo (approved):
One small change to the security stage should be reverted, and we need to discuss logging buckets in the project factory after this has been merged.

I would merge this once the small change is in.

fast/stages/0-org-setup/variables.tf
```diff
@@ -44,6 +44,7 @@ variable "factories_config" {
...
```
wiktorn: Why don't we set the default here, as for others? `datasets/classic/observability` is anyway what we ask in the README.md, and use of this data set is controlled elsewhere.

fast/stages/0-org-setup/README.md
```diff
@@ -297,6 +300,52 @@ This dataset is meant as a minimalistic starting point for organizations where a
...
+1. Create a folder `datasets/classic/observability`.
+2. Set the terraform variable `factories_config` to configure the `observability` factory path to use the new folder.
+
+```hcl
+factories_config = {
+  ...
+  observability = "datasets/classic/observability"
+  ...
+}
+```
```
wiktorn: I'd set this as default in `variables.tf`

modules/project-factory/projects.tf
vannicktrinquier: Sure, lets discuss in future release how to handle it.
Resolving it then

fast/stages/2-security/factory-projects.tf
vannicktrinquier: Oki i have added it back. 
Resolving the comment now.
I will merge later once all tests are passed.

fast/stages/3-gke-dev/main.tf
```diff
@@ -26,6 +26,23 @@ locals {
...
+      ]
+    ]
+  ])))
+  _cmek_keys_pubsub = toset(compact(flatten([
+    [for k, v in var.clusters : try(v.enable_features.upgrade_notifications.kms_key_name, null)],
+  ])))
+  service_encryption_key_ids = {
+    for k, v in {
+      "container.googleapis.com" = local._cmek_keys_container
+      "pubsub.googleapis.com"    = local._cmek_keys_pubsub
```
wiktorn, line 43: If I read the code above correctly, it is possible to have a `null` values in those sets. Shouldn't we remove them, before checking if the set length is longer than 0?

modules/folder/README.md
```diff
@@ -398,7 +398,7 @@ module "folder-sink" {
     no-gce-instances = "resource.type=gce_instance"
   }
 }
-# tftest modules=6 resources=18 inventory=logging.yaml e2e
+# tftest modules=6 resources=19 inventory=logging.yaml e2e
```
wiktorn:
```suggestion
# tftest inventory=logging.yaml e2e
```
Let's keep those numbers only in inventory (in other places as well)

fast/stages/3-gke-dev/main.tf
vannicktrinquier: Let me reply based on my understanding of your comment. If understood it wrong, let me know

The removal is actually already done by using the `toset(compact(flatten` above.
The function `compact` takes a list of strings and returns a new list with any null values or empty strings ("") removed. 
So when checking the length(v), we should be good ?
Let me know if make sense @wiktorn

modules/folder/README.md
vannicktrinquier: Yes, overall i am agree with you. Those #numbers are duplicated across inventory file and README.md.
I just followed the logic already implemented in past

But, if no concern for the team (cc @ludoo @juliocc ). This is ok for me to update the readme.md of this PR to remove them. 
Let me know

wiktorn: They were necessary earlier when we used "regex" parser, otherwise tests wouldn't work. Since @juliocc implemented a proper one, we can skip those.

fast/stages/3-gke-dev/main.tf
wiktorn: Sounds good to me!

fast/stages/0-org-setup/variables.tf
vannicktrinquier: As per now, only the hardened dataset were having some observability sets configured. See [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/tree/vannick/fsi-4-wave/fast/stages/0-org-setup/datasets/hardened/observability) So i did follow initially the same logic to focus on hardened instead.

But what u tell make sense, even if we dont have any observability files for classic dataset yet (empty folder or non existing folder), we can setup FAST to be ready to use them for any kind of datasets (for classic). This can be done by:
- updating the default value in variable.tf to `datasets/classic/observability`
- updating the `datasets/classic/default.yaml` to add the reference to the log project (add done for default.yaml in hardened dataset)

Anyway the factory observability will do nothing if the folder does not exists or no file are present. So it will be skip anyway.

On my side, sounds good to do the change
@ludoo @juliocc any comment ?

modules/folder/README.md
vannicktrinquier: Done, i have updated in the tests of this PR. 
Closing the conversation

fast/stages/0-org-setup/variables.tf
vannicktrinquier: Updated accordingly, let me know if any comments

fast/stages/0-org-setup/README.md
vannicktrinquier: Updated the readme to make sense with change done to set `datasets/classic/observability` as default

PR #3558, opened 2025-12-02
Add support for CMEK in logging bucket, big query dataset and gke notifications

vannicktrinquier:
Purpose of this PR is to prepare update of FAST hardened set to accommodate with CMEK encryption
It contains:
- Update of underlying bucket for dataset, logging bucket to handle context and kms_keys from context
- Add support to gke module to handle gke notification using pubsub topic with cmek

### DIFF
--- a/fast/stages/0-org-setup/README.md
+++ b/fast/stages/0-org-setup/README.md
@@ -7,6 +7,7 @@
   - [Configure defaults](#configure-defaults)
   - [Initial user permissions](#initial-user-permissions)
   - [First apply cycle](#first-apply-cycle)
+    - [Default project](#default-project)
     - [Importing org policies](#importing-org-policies)
     - [Local output files storage](#local-output-files-storage)
     - [Init and apply the stage](#init-and-apply-the-stage)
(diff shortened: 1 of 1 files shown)


modules/logging-bucket/main.tf
```diff
@@ -42,6 +42,12 @@ locals {
   }
 }
 
+# generates the service agent for logging.googleapis.com when doesn't exist already
+data "google_logging_project_cmek_settings" "settings" {
```
vannicktrinquier: Having issue to add a test for logging-bucket because of this data source 

When running the test the project id does not exist for test, which make the call to data source failing (and the test).

This data source is used to create the service agent (as logging is a special service that dont create the service agent with normal way ... need to call that API ...) 

 @ludoo any idea how the test can handle it ?

modules/gke-cluster-standard/variables.tf
```diff
@@ -243,9 +243,10 @@ variable "enable_features" {
     shielded_nodes       = optional(bool, false)
     tpu                  = optional(bool, false)
     upgrade_notifications = optional(object({
-      enabled     = optional(bool, true)
-      event_types = optional(list(string), [])
-      topic_id    = optional(string)
+      enabled      = optional(bool, true)
+      event_types  = optional(list(string), [])
+      topic_id     = optional(string)
+      kms_key_name = optional(string)
```
vannicktrinquier, line 249: @ludoo confirming before doing
Should we add a test case from the README.md [here](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/262c4670bad6cd500146993fbeb04e8bed9e5f03/modules/gke-cluster-standard/README.md#upgrade-notifications) or not needed @ludoo 


Same question for gke autopilot (which does not have anything for now in README.md) ?

ludoo: It would be good to change one of the existing examples and test this field, but if they are e2e we also need the key, so it's simpler to copy/paste one of the examples, add kms_key with just a fake key id, and don't use e2e for that.

modules/project-factory/projects-bigquery.tf
```diff
@@ -46,10 +47,15 @@ module "bigquery-datasets" {
...
+  encryption_key = each.value.encryption_key
+  friendly_name  = each.value.friendly_name
   location = coalesce(
     local.data_defaults.overrides.locations.bigquery,
     lookup(each.value, "location", null),
     local.data_defaults.defaults.locations.bigquery
   )
+
+  depends_on = [
+    module.projects-iam
```
ludoo: Instead of using an explicit dependency, we can just use project-iam for the project id (line 37 here). Same for buckets.

modules/gke-cluster-standard/variables.tf
vannicktrinquier: I actually did update example for gke-standard to add this field. Should be ok now>

And notification was not existing as example in gke-autopilot. I added it also. 

We should be good now

modules/project-factory/projects-bigquery.tf
vannicktrinquier: Done, should be good now

modules/logging-bucket/main.tf
vannicktrinquier: Outade to resolving the conversation

ludoo: Sorry, missed this: we'll have to skip tests for this then.

PR #3577, opened 2025-12-09
feat(compute-mig): add instance_lifecycle_policy support

### LINKED ISSUE #3576: compute-mig does not support setting instance_lifecycle_policy
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

The compute-mig module does not expose the instance_lifecycle_policy object.

### Release version

latest

### Terraform version

latest

### Steps to reproduce

n/a

### Expected behavior

create mig with lifecycle policy (e.g. repair in other zone)

### Actual behavior

Cannot create MIG with lifecycle policy

### Additional context

_No response_

frits-v:
Fixes #3576


modules/compute-mig/variables.tf
```diff
@@ -188,6 +188,20 @@ variable "instance_flexibility_policy_selections" {
   }
 }
 
+variable "instance_lifecycle_policy" {
```
ludoo, line 191: nit: I would call it `lifecycle_policy`

modules/compute-mig/variables.tf
```diff
@@ -188,6 +188,20 @@ variable "instance_flexibility_policy_selections" {
...
+  description = "The instance lifecycle policy for the MIG."
+  type = object({
+    force_update_on_repair    = optional(string)
```
ludoo: I would move `force_update_on_repair` to `on_repair.force_update`

frits-v: @ludoo I followed the upstream convention in all naming and structure, but happy to change it.

modules/compute-mig/variables.tf
frits-v: @ludoo it might be confusing considering terraform's lifecycle meta argument

modules/compute-mig/variables.tf
ludoo: We typically reorganize attributes / shorten names where it makes sense, don't just port the provider interface. It's not terribly important here, but still makes for a nicer user interface IMHO.

frits-v: agreed -- updated accordingly.

PR #3595, opened 2025-12-16
feat: project-factory folders support deletion_protection

joshmyers:
## What
project-factory buckets support `deletion_protection`

## Why
So we can set `deletion_protection = true` on buckets managed via project-factory.

Defaults to `false` for backwards compat.

---
**Checklist**

### DIFF
--- a/modules/alloydb/README.md
+++ b/modules/alloydb/README.md
@@ -63,14 +63,12 @@ module "alloydb" {
   cluster_name   = "db"
   instance_name  = "db"
   location       = var.region
-  initial_user = {
-    password = "changeit"
-  }
   network_config = {
     psa_config = {
       network = module.vpc.id
     }
   }
+
   deletion_protection = false
 }
 # tftest modules=3 resources=17 inventory=simple.yaml e2e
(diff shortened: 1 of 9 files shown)


modules/project-factory/folders.tf
```diff
@@ -99,6 +100,7 @@ module "folder-2" {
   for_each = {
     for k, v in local.folders_input : k => v if v.level == 2
   }
+  deletion_protection = lookup(each.value, "deletion_protection", false)
```
juliocc, line 103: Can you default to `null` instead of `false`

joshmyers: done

joshmyers:
@juliocc Before I address the failing tests, they are failing because the nulls are overriding the folder module default variable of [false](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/folder/variables.tf#L117), which means they are getting set to null and the provider default is `deletion_protection = true`, so this PR ends up being a change in behaviour, not a no-op. Is this desired? 

I think this makes sense given the provider default.

PR #3628, opened 2026-01-08
Use additive resource for perimeter resources in vpc-sc module when ignore changes is set

ludoo:
This should fix a corner case in the vpc-sc module, where a perimeter is defined with `ignore_resource_changes` but still defines resources internally. This change switches the module to use `google_access_context_manager_service_perimeter_resource` and `google_access_context_manager_service_perimeter_dry_run_resource` in those perimeters, to allow defining resources both at the perimeter level, and externally.

**Breaking Changes**

```upgrade-note
`modules/vpc-sc`: perimeters defined with `ignore_resource_changes` will recreate resource membership via additional resources.
```

### DIFF
--- a/modules/vpc-sc/perimeters-additive.tf
+++ b/modules/vpc-sc/perimeters-additive.tf
@@ -61,15 +61,7 @@ resource "google_access_context_manager_service_perimeter" "additive" {
           try(google_access_context_manager_access_level.basic[k].id, k)
         ]
       )
-      # use additive resources to avoid overriding externally defined ones
-      # resources = flatten([
-      #   for r in spec.value.resources : try(
-      #     local.ctx.resource_sets[r],
-      #     [local.ctx.project_numbers[r]],
-      #     [local.project_numbers[r]],
-      #     [r]
-      #   )
-      # ])
+      # use additive resources below to avoid overriding externally defined ones
       restricted_services = flatten([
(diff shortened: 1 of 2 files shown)


modules/vpc-sc/perimeters-additive.tf
```diff
@@ -36,14 +61,15 @@ resource "google_access_context_manager_service_perimeter" "additive" {
...
-      ])
+      # use additive resources to avoid overriding externally defined ones
+      # resources = flatten([
+      #   for r in spec.value.resources : try(
+      #     local.ctx.resource_sets[r],
+      #     [local.ctx.project_numbers[r]],
+      #     [local.project_numbers[r]],
+      #     [r]
+      #   )
+      # ])
```
juliocc: remove?

modules/vpc-sc/perimeters-additive.tf
```diff
@@ -234,13 +260,14 @@ resource "google_access_context_manager_service_perimeter" "additive" {
...
-        )
-      ])
+      # use additive resources to avoid overriding externally defined ones
+      # resources = flatten([
+      #   for r in status.value.resources : try(
+      #     local.ctx.resource_sets[r],
+      #     [local.ctx.project_numbers[r]],
+      #     [local.project_numbers[r]], [r]
+      #   )
+      # ])
```
juliocc: remove?

PR #3650, opened 2026-01-14
feat(logging-bucket): support locked parameter for project parent types

singhal0306:
- Add locked parameter to project bucket resources with default value of false.

---
**Checklist**

### DIFF
--- a/modules/artifact-registry/main.tf
+++ b/modules/artifact-registry/main.tf
@@ -71,7 +71,7 @@ resource "google_artifact_registry_repository" "registry" {
   dynamic "docker_config" {
     # TODO: open a bug on the provider for this permadiff
     for_each = (
-      local.format_string == "docker" && try(local.format_obj.standard.immutable_tags, null) == true
+      local.format_string == "docker" && try(local.format_obj.standard.immutable_tags, null) != null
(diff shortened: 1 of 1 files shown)


modules/logging-bucket/variables.tf
```diff
@@ -91,6 +91,13 @@ variable "tag_bindings" {
   nullable    = false
 }
 
+variable "locked" {
+  description = "Whether the bucket is locked. Locked buckets may only be deleted if they are empty"
+  type        = bool
+  default     = false
+  nullable    = false
```
juliocc: Do not provide a default and allow nulls. This will fallback to the provider default

modules/logging-bucket/variables.tf
```diff
@@ -91,6 +91,13 @@ variable "tag_bindings" {
   nullable    = false
 }
 
+variable "locked" {
```
juliocc: This variable only makes sense when working with project-level buckets. Please add a validation to ensure this remains unset for organization, folder and billing account buckets.

singhal0306 (commented):
Formatting is done and all the changes requested have been resolved.

PR #3654, opened 2026-01-18
improve configurability and resource references for internal ALB

singhal0306:
…nces

- Allow overriding names for backend buckets and instance groups
- Add optional per-group description with default value
- Use self_link for instance group backend references
- Fix HTTP proxy name to use http_proxy_config

---
**Checklist**


modules/net-lb-app-ext/backend-service.tf
```diff
@@ -19,7 +19,7 @@
 locals {
   group_ids = merge(
     {
-      for k, v in google_compute_instance_group.default : k => v.id
+      for k, v in google_compute_instance_group.default : k => v.self_link
```
juliocc: why switch here, and only here, to self link?

juliocc (commented):
Can you confirm if other load balancer modules (e.g. net-lb-app-int, net-lb-app-int-regional, etc) would need similar update?

modules/net-lb-app-ext/backend-service.tf
singhal0306: Yes I guess so, other modules also require this change, but for network endpoint groups i found discrepency as the one example usage: use **id** but another one use **self link** and description mention to use only fully qualified URLs.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_backend_service

image

ludoo: Other modules **do not** require this change unless absolutely necessary. We've intentionally switched from long-form self links to short-form IDs, and those are required in multiple places.

singhal0306: Yes you are correct, while importing my resources i was getting a change that it needs self_link but after importing and reverting back to id, i am not seeing any changes to my backend services.

singhal0306: Let me update the PR

singhal0306:
@juliocc please merge the PR.

singhal0306:
I can pick this up 👍
I’ll review the other modules to confirm whether the same changes are required, and update them accordingly if needed.

PR #2390, opened 2024-06-26
Add AssuredWorkload support to the folder module

averbuks:
Add AssuredWorkload support to the folder module
---
**Checklist**

I applicable, I acknowledge that I have:
- [X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X] Ran `terraform fmt` on all modified files
- [X] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [X] Made sure all relevant tests pass

PR #3666, opened 2026-01-22
Add support for the Assured Workloads in the project factory

lopezvit:
There was an original PR #2390 that added AssuredWorkload support at the folder module, but was not available to the project-factory module, nor to the stage 2-project-factory.

This PR fixes (enum values) and expands on the original PR with the following characteristics:

* Updated/Fixed `folder` module
* Expands `project-factory` module to use AssuredWorkloads
* Expands stage `2-project-factory` `folder.schema.json` to support Assured Workloads. This change also introduces the `organization_id` as a variable, sourced from `0-globals`, as it's required for Assured Workloads. Fixed some other variables documentation to clearly state where they were sourced, as it was wrong.
* Adds validated support for Assured Workloads to the `0-org-setup` stage. This configuration is optional and commented out by default, ready to be enabled for implementations that require it.
* Updated other stages that might need AssuredWorkloads in the future, but only the folder.schema.json
  * `2-networking`
  * `2-security`

---
**Checklist**

### DIFF
--- a/tests/fast/stages/s2_project_factory/simple.tfvars
+++ b/tests/fast/stages/s2_project_factory/simple.tfvars
@@ -17,3 +17,8 @@ data_defaults = {
     storage = "europe-west1"
   }
 }
+organization = {
+  domain      = "fast.example.com"
+  id          = 123456789012
+  customer_id = "C00000000"
+}


fast/stages/2-project-factory/variables-fast.tf
```diff
@@ -71,6 +71,16 @@ variable "kms_keys" {
...
 
+variable "organization" {
+  # tfdoc:variable:source 0-globals
+  description = "Organization details."
+  type = object({
+    domain      = string
+    id          = number
+    customer_id = string
+  })
+}
```
juliocc, line 82: why do you need this in the project factory?

lopezvit: Assured Workloads has a mandatory field called organization that expects the organization id. Since the value is there from the output of stage-0 is for free to have it there.
This is how it looks in my configuration:
```
# yaml-language-server: $schema=../../../../schemas/folder.schema.json

name: <REDACTED>
assured_workload_config: 
    compliance_regime        : "REGIONAL_CONTROLS"
    display_name             : "workload-name"
    location                 : "europe-north1"
    organization             : "${organization.id}"
    enable_sovereign_controls: true

```

PR #3682, opened 2026-01-27
Add custom bucket name for project-factory module

kovagoadam:
Add custom bucket name for project-factory module
---
Added missing logic behind bucket name attribute in project-factory module
**Checklist**


modules/project-factory/projects-buckets.tf
```diff
@@ -21,6 +21,7 @@ locals {
         project_key    = k
         project_name   = v.name
         name           = name
+        bucket_name    = lookup(opts, "name", "${v.name}-${name}")
```
ludoo: why not reuse `name`?

kovagoadam: I cannot reuse name because it holds the loop key (the suffix). I need that original value to construct the default format (${v.name}-${name}) in the lookup function. If I reused name for the result of the lookup, I would lose the reference required for the default value calculation. If I override name, the following happens:
```
╷
│ Error: Invalid index
│ 
│   on modules/project-factory/outputs.tf line 66, in locals:
│   66:           module.buckets["${k}/${sk}"].name
│     ├────────────────
│     │ module.buckets is object with 11 attributes
│ 
│ The given key does not identify an element in this collection value.
╵
```
Does this approach look reasonable to you, or would you prefer a different implementation?

ludoo: Possible alternative implementation which I would prefer as it makes it explicit what the bucket key is.

```hcl
locals {
  projects_buckets = flatten([
    for k, v in local.projects_input : [
      for name, opts in lookup(v, "buckets", {}) : {
        key            = "${k}/${name}"
        project_key    = k
        project_name   = v.name
        name           = lookup(opts, "name", "${v.name}-${name}")
```

```hcl
module "buckets" {
  source         = "../gcs"
  for_each       = { for k in local.projects_buckets : k.key => k }
  project_id     = module.projects-iam[each.value.project_key].project_id
  prefix         = each.value.prefix
  name           = each.value.name
```

kovagoadam: Thanks for the suggestion, I have updated to code. This should make the bucket key name more explicit.

PR #3696, opened 2026-02-03
Net-vpc-factory

sruffilli:
This PR (re :D)introduces module net-vpc-factory, a minimal factory that deals with vpcs, subnets, and firewall rules creation, meant to be embedded into other factories, starting with 2-data-platform and 2-networking.


fast/stages/2-networking/factory-cloudnat.tf
```diff
@@ -31,7 +31,7 @@ locals {
           logging_filter            = try(nat_config.logging_filter, null)
           router_asn                = try(nat_config.router_asn, null)
           router_create             = try(nat_config.router_create, true)
-          router_network            = module.vpcs[vpc_key].self_link
+          router_network            = module.vpc-factory.vpcs[vpc_key].self_link
```
juliocc: `.id`  🙏

PR #3697, opened 2026-02-03
Add basic observability example to classic dataset

juliocc:
---
**Checklist**

### DIFF
--- a/tools/duplicate-diff.py
+++ b/tools/duplicate-diff.py
@@ -62,7 +62,6 @@ duplicates = [
     ],
     [
         "fast/stages/0-org-setup/schemas/observability.schema.json",
-        "modules/project-factory/schemas/observability.schema.json",
         "modules/project/schemas/observability.schema.json",
     ],
     [


modules/project-factory/schemas/observability.schema.md
```diff
@@ -0,0 +1,166 @@
+# Observability Schema
```
ludoo: do we need the schema here, if the module only supports the factory path? probably yes, just wondering

juliocc: Yes, all resources all the relevant resources are created at the project level (notification channels, alerts, etc).

ludoo: But not specified in YAML which is processed by the PF. Also wondering this, as I'm adding support for aspect types and policy tags. My preference would be that whenever the pf only passes a path to the underlying module, it should not carry those schemas. The modules will need to be present anyway, and those provide the schema as they are the ones validating it.

Let's discuss this for a quick minute.

juliocc: I see your point now. The project module supports both (factory path and through variables) but the project factory only exposes. I'm removing this

PR #3699, opened 2026-02-04
Add support for bucket logging configuration in module gcs and project-factory

vannicktrinquier:
Add support for usage and storage logs in module `gcs` and `project factory`.
See https://docs.cloud.google.com/storage/docs/access-logs

### DIFF
--- a/modules/gcs/main.tf
+++ b/modules/gcs/main.tf
@@ -125,6 +125,7 @@ resource "google_storage_bucket" "bucket" {
   dynamic "logging" {
     for_each = var.logging_config == null ? [] : [""]
     content {
+      # log_bucket is actually a storage bucket (and not a logging bucket)
       log_bucket = var.logging_config.log_bucket == null ? null : lookup(
         local.ctx.storage_buckets,
         var.logging_config.log_bucket,


modules/gcs/main.tf
```diff
@@ -125,7 +125,11 @@ resource "google_storage_bucket" "bucket" {
   dynamic "logging" {
     for_each = var.logging_config == null ? [] : [""]
     content {
-      log_bucket        = var.logging_config.log_bucket
+      log_bucket = var.logging_config.log_bucket == null ? null : lookup(
+        local.ctx.storage_buckets,
```
ludoo, line 130: the context should be `logging_buckets`

juliocc: It's weird but usage logs are indeed sent to a GCS bucket

> Once set up, both usage logs and storage logs are automatically generated for the specified bucket and stored as new objects in a bucket that you specify.
>
> Usage logs and storage logs are subject to the same pricing as other objects stored in Cloud Storage.

ludoo: Then the attribute name should be storage bucket or bucket 😂

vannicktrinquier: Yes, naming is confusing but indeed a storage bucket and not a logging bucket. See from terraform docs:
image

Initially i did the same mistake when i started implementation by thinking it was logging bucket.

```
the context should be logging_buckets
Then the attribute name should be storage bucket or bucket 😂
```
Then, i deduct nothing to change in current PR as we are using `$storage_buckets` for context, and the attribute is using existing Terraform attribute name.

Please confirm ?  and if yes, i will resolve the conversation and merge after

PR #3704, opened 2026-02-04
fix(project-factory): Correctly interpolate IAM principals in tags

lopezvit:
When creating Tags in a project using the project-factory, the iam of said tags currently require the members to be set statically, because, when the tags are created, the service accounts are not yet created. 
This PR moves the creation of the Tags inside the module "projects-iam" of the project-factory instead, which does not break anything, but allows the use of dynamic variables.

---
**Checklist**


ludoo:
Ok, so the approach to follow here is the same we're following in th FAST stage 0 for org-level tags:

- the first module call creates tags, and [sets](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/fast/stages/0-org-setup/organization.tf#L99) `tags_config.ignore_iam` to `true` so that IAM bindings in the tags structure are ignored
- the second module receives the tag structure again, and [sets](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/fast/stages/0-org-setup/organization.tf#L168) `tags_config.force_context_ids` to `true` so that tag IDs are looked for in context (which has been populated with the previously generated tag key and value IDs)

This is the same approach we need to follow in the project factory, or tags won't be available in IAM conditions, which would break a lot of the code we already have, and prevent several use cases from working.

The plan is

- implement the same `tags_config` mechanism in the project module
- leverage it from the project factory mirroring what the FAST stage 0 is doing

If you are up for it I can help you review the implementation. If it feels too much, I can do it fairly quickly since the original implementation in FAST is mine, and I have a good understanding of how it works.

Let me know your preference, and thanks for raising this.

ludoo:
Ping here if there's anything you need me to clarify, that piece of code might not be the most straightforward :)

lopezvit:
With your explanation seemed like something fast (and it was) so I have done it already if you want to review it now @ludoo 😉

modules/project-factory/projects.tf
```diff
@@ -186,5 +189,12 @@ module "projects-iam" {
   )
   shared_vpc_host_config    = each.value.shared_vpc_host_config
   shared_vpc_service_config = each.value.shared_vpc_service_config
-  universe                  = each.value.universe
+  tag_bindings = merge(
```
ludoo: I would leave this in the first module call, as I don't think it makes sense to apply a tag to the same project where it's created. But if this is actually a use case you have, you need to remove it from the first module call, as now you have it twice. :)

lopezvit: Sure, I'll make a new commit, I just copy and pasted, good catch ;)
I'll simply check the tests that are currently broken before making the commit.

ludoo (approved):
Thanks a lot for this, and being patient. :)

PR #3712, opened 2026-02-06
Fix context replacement in project factory tag IAM

lopezvit:
I created a PR yesterday #3704 and I realised that it introduced a bug that I'm correcting in this new PR. It is not done yet, as I would also like to add a `move` block, but this bug is more important, so I'm pushing it first.

---
**Checklist**

### DIFF
--- a/modules/project-factory/README.md
+++ b/modules/project-factory/README.md
@@ -47,7 +47,6 @@ The code is meant to be executed by a high level service account with powerful p
 - [Variables](#variables)
 - [Outputs](#outputs)
 - [Tests](#tests)
-  - [Tags with $iam_principals interpolation](#tags-with-iam_principals-interpolation)
 <!-- END TOC -->
 
 ## Folder hierarchy


lopezvit:
Hi @ludoo,
I considered using a `moved` block to transition these resources, but as far as I can tell, Terraform requires static resource identifiers and doesn't support dynamic references or `for_each` keys within them. If there is a way to achieve this that I've missed, I'd love to learn how. So, something like this in `modules/project-factory/projects.tf` cannot work, right?:
```
moved {
  from = module.projects.google_tags_tag_value_iam_binding.default
  to   = module.projects-iam.google_tags_tag_value_iam_binding.default
}
```
Similarly, I encountered what seems to be a limitation in the testing framework regarding the `tag_value` field. Since this attribute is a computed ID generated by GCP, it remains 'known after apply' and is not available in the `terraform plan `output that the tests validate against. I'm mentioning this because the bug I introduced in the previous PR cannot be checked by tests only, and could only be detected when executing in real infrastructure. Please let me know if there is a way to verify computed attributes like this that I'm not aware of.

modules/project-factory/projects.tf
```diff
@@ -61,6 +61,12 @@ locals {
       for kk, vv in v.service_agents : "service_agents/${k}/${kk}" => vv.iam_email
     }
   ]...)
+  proj_tag_keys = {
```
ludoo, line 64: You could move these below and avoid the additional locals, if they're not used anywhere else.

ludoo (approved):
Thanks, this slipped in my review. :)

I would not worry too much about moved blocks, IAM is idempotent. Worst case you'll need to apply twice.

modules/project-factory/projects.tf
lopezvit: I guess it is better to leave it as similar as possible to the organizations, right? Maybe it is more clear like this.

ludoo:
Well I was too quick and did not think this approach through for projects: this approach will pollute the tag context namespace, as it's not using project-prefixed keys. So if a project redefines a tag defined at org level, the more important org level tag will not be available at the project level.

Let me close this and roll back the previous merge, the approach for projects needs to be different from the org, and use contexts of the form '$tag_values:project_key/tag_key/tag_value`.

Sorry for this, but the project factory is our most complex module with a lot of internal machinery to ensure the crisscrossing of dependencies and contexts is usable for all main use cases.

PR #3713, opened 2026-02-07
Revert #3704

ludoo:
This reverts the changes to how project-level resource manager tags are handled in the project factory done in #3704.

Those changes addressed the problem of not having the full IAM principals context available in project tags IAM, but introduced a worse issue, where all tag key/value spaces for projects were merged together in contexts. This not only introduced the issue of overlapping project keys/values, but also risked shadowing the organization-level tags.

I will prepare a new PR to address the original issue, with proper namespacing of project-level tag resources and tests to catch any overlaps.


lopezvit:
Hi @ludoo do you have any ETA about this? should I create an enhancement request so it doesn't get forgotten?

ludoo:
it's been merged more than 2 months ago :)

lopezvit:
ah! I thought that you simply revert it, but not fix it. Nice, I'll cherry pick this commit in my production repo, Thanks

lopezvit:
for those looking for it, it is in #3714

ludoo:
Ach, sorry. I am a victim of too many context switches. :) Thanks for doing the research!

PR #3715, opened 2026-02-09
Refactor subnets mgmt in net-vpc-factory

sruffilli:
Refactor subnets mgmt in net-vpc-factory

This commit removes the ability to define subnets inline within the VPC `.config.yaml` across `net-vpc-factory` and enforces file-based subnet definitions using the `subnets/` subdirectory pattern.

Key changes include:
- **Module Updates:** Removed standard and non-standard inline subnet arguments (`subnets`, `subnets_private_nat`, `subnets_proxy_only`, `subnets_psc`) from the factory module's internal `vpcs` instantiation.
- **Schema & Docs:** Renamed `vpc.schema.json` to `vpc-factory.schema.json`, stripping inline subnet definitions from the JSON schema, and updated the module `README.md` to reflect the new file-based only approach.
- **Stage 2 Networking Configs:** Refactored datasets to migrate inline `subnets_proxy_only` definitions into standalone files like `subnets/prod-proxy.yaml`. Added a new exported output `subnet_ips` for downstream usage.

### DIFF
--- a/tests/modules/net_vpc_factory/examples/example.yaml
+++ b/tests/modules/net_vpc_factory/examples/example.yaml
@@ -15,9 +15,9 @@
 values:
   module.net-vpc-factory.module.firewall["data-vpc-0"].google_compute_firewall.custom-rules["allow-iap"]:
     allow:
-    - ports:
-      - '22'
-      protocol: tcp
+      - ports:
+          - "22"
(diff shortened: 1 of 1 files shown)


ludoo (approved):
should be merged on fast-dev

Issue #3735, opened 2026-02-12
Compute VM module fails if associated tag is NGFW tag


KonradSchieban:
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

Deployments of instances with the compute-vm module fail, if tags via `tag_binding_immutable` are NGFW tags. The reason is that in the current logic, tags are also associated to the boot disk, which is not possible for NGFW tags.

This can be checked in the UI when trying to add a secure tag (with NGFW purpose) to a disk:

<img width="1460" height="560" alt="Image" src="https://github.com/user-attachments/assets/4576e228-5fa6-48b4-88d4-7e51709832db" />

Otherwise, non-NGVW tags can be added to disks:

<img width="1467" height="609" alt="Image" src="https://github.com/user-attachments/assets/4e54e49d-6953-4b7c-8c6b-9471ae18f194" />

### Release version

6b6e50cc3

### Terraform version

v1.14.3

### Steps to reproduce

Deploy instance from compute-vm module with `tag_binding_immutable` entries where tags are NGFW tags.

### Expected behavior

Instances are deployed, NGFW tags are added to instances, and non-NGFW tags are added to both instances and disks.

### Actual behavior

Error message when reproducing the error:

<img width="946" height="167" alt="Image" src="https://github.com/user-attachments/assets/e6a3d198-4682-4b14-8b5f-37dc83597f42" />

PR #3736, opened 2026-02-12
Add support for NGFW Secure Tags on Compute Engine VMs

KonradSchieban:
This PR fixes issue https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/issues/3735

The way the issue is addressed is via an additional tag bindings variable for NGFW tags.

---
**Checklist**


modules/compute-vm/variables.tf
```diff
@@ -540,6 +540,20 @@ variable "tag_bindings_immutable" {
   }
 }
 
+variable "tag_bindings_immutable_ngfw" {
```
ludoo, line 543: We already support network scoped resource tags in other modules, I would use the same variable name here.

ludoo: `network_tag_bindings_immutable` probably

KonradSchieban:
Turns out this was already addressed in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/1895 and works correctly with network-scoped RM tags if the variable `network_tag_bindings` is used instead.

PR #3805, opened 2026-03-23
Compute VM module refactor

### LINKED ISSUE #3796: Snapshot schedule fails in compute-vm module for attached disks
When passing in an already existing disk in the compute-vm module, the snapshot schedules are computed in a loop that injects dynamic values, if the disk attach id references a dynamically created resource.

This makes the module fail

```bash
Error: Invalid for_each argument
│ 
│   on ../cloud-foundation-fabric/modules/compute-vm/resource-policies.tf line 174, in resource "google_compute_region_disk_resource_policy_attachment" "attached":
│  174:   for_each = {
│  175:     for attachment in local.disk_regional_schedule_attachments :
│  176:     "${attachment.disk_key}-${attachment.snapshot_schedule}" => attachment
│  177:   }
│     ├────────────────
│     │ local.disk_regional_schedule_attachments will be known only after apply
│ 
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will
│ identify the instances of this resource.
```

ludoo:
This is now feature complete, details of these changes are in the ADR. The deletion protection field has been left as in the previous version, @juliocc if you feel strongly about it it's a trivial change. This fixes #3796

This PR also contains a small change to the tfdoc tool, to properly parse formatting in headings when generating the TOC. And a new tool to run `terraform fmt` on README examples.

**Breaking Changes**

```upgrade-note
`modules/compute-vm`: the module interface has been refactored, underlying resources will keep their state id, but callers will need to update and align module attributes.
```

### DIFF
--- a/modules/net-lb-app-int-cross-region/recipe-cross-reg-int-app-lb-vm-dns/README.md
+++ b/modules/net-lb-app-int-cross-region/recipe-cross-reg-int-app-lb-vm-dns/README.md
@@ -107,7 +107,7 @@ vpc_config = {
   instances_config = {
     # both attributes are optional
     machine_type = "e2-small"
-    zones = ["b", "c"]
+    zones        = ["b", "c"]
   }
 }
 # tftest modules=5 resources=15
@@ -131,8 +131,8 @@ vpc_config = {
       "projects/my-project/global/networks/test",
       "projects/my-other-project/global/networks/test"
     ]
-    domain          = "foo.example."
-    hostname        = "lb-test"
+    domain   = "foo.example."
+    hostname = "lb-test"
   }
 }
 # tftest modules=5 resources=15
--- a/tools/format_tftest.py
+++ b/tools/format_tftest.py
@@ -16,41 +16,26 @@
 '''Format Terraform code blocks with tftest directives in README files.'''
 
 import os
-import re
 import subprocess
 import sys
+from pathlib import Path
 
 import click
+import marko
+
+# Add fabric root to sys.path to import from tests
+FABRIC_ROOT = Path(__file__).resolve().parents[1]
+sys.path.append(str(FABRIC_ROOT))
+
+try:
+  from tests.examples.utils import get_tftest_directive
+except ImportError:
+  print('Error: Could not import tests.examples.utils', file=sys.stderr)
+  sys.exit(1)
 
 
-def format_hcl_block(match):
-  content = match.group(1)
-  if '# tftest' not in content:
-    return match.group(0)
-
-  try:
-    proc = subprocess.run(['terraform', 'fmt', '-'], input=content, text=True,
-                          capture_output=True, check=True)
-    # The output from terraform fmt usually includes a trailing newline.
-    # To keep the format stable, we reconstruct the block exactly.
-    return f"```hcl\n{proc.stdout}```"
-  except subprocess.CalledProcessError:
-    return match.group(0)
-
-
-@click.command()
-@click.argument('paths', type=click.Path(exists=True), nargs=-1)
-@click.option('--check', is_flag=True,
-              help='Check if files need formatting without changing them.')
-def main(paths, check):
-  """Format Terraform code blocks with tftest directives in README files.
-
-  PATHS can be specific README.md files or directories to search recursively.
-  If no paths are provided, searches the current directory recursively.
-  """
-  if not paths:
-    paths = ('.',)
-
+def find_readme_files(paths):
+  '''Find all README.md files in the given paths.'''
   files_to_check = []
   for path in paths:
     if os.path.isfile(path) and os.path.basename(path) == 'README.md':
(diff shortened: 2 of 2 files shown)


modules/compute-vm/README.md
sruffilli, line 187: Missing inventory? Albeit probably WAI as you're testing a very similar inventory in the example below.

modules/compute-vm/test.tfvars
juliocc, line 1: remove file

modules/compute-vm/variables.tf
```diff
@@ -14,119 +14,87 @@
...
-    size              = string
-    snapshot_schedule = optional(list(string))
-    source            = optional(string)
-    source_type       = optional(string)
-    options = optional(
-      object({
-        architecture           = optional(string)
-        auto_delete            = optional(bool, false) # applies only to vm templates
-        mode                   = optional(string, "READ_WRITE")
+    # auto_delete can only be specified dor READ_WRITE, force null otherwise
```
juliocc: s/dor/for/

modules/compute-vm/variables.tf
```diff
@@ -14,119 +14,87 @@
...
+    mode = optional(string, "READ_WRITE")
```
juliocc, line 24: no validation for this enum

modules/compute-vm/variables.tf
```diff
@@ -14,119 +14,87 @@
...
-    ]) == length(var.attached_disks)
-    error_message = "auto_delete can only be specified on READ_WRITE disks."
-  }
-  validation {
-    condition = alltrue([for d in var.attached_disks :
-      (d.options.architecture == null || contains(["ARM64", "X86_64"], d.options.architecture))
-    ])
-    error_message = "Architecture can be null, 'X86_64' or 'ARM64'."
-  }
+  default = {}
```
juliocc: and `nullable=false`?

modules/compute-vm/variables.tf
```diff
@@ -14,119 +14,87 @@
...
+    auto_delete = optional(bool, true) # applies only to vm templates
     device_name = optional(string)
-    # TODO: size can be null when source_type is attach
-    size              = string
-    snapshot_schedule = optional(list(string))
-    source            = optional(string)
-    source_type       = optional(string)
-    options = optional(
-      object({
-        architecture           = optional(string)
```
juliocc, line 49: this field is nowhere else now. WAI?

modules/compute-vm/variables.tf
```diff
@@ -139,9 +107,13 @@ variable "can_ip_forward" {
...
 
 variable "confidential_compute" {
-  description = "Enable Confidential Compute for these instances."
-  type        = bool
-  default     = false
+  description = "Confidential Compute configuration. Set to 'SEV' or 'SEV_SNP' to enable."
+  type        = string
+  default     = null
+  validation {
+    condition     = var.confidential_compute == null || contains(["SEV", "SEV_SNP"], coalesce(var.confidential_compute, "-"))
```
juliocc: nit: `||` now short-circuits so the coalesce on the right side is not needed anymore.

modules/compute-vm/variables.tf
```diff
@@ -301,6 +269,58 @@ variable "labels" {
...
+  description = "Instance lifecycle and operational configurations."
+  type = object({
+    allow_stopping_for_update  = optional(bool, true)
+    deletion_protection        = optional(bool, false)
+    key_revocation_action_type = optional(string, "NONE")
+    graceful_shutdown = optional(object({
+      enabled           = optional(bool, false)
+      max_duration_secs = optional(number)
+    }))
+  })
```
juliocc, line 292: we're already breaking the interface, I'd leave all these defaults up to the provider

modules/compute-vm/variables.tf
```diff
@@ -386,63 +431,18 @@ variable "options" {
       values = list(string)
       in     = optional(bool, true)
     })), {})
-    spot               = optional(bool, false)
```
juliocc, line 389: where is spot now, i don't see it?

modules/compute-vm/variables.tf
```diff
@@ -14,119 +14,87 @@
...
+      }), {})
     }), {})
-    use_independent_disk = optional(bool, false)
+    source = optional(object({
+      attach = optional(string)
+      disk   = optional(string)
+      image  = optional(string)
+      # instant_snapshot = optional(string)
+      snapshot = optional(string)
+    }), { image = "projects/debian-cloud/global/images/family/debian-11" })
```
juliocc: nit: debian-cloud/debian-13
(shorter and bump version)

modules/compute-vm/variables.tf
sruffilli: `var.scheduling_config.provisioning_model`

modules/compute-vm/variables.tf
ludoo: Yes, it's inferred from boot disk since only one architecture is valid for an instance, and it's redundant here.

modules/compute-vm/variables.tf
ludoo: I disagree here: a lot of effort went in this module to provide an interface that does not require extensive configuration for proptotyping use, which has always been one of our goals. The defaults here accomplish that, or provide defaults that are needed by the provider (key). I am marking this resolved, we can revisit in the future but it's a larger discussion.

modules/compute-vm/variables.tf
ludoo: yup, marking as resolved

PR #3818, opened 2026-03-27
feat(2-networking): add NS And DS record dynamically from sub-zones records

lopezvit:
Allows the creating of delegated public sub-zones for the different environments.
Before this, it was necessary to create the public sub-zones first and copy the DS and NS records manually and thus required to apply it twice.

---
**Checklist**


ludoo (commented):
This looks nice and useful to me, @sruffilli can you review since net is basically your baby? :)

sruffilli (approved):

PR #3859, opened 2026-04-11
Automated PR Review

ludoo:
Adds a GitHub Action workflow to automatically review PRs using Vertex AI when the `automated-review` label is applied. Enforces repository conventions based on GEMINI.md and CONTRIBUTING.md.

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -73,19 +73,11 @@ jobs:
       - id: run-review
         name: Run Gemini PR Review
         run: |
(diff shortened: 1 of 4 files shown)


ludoo:
Thanks for the review! I've removed the redundant `` logic as suggested. Regarding `google-cloud-aiplatform` in `tools/requirements.txt`: since this specific script is only intended to be run autonomously via GitHub Actions within a dedicated runner (and not typically executed manually by developers locally), we decided to keep the dependency installation isolated to the workflow file rather than expanding the project's core requirements.

PR #3874, opened 2026-04-16
Add DNS Armor support

lnesteroff:
Adds  support for Cloud DNS Threat Detection (DNS Armour) into the Project Factory schema.
---
**Checklist**


ludoo (commented):
Skimmed the changes, and this looks like a single resource module, which we never do. Let's first discuss before going ahead.

ludoo (changes requested):
Need a discussion, as this is adding a ton of wrapping boilerplate for a 10-lines resource.

modules/project-factory/variables-projects.tf
```diff
@@ -231,6 +231,13 @@ variable "projects" {
       friendly_name  = optional(string)
       location       = optional(string)
     })), {})
+    dns_threat_detector = optional(object({
+      enabled                  = bool
+      excluded_networks        = optional(list(string), [])
+      labels                   = optional(map(string), {})
+      location                 = optional(string)
+      threat_detector_provider = optional(string)
+    }))
```
LucaPrete, line 240: should we make this default to {} and set enabled as optional (by default to set to false)?

modules/project-factory/README.md
LucaPrete, line 1: Should we add a small example, so that people know how to use DNS armor here?
Trivial but always worth to be explicit.

LucaPrete:
Hi @lnesteroff thanks for your PR!
I just left a couple of comments there. Almost ready :)

modules/project-factory/variables-projects.tf
lnesteroff: I aligned it to the way we did shared_vpc_host_config in the same file. Do you think we should set shared_vpc_host_config.enabled to optional(bool, false) and default to {} as well?

ludoo (changes requested):
Please look at the automated review above, and either implement the changes or provide a rationale for why they are not relevant.

lnesteroff:
> Please look at the automated review above, and either implement the changes or provide a rationale for why they are not relevant.

Not sure why It's saying that location is required. It's optional: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/network_security_dns_threat_detector#location-1

juliocc (approved):

PR #3875, opened 2026-04-16
Add Agent Engine identity type

LucaPrete:
- Uses by default agent identities for agents and allows users to optionally use service accounts
- Fixes missing defaults in variables
- Replaces the `service_account` output with the generic `identity` output
- Minor improvements to conditional logic for readability / formatting

---
**Checklist**

### DIFF
--- a/modules/agent-engine/identity.tf
+++ b/modules/agent-engine/identity.tf
@@ -1,5 +1,5 @@
 /**
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,20 @@
  */
 
 locals {
+  _effective_identity = coalesce(
+    try(google_vertex_ai_reasoning_engine.managed[0].spec[0].effective_identity, null),
+    try(google_vertex_ai_reasoning_engine.unmanaged[0].spec[0].effective_identity, null)
+  )
+  effective_identity = (
+    local._effective_identity == null
+    ? "principal://${local._effective_identity}"
+    : null
+  )
+  identity = coalesce(
+    local.effective_identity,
+    local.service_account_email,
+    "service-{your_project_number}@gcp-sa-aiplatform-re.iam.gserviceaccount.com"
+  )
   service_account_email = (
     var.service_account_config.create
     ? try(google_service_account.service_account[0].email, null) # use managed SA, when creating
@@ -47,7 +61,7 @@ resource "google_service_account" "service_account" {
   )
 }
 
-resource "google_project_iam_member" "default" {
+resource "google_project_iam_member" "iam_member_sa" {
   for_each = (
     var.service_account_config.create
     && var.agent_engine_config.identity_type == "SERVICE_ACCOUNT"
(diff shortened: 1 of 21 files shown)


juliocc (approved):
LGTM. Adding a test as Gemini suggests would be nice

LucaPrete:
Thanks @juliocc that comment actually triggered lots of other things that were missing. Please, put request changes, so we're sure it doesn't get merged before I fix something else. I'm waiting for SWE input.

juliocc:
I’ve approved because the code meets my requirements. If it’s not ready for merge, please mark it as a Draft.

LucaPrete:
@juliocc I went through a larger change. It would be good to please have a second round of reviews whenever you have a chance. Thanks!

modules/agent-engine/identity.tf
```diff
@@ -0,0 +1,85 @@
...
+resource "google_project_iam_member" "iam_member_sa" {
+  for_each = (
+    var.service_account_config.create
+    && var.agent_engine_config.identity_type == "SERVICE_ACCOUNT"
+    ? toset(local.roles)
+    : toset([])
+  )
+  role    = each.key
+  project = local.project_id
+  member  = google_service_account.service_account[0].member
```
juliocc, line 73: why not use `local.effective_identity` here?

modules/agent-engine/identity.tf
```diff
@@ -0,0 +1,85 @@
...
+  )
+  effective_identity = (
+    local._effective_identity == null
+    ? "principal://${local._effective_identity}"
+    : null
+  )
+  identity = coalesce(
+    local.effective_identity,
+    local.service_account_email,
+    "service-{your_project_number}@gcp-sa-aiplatform-re.iam.gserviceaccount.com"
```
juliocc, line 30: `{your_project_number}` is ok?

modules/agent-engine/variables.tf
```diff
@@ -114,15 +123,6 @@ variable "deployment_config" {
...
   }
-  validation {
-    condition = (
-      var.deployment_config.source_files_config == null ? true : (
-        (var.deployment_config.source_files_config.python_spec != null ? 1 : 0) +
-        (var.deployment_config.source_files_config.image_spec != null ? 1 : 0)
-      ) <= 1
-    )
-    error_message = "Only one of 'python_spec' or 'image_spec' can be specified within 'source_files_config'."
-  }
```
juliocc, line 125: this was recently introduced by #3885 to allow byoc. Confirm this is not breaking that.

tests/modules/agent_engine/examples/deletion-protection.yaml
```diff
@@ -13,26 +13,14 @@
 # limitations under the License.
 
 values:
-  module.agent_engine.google_project_iam_member.default["roles/aiplatform.user"]:
+  module.agent_engine.google_project_iam_member.iam_member_identity["roles/aiplatform.user"]:
```
juliocc, line 16: `default` changed to `iam_member_identity`. Review if a moved block is warrented.

juliocc (changes requested):

modules/agent-engine/identity.tf
LucaPrete: Yes. Don't want to ask it to the user just for printing it in output, neither using a data block

tests/modules/agent_engine/examples/deletion-protection.yaml
LucaPrete: What do you mean?
This correctly reflects what I want

modules/agent-engine/variables.tf
LucaPrete: Actually #3885 removed the defaults I put there, which is not desirable.

This removes the constraint from the variable but I changed the code so that if the image block is specified, the other block is not added to the resource.

It's a bit more opaque but better imho than not having those defaults, which are used most of the times and were causing the first readme example to fail (at apply)

modules/agent-engine/identity.tf
LucaPrete: This block is only used when the user assigns a service account to the agent and it's mutually exclusive with the one below (used for agent identity only)

tests/modules/agent_engine/examples/deletion-protection.yaml
juliocc: `google_project_iam_member.default` was renamed to `google_project_iam_member.iam_member_identity`. That will cause a recreation of the service account and anything that depends on it.

If you're ok with that, then ignore this.

juliocc (approved):

PR #3881, opened 2026-04-17
Add support for auto_update_policy to modules/cloud-function-v2

### LINKED ISSUE #3847: [FR]: Support `auto_update_policy` in cloudfunctions module
### The feature request

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions2_function#automatic_update_policy-1

No way to use auto_update_policy - in the main.tf block.

### Proposed solution

add capability to consume auto_update_policy

### Additional context

_No response_

vsingh55:
Currently, the `cloud-function-v2` module does not support configuring the `automatic_update_policy` / `on_deploy_update_policy` for the Cloud Function's runtime.
   
This PR addresses issue #3847 by adding an `automatic_update_policy` boolean property to the `function_config` variable.
   
- Setting `automatic_update_policy = true` renders the `automatic_update_policy {}` block.
- Setting `automatic_update_policy = false` renders the `on_deploy_update_policy {}` block.
- Leaving it `null` (the default) preserves the existing behavior.

Closes #3847

---
**Checklist**

<!--
**Breaking Changes**

```upgrade-note
`fast/stages/0-boostrap`: example upgrade note 1.

### CODE REVIEW

**Review by juliocc — CHANGES_REQUESTED** [review]
Looking at the documentation, it seems that `on_deploy_update_policy` and `automatic_update_policy` are mutually exclusive:
>**automatic_update_policy** - (Optional) Security patches are applied automatically to the runtime without requiring the function to be redeployed. This should be specified as an empty block and cannot be set alongside on_deploy_update_policy.
>
>**on_deploy_update_policy** - (Optional) Security patches are only applied when a function is redeployed. This should be specified as an empty block and cannot be set alongside automatic_update_policy.

However, the current proposal uses a single value to toggle between 3 states:
- `automatic_update_policy = true`: enables automatic security updates
- `automatic_update_policy = false`: **enables** security updates only on deploy
- `automatic_update_policy = null`: disables security updates

While I can see the value of having a single value, the naming is confusing. In particular having `automatic_update_policy = false` enable automatic updates is really counter-intuitive and potentially dangerous.

I suggest splitting the new field in two separate boolean fields, similar to what the underlying provider does.

Could you also add a short section in the README (e.g. 'Update Strategies') explaining the new field(s)?

**Comment by vsingh55** [comment]
> Looking at the documentation, it seems that `on_deploy_update_policy` and `automatic_update_policy` are mutually exclusive:
> 
> > **automatic_update_policy** - (Optional) Security patches are applied automatically to the runtime without requiring the function to be redeployed. This should be specified as an empty block and cannot be set alongside on_deploy_update_policy.
> > **on_deploy_update_policy** - (Optional) Security patches are only applied when a function is redeployed. This should be specified as an empty block and cannot be set alongside automatic_update_policy.
> 
> However, the current proposal uses a single value to toggle between 3 states:
> 
> * `automatic_update_policy = true`: enables automatic security updates
> * `automatic_update_policy = false`: **enables** security updates only on deploy
> * `automatic_update_policy = null`: disables security updates
> 
> While I can see the value of having a single value, the naming is confusing. In particular having `automatic_update_policy = false` enable automatic updates is really counter-intuitive and potentially dangerous.
> 
> I suggest splitting the new field in two separate boolean fields, similar to what the underlying provider does.
> 
> Could you also add a short section in the README (e.g. 'Update Strategies') explaining the new field(s)?

Hi @juliocc,

Thank you for the thoughtful review and for catching that—you make a great point. I completely agree that toggling between the two states with a single boolean was counter-intuitive and could easily lead to unintended or unsafe behaviors.

I've updated the PR to align with your suggestions and the upstream provider:

Split the single toggle into two separate boolean fields: automatic_update_policy and on_deploy_update_policy.
Added a validation block to safely restrict setting both to true concurrently.
Added an "Update Strategies" section to the README.md to clearly document how these exclusive fields should be used.
Regenerated the docs and formatted the module.
The branch has been force-pushed with the amended commit. Please let me know if there's anything else that needs adjusting!

File: modules/cloud-function-v2/variables.tf [hunk 1]
```diff
@@ -157,6 +159,13 @@ variable "function_config" {
     runtime         = "python310"
     timeout_seconds = 180
   }
+  validation {
+    condition = !(
+      coalesce(var.function_config.automatic_update_policy, false) &&
+      coalesce(var.function_config.on_deploy_update_policy, false)
```
juliocc, line 165: function_config is nullable so this will fail if `function_config = null`.

In fact I think this variable should be
```
nullable = false
default = {}
```
@ludoo @LucaPrete wdyt? Should we change it?

modules/cloud-function-v2/README.md
```diff
@@ -430,16 +431,22 @@ module "cf_http" {
 }
 # tftest inventory=direct-vpc-egress.yaml
 ```
+
+## Update Strategies
+
+Cloud Functions v2 supports applying security updates to the underlying runtime automatically (`automatic_update_policy`) or only upon deployment (`on_deploy_update_policy`).
+
+These two policies are mutually exclusive and are exposed via the `function_config.automatic_update_policy` and `function_config.on_deploy_update_policy` boolean fields. Set at most one of these to `true`. If neither is provided, the provider defaults apply.
```
juliocc, line 439: and a code example please 🙏

juliocc (changes requested):

vsingh55:
Hi @juliocc,

Good catches on both fronts, thank you!

Changed function_config to nullable = false and set the explicit top-level mapping to default = {} so that the custom validation block gracefully evaluates.
Documented a configuration example block natively inside the ## Update Strategies section within README.md to demonstrate how users can flag these toggles.
The codebase has been reformatted and I pushed up the amended commit!

modules/cloud-function-v2/README.md
```diff
@@ -430,16 +431,41 @@ module "cf_http" {
...
+  bundle_config = {
+    path = "assets/sample-function/"
+  }
+  function_config = {
+    automatic_update_policy = true
+  }
+  depends_on = [
+    google_project_iam_member.bucket_default_compute_account_grant,
+  ]
+}
```
juliocc, line 457: please add the tftest flag here.

juliocc (changes requested):

juliocc (approved):

PR #3885, opened 2026-04-20
Add agent engine BYOC support

heyanand:
Includes updates to support Bring Your Own Container deployment for Agent Engine
---
**Checklist**

**Breaking Changes**

```upgrade-note
`terraform-google-provider`: version updated to 7.28 which includes BYOC support.
`terraform-google-provider-beta`: version updated to 7.28 which includes BYOC support
`modules/agent-engine`: updated variable name from `deployment_files` to `deployment_config` for introducing `container_config`.  Evaluate recreation of resources before using this version.
```


modules/agent-engine/variables.tf
```diff
@@ -66,6 +67,9 @@ variable "context" {
 variable "deployment_files" {
   description = "The to source files path and names."
   type = object({
+    container_config = optional(object({
+      image_uri = string
+    }), null)
```
juliocc: Remove null, it's the default for optional: `optional(x, null)` is the same as just `optional(x)`

modules/agent-engine/agent-managed.tf
```diff
@@ -139,11 +147,29 @@ resource "google_vertex_ai_reasoning_engine" "managed" {
...
+        dynamic "image_spec" {
+          for_each = (
+            try(var.deployment_files.source_config.image_spec, null) == null
+            ? {}
+            : { 1 = 1 }
+          )
+          content {
+            build_args = var.deployment_files.source_config.image_spec.build_args
+          }
         }
```
juliocc, line 207: - `inline_source` is not dynamic, is this intended?
- the `python_spec` block checks for `var.deployment_files.source_config.image_spec`. I think it should check for the python-specific fields.

I think the problem is that you're mixing `inline_source`, `image_spec` and `python_source` in a flat `source_config` object. Perhaps the `var.deployment_files.source_config` should reflect the provider structure instead of trying to merge everything.

modules/agent-engine/variables.tf
```diff
@@ -66,6 +67,9 @@ variable "context" {
 variable "deployment_files" {
   description = "The to source files path and names."
   type = object({
+    container_config = optional(object({
+      image_uri = string
```
juliocc, line 73: This is under deployment_files but it's an URI... Is this the right place to put this field?

juliocc (changes requested):
This needs a bit of work before we can merge. I'd prefer not to overload `deployment_files` the way this PR is proposing.

Please check with @LucaPrete, he's the main author of this module.

modules/agent-engine/variables.tf
LucaPrete: Logically yes. The three sections indeed achieve the same end result.
But probably it's a good time to choose a better name? It can be simply something that doesn't contain the word files.

A proposal may be `deployment_config`, or `source_config` and change the inner `source_config` to `source_files_config`

heyanand: thanks @LucaPrete the latest commit has these changes with the naming conventions

modules/agent-engine/agent-managed.tf
heyanand: `inline_source` had been made dynamic.
`python_spec` check has been replaced with the right variable. `deployment_files` has been replaced by `deployment_config` with `source_files_config` implementing the mutually exclusive configs

modules/agent-engine/agent-managed.tf
```diff
@@ -120,42 +120,90 @@ resource "google_vertex_ai_reasoning_engine" "managed" {
...
+        dynamic "python_spec" {
+          for_each = (
+            try(var.deployment_config.source_files_config.python_spec, null) != null
+            ? { 1 = 1 }
+            : {}
+          )
+          content {
+            entrypoint_module = try(var.deployment_config.source_files_config.python_spec.entrypoint_module, "agent")
+            entrypoint_object = try(var.deployment_config.source_files_config.python_spec.entrypoint_object, "agent")
+            requirements_file = try(var.deployment_config.source_files_config.python_spec.requirements_file, "requirements.txt")
```
juliocc: I think these `try()` are not needed. The `for_each` condition together with the variable type guarantee the fields are defined.

modules/agent-engine/variables.tf
```diff
@@ -65,40 +66,63 @@ variable "context" {
...
-    }), null)
-    source_config = optional(object({
-      entrypoint_module = optional(string, "agent")
-      entrypoint_object = optional(string, "agent")
-      requirements_path = optional(string, "requirements.txt")
-      source_path       = optional(string, "./src/source.tar.gz")
-    }), null)
+    }))
+    source_files_config = optional(object({
+      source_path = optional(string)
```
juliocc: Source path defaulted to `./src/source.tar.gz` before. Now it's null. Double check if this is OK.

juliocc (changes requested):

juliocc (approved):

juliocc:
All good. Thanks @heyanand

modules/agent-engine/variables.tf
heyanand: Fixed this

PR #3937, opened 2026-05-06
Move org policies lower in the project factory dependency chain to support extended context

LucaPrete:
This enables the following use case:

```yaml
org_policies:
  compute.restrictSharedVpcSubnetworks:
    rules:
    - allow:
        values:
        - projects/${projects["host"]}/regions/${user_config["region"]}/subnetworks/${user_config["subnet_name"]}
```


ludoo (approved):

ludoo:
Let's wait for @juliocc to also chime in, so we're all aligned.

PR #3938, opened 2026-05-06
Fix custom_roles not permeating when used in dns zone IAM

lopezvit:
In case that you want to use custom roles for the IAM Binding of the dns zones, those custom roles needs to be passed down to the DNS factory in the context. Currently, there is a bug preventing it that this PR solves. 

---
**Checklist**

- [x]  Made sure all relevant tests pass


ludoo (approved):
thanks!

Issue #3913, opened 2026-04-27
[FR]: Use of “google_compute_region_health_check” resource on net-lb-app-int module


marco-gcp:
### The feature request

This is a small improvement.

The net-lb-app-int module primarily uses regional resources. Consider using regional health checks as well, as has already been done for the other “net-lb-XYZ” regional modules.

### Proposed solution

Use **google_compute_region_health_check** resource instead of google_compute_health_check resource.

PR #3939, opened 2026-05-06
Add support for regional health checks to net-lb-app-int module

ludoo:
This PR addresses #3913 by adding support for regional health checks in the `net-lb-app-int` module.

### Description

To avoid breaking changes for existing users, the module now supports **both** global and regional health checks:
- Added an optional `is_regional` boolean flag to `health_check_configs` (defaulting to `false`).
- When `is_regional` is set to `true`, the module creates a `google_compute_region_health_check` resource instead of `google_compute_health_check`.
- Updated backend services (`backend-service.tf`) and outputs (`outputs.tf`) to merge and support both global and regional health check IDs.
- Added a new test case (`health-checks-regional`) to verify regional health check functionality.
- Regenerated documentation.

### Verification Results

All 16 tests in `tests/modules/net_lb_app_int` passed successfully:
- 15 existing tests passed (no regression on default global health checks).
- 1 new test passed (verifying regional health checks).


juliocc (approved):
I agree with gemini :)

PR #3951, opened 2026-05-09
Add support for SCIM to workforce identity

juliocc:
---
**Checklist**

### DIFF
--- a/modules/organization/README.md
+++ b/modules/organization/README.md
@@ -987,8 +987,8 @@ module "org" {
             }
           }
           scim_tenant = {
-            scim_tenant_id = "my-scim-tenant"
-            display_name   = "My SCIM Tenant"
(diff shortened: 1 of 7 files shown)


ludoo:
We should also expose this via schema IMHO

juliocc:
> We should also expose this via schema IMHO

It's coming :)

ludoo (approved):

modules/organization/README.md
```diff
@@ -985,20 +986,20 @@ module "org" {
               }
             }
           }
-          oauth2_client_config = {
```
ludoo, line 988: I would keep oauth config as it's for illustrative purposes and the more general case (but only looking at the diff so it might still be elsewhere in this file)

juliocc: Good catch. Added it back

PR #3960, opened 2026-05-13
Declare contents: read on linting and tests workflows

arpitjain099:
Two CI workflows ran without a `permissions:` block. Linting is yaml/style/boilerplate checks, tests is terraform/tofu plan + unit tests with a provider matrix. Both are pure read — no commit/comment/release writes.

`contents: read` covers the checkout step. YAML validates.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -23,6 +23,10 @@ on:
     branches:
       - master
       - fast-dev
+
+permissions:
+  contents: read
+
 jobs:
   linting:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


ludoo:
Can you explain what this is addressing?

arpitjain099:
@ludoo No existing failure, this is preventive least-privilege hardening. Both `linting.yml` and `tests.yml` do checkout + lint/test only, so `contents: read` is the actual minimum scope they need (currently they inherit the repo default token scope which can be broader).

The benefit is defense-in-depth: if a third-party action used by either workflow is later compromised (recent example: the tj-actions/changed-files supply-chain attack, CVE-2025-30066), the over-scoped `GITHUB_TOKEN` gets exfiltrated. Pinning to `contents: read` caps the blast radius.

Happy to close if not the kind of change this repo accepts.

ludoo:
Thanks for the additional comment, it makes absolute sense.

ludoo (approved):

PR #3963, opened 2026-05-14
Add yaml reference doc for project factory stage

hsharma250:
This PR adds a documentation md file to 2-project-factory stage, as currently our README.md file doesn't cover all the properties we can set in YAML file and a description about it.
This is a ask by one of Google's customer and we thought that it is worth adding in public repository, please let us know if you can suggest a better solution.

---

### Changes

*   **Documentation:** Create YAML reference md file and Updated stage 2 README to mention the new file.


juliocc:
We already have autogenerated markdown files for each schema.json file. 

I see very little benefit having an additional tabular format.

PR #3966, opened 2026-05-14
Support factories_config in project factory module data defaults and overrides.

joshmyers:
## What
Support factories_config in data_defaults

## Why
There is currently no way to pass a default `factories_config.custom_roles` (or any other factories_config attribute) through data_defaults or data_overrides, those variables don't expose factories_config at all, so every project YAML has to repeat the path explicitly.

Not sure if this is intentional vs a miss, maybe the assumption was that factories_config paths would always be per project but for us it's a pretty natural thing to want e.g. a shared custom_roles folder that applies across all projects in a stack without having to wire it in every YAML file. For $reasons we can't use org level custom roles so revert to using `google_project_iam_custom_role` resources. 

This PR exposes factories_config in both data_defaults and data_overrides with the same optional fields already used in var.factories_config, and apply the standard override/default coalesce pattern in projects-defaults.tf that the rest of the fields follow.

We also wire `custom_role_id` into the `module.projects-iam` context so that custom roles created in the same root module are resolvable by IAM bindings in the same apply (otherwise they're unknown at plan time and the $custom_roles: interpolation falls back to the literal string).

Should be a non braking change. Have tested with setting `factories_config.custom_roles` as data_defaults/overrides and without. Get the expected changes.

---
**Checklist**


joshmyers:
cc @ludoo / @juliocc 🙏

ludoo:
So the approach was half-intentional: in our minds, rather than creating the same custom role across all projects it should be defined in the org. So while the changes are ok, I am curious as to why you are following a per-project approach as I'm pretty sure there's a good reason for it.

joshmyers:
@ludoo We are (unfortunately) in a shared org and don't have permissions to manage some org level resources, so can't use `google_organization_iam_custom_role` which indeed would be ideal.

We have split our project-factory into levels (so not a single state managing 1K+ resources), a single project-factory state only deals with a subset of the projects/folder hierarchy and defining on this level would help in this situation or others where you may want to provide custom_roles as data_defaults for a subset of projects.

ludoo:
Yes, I suspected that was the case, so you actually need this. Good point, and we shouldn't assume environments are always optimal. :) Thanks for this, and your patience. I'm giving it a quick read and approving.

ludoo (approved):
Change is good, so we can merge so as to unblock you. We might then also want to trickle up these configs in the schemas, and the defaults of the FAST stages. This can be a separate PR.

joshmyers:
> Change is good, so we can merge so as to unblock you. We might then also want to trickle up these configs in the schemas, and the defaults of the FAST stages. This can be a separate PR.

I did wonder about this / where I should draw the line. Felt like a decent starting point for potentially more work without increasing the scope of this change. Thank you 👍

PR #3970, opened 2026-05-15
Add IAM deny policies support

kovagoadam:
Add support IAM deny policies to `project`, `folder`, `organization` modules.
---
This PR introduces native support for IAM Deny policies to the organization, folder, and project modules. IAM Deny is a critical security primitive for a cloud environment, enabling teams to set centralized, preventative guardrails that strictly override any 'Allow' permissions. Integrating this feature directly into the hierarchy modules allows users to manage their Deny rules alongside standard IAM and Organization Policies in a consistent, DRY manner, all while maintaining backward compatibility, for those that do not need operational security features.

**Checklist**

### DIFF
--- a/modules/folder/variables-iam.tf
+++ b/modules/folder/variables-iam.tf
@@ -113,5 +113,6 @@ variable "iam_deny_policies" {
       exception_permissions = optional(list(string), [])
     }))
   }))
-  default = null
+  default  = {}
+  nullable = false
 }
--- a/modules/folder/variables-iam.tf
+++ b/modules/folder/variables-iam.tf
@@ -115,4 +115,18 @@ variable "iam_deny_policies" {
   }))
   default  = {}
   nullable = false
+  validation {
+    # Ensure denied_principals and denied_permissions are explicitly not null
+    # (to prevent HCL evaluation errors in loops) and contain at least one
+    # element (required by the GCP API).
+    condition = alltrue(flatten([
+      for k, v in var.iam_deny_policies : [
+        for r in v.rules : (
+          try(length(r.denied_principals) > 0, false) &&
+          try(length(r.denied_permissions) > 0, false)
+        )
+      ]
+    ]))
+    error_message = "Each rule in iam_deny_policies must have at least one denied principal and one denied permission."
+  }
 }
--- a/modules/folder/variables-iam.tf
+++ b/modules/folder/variables-iam.tf
@@ -94,3 +94,24 @@ variable "iam_by_principals_conditional" {
     error_message = "IAM bindings with the same condition title must have identical expressions and descriptions."
   }
 }
+
+variable "iam_deny_policies" {
+  description = "IAM Deny policies to be applied to the folder."
+  type = map(object({
+    display_name = optional(string)
+    rules = list(object({
+      description        = optional(string)
+      denied_principals  = list(string)
+      denied_permissions = list(string)
+      denial_condition = optional(object({
+        expression  = string
+        title       = optional(string)
+        description = optional(string)
+        location    = optional(string)
+      }))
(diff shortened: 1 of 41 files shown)


modules/organization/deny-policies.tf
```diff
@@ -0,0 +1,28 @@
...
+          content {
+            title       = try(rule.value.denial_condition.title, null)
+            expression  = rule.value.denial_condition.expression
+            description = try(rule.value.denial_condition.description, null)
+            location    = try(rule.value.denial_condition.location, null)
+          }
+        }
+        denied_permissions    = rule.value.denied_permissions
+        exception_principals  = try(rule.value.exception_principals, [])
+        exception_permissions = try(rule.value.exception_permissions, [])
```
juliocc: most of these try() are not needed. The for_each conditions ensure the inner objects are non-null

modules/folder/variables-iam.tf
```diff
@@ -94,3 +94,24 @@ variable "iam_by_principals_conditional" {
...
+        expression  = string
+        title       = optional(string)
+        description = optional(string)
+        location    = optional(string)
+      }))
+      exception_principals  = optional(list(string), [])
+      exception_permissions = optional(list(string), [])
+    }))
+  }))
+  default = null
```
juliocc: i'd make use `default = {}` + `nullable = false`. Simplifies a bit the outer for_each as the value is alway a map.

modules/organization/deny-policies.tf
```diff
@@ -0,0 +1,28 @@
...
+      }
```
juliocc, line 49: the `*_principals` and `expression` fields should be passed through context

juliocc (changes requested):
This looks really nice. 

We've avoided IAM deny for a while but this is exactly how I'd implement it. Thanks for taking the time to do this.

I've left a couple of comments but overall LGTM.

ysolt:
Amazing work @kovagoadam !! 

I just wonder if `roles/iam.denyAdmin`, `roles/iam.denyReviewer` IAM configuration is also needed on all iac service accounts and `$iam_principals:gcp-organization-admins`  as well to make Deny Policies fully functioning. Thanks!

@norbert-loderer FYI

modules/organization/deny-policies.tf
kovagoadam: I have tried this initially, however deny policy only supports the IAM V2 API, therefore only supporting the new principal format, which is currently not supported by Fabric, as the context contains every principal, whereas this needs it separated.
Reference docs: https://docs.cloud.google.com/iam/docs/principal-identifiers#deny
Let me know how this should be implemented.

kovagoadam:
Thanks for your time for taking a look at my PR. I have implemented some of the changes you requested. Let me know if you want any other changes.

modules/organization/deny-policies.tf
juliocc: That's a very good point. I don't think we enforce any formating to the values of $iam_principals. In fact we use principalSets in some places.

Anyway, let's leave it as is for now. We'll add it later if the need comes up.

juliocc (approved):

juliocc:
@kovagoadam I fixed a couple of linting errors and also added iam_principals context expansion. Hope you don't mind.

Thanks for this PR!

kovagoadam:
Hi @juliocc! Thanks for fixing the linting errors and merging the PR! Just a quick heads-up regarding the iam_principals context expansion: as I mentioned earlier in the thread, that context actually doesn't work as expected right now. Since this is already merged, would you like me to open a quick follow-up PR to address/remove it?

juliocc:
sure, go ahead. What's the problem?

PR #3978, opened 2026-05-21
Add support for arbitrary addresses in FAST networking stage

sruffilli:
Add support for arbitrary addresses in FAST networking stage

### DIFF
--- a/fast/stages/2-networking/schemas/addresses.schema.json
+++ b/fast/stages/2-networking/schemas/addresses.schema.json
@@ -4,7 +4,7 @@
   "type": "object",
   "additionalProperties": false,
   "properties": {
-    "external_addresses": {
+    "external": {
       "type": "object",
       "additionalProperties": {
         "type": "object",
@@ -28,7 +28,7 @@
         }
       }
     },
-    "global_addresses": {
+    "global": {
       "type": "object",
       "additionalProperties": {
         "type": "object",
@@ -40,7 +40,7 @@
         }
       }
     },
-    "internal_addresses": {
+    "internal": {
       "type": "object",
       "additionalProperties": {
         "type": "object",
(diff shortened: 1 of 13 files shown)


fast/stages/2-networking/schemas/addresses.schema.json
```diff
@@ -0,0 +1,116 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Addresses",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "external_addresses": {
```
ludoo: nit: `external`/`global` etc? this is the addresses schema so I tend to avoid logical repetitions, YMMV so ignore me if you feel differently

ludoo (approved):

sruffilli:
Replying to a machine sounds weird but hey that's 2026 for you.

> ### 1. Handle empty YAML files gracefully
> In `fast/stages/2-networking/factory-addresses.tf`, the `for_each` condition uses `length(v.addresses) > 0`. If a user creates an empty `addresses.yaml` file (or one containing only comments), `yamldecode` will return `null`. The `try(..., {})` in `factory-vpcs.tf` only catches errors, so `v.addresses` will evaluate to `null`, causing `length(null)` to throw an error during the plan.

-1, I'm happy with the plan failing here

> ### 2. Correct `global_addresses.ipv6` schema type
> In `fast/stages/2-networking/schemas/addresses.schema.json`, the `ipv6` property under `global_addresses` is defined as an object. However, in the `net-address` module, `global_addresses.ipv6` is a boolean (`optional(bool, false)`).
> 

-1, global_addresses.ipv6 in the net-address module  is defined as  optional(map(string))

fast/stages/2-networking/schemas/addresses.schema.json
sruffilli: makes sense

Issue #3976, opened 2026-05-20
[FR]: Support a concept of (IAM) "role sets" to simplify IAM bindings


lyricnz:
### The feature request

At the moment, every IAM binding needs to name every single IAM role which is required.  For some applications, this can be dozens of roles, each of which need to be replicated in several places (eg: when a logical-role-set is required at a couple of different Folders/Projects).  It would be much more convenient (and DRY) if you could define a "list of roles" and give it a name, then use this list of roles anywhere where you are creating IAM bindings.

(Yes, the native IAM bindings only support one role -> members)

### Proposed solution

- Add one (or more?) YAML files which define IAM role sets (role-set-name to list-of-roles)
- Whenever IAM bindings are being applied, if the IAM role is actually a role-set, expand the list of roles before passing to low-level resources

We actually have an implementation of this, entirely in 0-org-setup before passing fully-expanded lists to factories etc (it's a little awkward sometimes), but I can imagine a solution that uses ctx etc in the modules, rather than preprocessing everything in FAST.

FWIW, here's a sample of our code:

```terraform
# Expand $iam_role_sets: references and add computed attributes
  # The factory module expects these computed attributes for organizing folders:
  # - key: folder path (e.g., "workloads/int")
  # - level: hierarchy depth (1-3) - determines which folder module to use
  # - parent_key: parent folder path - used to auto-link folder hierarchy
  # - parent: explicit parent override from YAML (optional)
  folders_with_expanded_roles = {
    for key, folder_data in local._folders_raw : key => merge(
      folder_data,
      # Add computed attributes (same as factory calculates in folders.tf:43-50)
      {
        key        = key
        level      = length(split("/", key))
        parent_key = dirname(key)
        parent     = lookup(folder_data, "parent", null)
      },
      # Expand role sets in iam_by_principals
      # Converts: $iam_role_sets:engineer -> [roles/compute.admin, roles/storage.admin, ...]
      lookup(folder_data, "iam_by_principals", null) != null ? {
        iam_by_principals = {
          for principal, role_list in folder_data.iam_by_principals :
          principal => flatten([
            for role_or_set in role_list :
            startswith(role_or_set, "$iam_role_sets:") ?
            local.iam_role_sets[trimprefix(role_or_set, "$iam_role_sets:")] :
            [role_or_set]
          ])
        }
      } : {},

...
```

### Additional context

Example role-sets (simplified):

```yaml
# Viewer
[...]

PR #3980, opened 2026-05-21
Support IAM Role Sets in Context

ludoo:
Draft ADR for supporting IAM role sets in the context variable.

Refs #3976


lyricnz:
Thanks for writing this up so clearly. The overall design looks good to me - scoping expansion to `iam_by_principals*`, using the existing `context` variable, and adding factory support is the right approach. I agree with the general direction.

One thing I'd like to push back on gently is the decision to exclude the `organization` module.

The reasoning given is that org-level IAM should be carefully curated and that the savings don't justify the indirection. I understand that, but I'd argue role sets actually support careful curation rather than undermining it. Naming a set `security_auditor = ["roles/securitycenter.viewer", "roles/logging.viewer", "roles/iam.securityReviewer"]` is exactly the kind of deliberate, documented policy decision you want at the org level. The name becomes part of your policy vocabulary.

In practice, org-level IAM hits the same repetition problem described in the ADR. In a FAST stage-0 setup, several platform team service accounts might all need the same combination of org-level roles. Without role sets you repeat that list for each principal, and if the definition ever changes you have to update it in multiple places.

There's also a consistency argument. If a user learns that role sets work in `folder` and `project`, excluding `organization` is a gap they have to discover the hard way. The mental model should be uniform across the resource hierarchy.

The implementation cost is low - the logic is identical to what's already in folder and project. I don't think the added indirection meaningfully hurts auditability; the resolved roles are visible in the plan and the definition is a single named source of truth.

This is your project and your call. I just think the pragmatic benefit is real, the cost is small, and the consistency gain matters to users working at the org level.

lyricnz:
There's doesn't seem to be a lot of interest in this ADR...

ludoo:
@lyricnz so we looked at the implementation here (which is the correct one, minus the org module which could be added), and we still don't like adding this level of indirection. :) I am closing this, we can revisit in the future if the need surfaces again and we figure out it's a common use case, which at the moment it isn't.

Thanks for spending time on proposing and trying to push this forward, we gave it our best shot but still don't like it.

PR #3982, opened 2026-05-22
Enable remaining PSC config options in modules/cloudsql-instance

btuckVZ:
The `google_sql_database_instance` resource exposes a few PSC config options that aren't currently included in the module. Enable those config options via the `network_config` variable.

One important note:
The previous implementation exposed some of the PSC config via `network_config.connectivity.psc_allowed_consumer_projects`. This has been moved to `network_config.connectivity.psc_config.allowed_consumer_projects` to match the Terraform resource. Supplying the older parameter will cause a validation error.

---
**Checklist**

I applicable, I acknowledge that I have:
- [X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X] Ran `terraform fmt` on all modified files
- [X] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [X] Made sure all relevant tests pass

**Breaking Changes**

```upgrade-note
`modules/cloudsql-instance`: network_config.connectivity.psc_allowed_consumer_projects obsoleted.
```

PR #4000, opened 2026-05-28
Enable remaining PSC config options in modules/alloydb

btuckVZ:
The `google_alloydb_cluster` and `google_alloydb_instance` resources expose a few PSC config options that aren't currently included in the module. Enable those config options via the `network_config` variable.

Code decisions match feedback received in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/3982.

---
**Checklist**


modules/alloydb/variables.tf
```diff
@@ -277,6 +277,13 @@ variable "network_config" {
     }))
     psc_config = optional(object({
       allowed_consumer_projects = list(string)
+      psc_interface_configs = optional(object({
+        network_attachment_resource = optional(string)
+      }))
```
juliocc: what do you think about making this simply `network_attachment_uris = optional(list(string))`? 

This is closer to the type of the variable in the Cloud SQL PR you opened

btuckVZ: On https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/alloydb_instance, it seems only 1 interface can be specified:
```
[psc_interface_configs]
(https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/alloydb_instance#psc_interface_configs-1) - (Optional) Configurations for setting up PSC interfaces attached to the instance which are used for outbound connectivity. Currently, AlloyDB supports only 0 or 1 PSC interface.
```

The AlloyDB and Cloud SQL TF resources have some differences in them that don't quite make these like-for-like. Open to ideas though.

juliocc: even better then, just a single value should be enough?

btuckVZ: Yep, happy to distill down to one value, then can revisit if it ever supports multiple values.

juliocc (approved):

btuckVZ:
@juliocc I could use an extra set of eyes on this one. I'm not sure why `deletion_policy` would be getting checked on the instance...that attribute is on the cluster. I'm not seeing it in the other tests either.

juliocc:
> @juliocc I could use an extra set of eyes on this one. I'm not sure why `deletion_policy` would be getting checked on the instance...that attribute is on the cluster. I'm not seeing it in the other tests either.

You probably have a newer version locally of the terraform provider. I updated the inventory for you. Hopefully it will go through this time.

btuckVZ:
> > @juliocc I could use an extra set of eyes on this one. I'm not sure why `deletion_policy` would be getting checked on the instance...that attribute is on the cluster. I'm not seeing it in the other tests either.
> 
> You probably have a newer version locally of the terraform provider. I updated the inventory for you. Hopefully it will go through this time.

Got it, that's a helpful piece of information moving forward on how this whole process works. I appreciate the assistance. :)

Issue #3900, opened 2026-04-23
Module project is trying to add permissions on non existing service-accounts


kkluska-rtbh:
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

After upgrading the module version to `v55.1.0` terraform is trying to provision additional service accounts 

```
Terraform will perform the following actions:
  # module.project["XYZ"].google_project_iam_member.service_agents["aiplatform-pie"] will be created
  + resource "google_project_iam_member" "service_agents" {
      + etag    = (known after apply)
      + id      = (known after apply)
      + member  = "serviceAccount:service-11111111111@gcp-sa-aiplatform-pie.iam.gserviceaccount.com"
      + project = "XYZ"
      + role    = "roles/aiplatform.serviceAgent"
    }
  # module.project["XYZ"].google_project_iam_member.service_agents["vertex-sandbox"] will be created
  + resource "google_project_iam_member" "service_agents" {
      + etag    = (known after apply)
      + id      = (known after apply)
      + member  = "serviceAccount:service-11111111111@gcp-sa-vertex-sandbox.iam.gserviceaccount.com"
      + project = "XYZ"
      + role    = "roles/aiplatform.agentSandboxServiceAgent"
    }
Plan: 2 to add, 0 to change, 0 to destroy.

```

But those default service accounts where not created by GCP in the project. It looks like the change has been introduced in https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/pull/3877 .

### Release version

25bd428

### Terraform version

Terraform v1.14.5

### Steps to reproduce

Upgrade module version from `v55.0.0` to `v55.1.0`. 

from:
```
module "project" {
  source   = "github.com/GoogleCloudPlatform/cloud-foundation-fabric//modules/project?depth=1&ref=v55.0.0"
...
```

to:
```
module "project" {
  source   = "github.com/GoogleCloudPlatform/cloud-foundation-fabric//modules/project?depth=1&ref=v55.1.0"
...
```

### Expected behavior

It should not try to give permissions for those service accounts as they are non existing. 

### Actual behavior

Module is trying to add permissions on non existing service accounts.

### Additional context

Screenshot from GCP IAM console: 

<img width="1410" height="330" alt="Image" src="https://github.com/user-attachments/assets/aff2b5f5-f57d-4984-b1c8-50e4dab82562" />

juliocc:
[...]

Issue #3788, opened 2026-03-12
Project creation fails when enabling aiplatform.googleapis.com API


rsnowden-canva:
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Blueprints
- [ ] Other

### The problem

When creating a new GCP project with `aiplatform.googleapis.com` enabled, terraform apply fails with:
```
Error 400: Service account service-<PROJECT_NUMBER>@gcp-sa-vertex-ex-cc.iam.gserviceaccount.com does not exist., badRequest
```
The `vertex-ex-cc` service agent is created "lazily" on first use of the API, rather than on API enablement.

Fix: Set `skip_iam: true` for `vertex-ex-cc`in `modules/project/service-agents.yaml`.

### Release version

67b44ee

### Terraform version

>= 1.10.0

### Steps to reproduce

Create a new GCP project with  `aiplatform.googleapis.com`  enabled:

```
services:
  -  aiplatform.googleapis.com
```

### Expected behavior

1. Project is created
2. `aiplatform.googleapis.com` is enabled
3. `service-<PROJECT_NUMBER>@gcp-sa-vertex-ex-cc.iam.gserviceaccount.com` is granted `roles/aiplatform.extensionCustomCodeServiceAgent`
4. Apply successful

### Actual behavior

1. Project is created
2. `aiplatform.googleapis.com` is enabled
3. `Error 400: Service account service-<PROJECT_NUMBER>@gcp-sa-vertex-ex-cc.iam.gserviceaccount.com does not exist., badRequest`
4. Apply failed

PR #4005, opened 2026-05-31
Allow skipping default IAM roles for specific service agents in modules/project

### LINKED ISSUE #3998: [FR]: Allow excluding specific service agents from default IAM role grants
## The feature request

Add a module-level option to exclude specific service agents from automatic default IAM role grants.

This would be similar in spirit to the existing `skip_iam` metadata flag in `service-agents.yaml`, but configurable by module consumers at runtime.

## Use case

When upgrading from older Fabric versions to Fabric v50+, we hit a case where the project module tries to grant a default IAM role to a Google-managed service agent that does not exist yet in the project.

The specific case:

- **API**: `aiplatform.googleapis.com`
- **Service agent**: `vertex-moss-ft`
- **Email format**: `service-PROJECT_NUMBER@gcp-sa-vertex-moss-ft.iam.gserviceaccount.com`
- **Role**: `roles/aiplatform.tuningServiceAgent`
- **Feature**: Vertex AI Managed OSS Fine Tuning

This service agent appears to be created lazily by Google Cloud only when OSS model fine-tuning is actually used. In projects where `aiplatform.googleapis.com` is enabled but OSS model fine-tuning has never been triggered, the service agent may not exist.

Fabric still tries to create the IAM member because the service agent is listed in `service-agents.yaml` with a default role and `skip_iam = false`.

This causes `terraform apply` to fail:

```
Error: Request `Create IAM Members roles/aiplatform.tuningServiceAgent serviceAccount:service-PROJECT_NUMBER@gcp-sa-vertex-moss-ft.iam.gserviceaccount.com for project "PROJECT_ID"` returned error: 
Error applying IAM policy for project "PROJECT_ID": Error setting IAM policy for project "PROJECT_ID": googleapi: Error 400: 
Service account service-PROJECT_NUMBER@gcp-sa-vertex-moss-ft.iam.gserviceaccount.com does not exist., badRequest
```

## Current behavior

The module currently supports:

```hcl
service_agents_config = {
  grant_default_roles = false
}
```

However, this is too broad. It disables default IAM role grants for **all** known service agents, not only the problematic one.

In our environment, setting `grant_default_roles = false` produced a plan to remove 160+ existing service-agent IAM grants across multiple projects, which is not safe as a workaround.

## Current workarounds

Users must choose between imperfect options:

1. **Set `grant_default_roles = false`** - Too broad, can remove many existing grants
2. **Manually trigger the feature to force SA creation** - Not always possible, may require org policy changes
3. **Vendor/fork the module** - Works, but creates maintenance burden
4. **Patch `service-agents.yaml` globally** - Only works if SA should be skipped everywhere, not flexible per-project

## Proposed solution

Add a selective exclusion option to `service_agents_config`:

```hcl
variable "service_agents_config" {
  description = "Automatic service agent configuration options."
  type = object({
    create_primary_agents            = optional(bool, true)
    grant_default_roles              = optional(bool, true)
    grant_service_agent_editor       = optional(bool, true)
    skip_default_role_service_agents = optional(set(string), [])
  })
  default  = {}
  nullable = false
}
```

Update the `service_agent_roles` filter in `service-agents.tf`:

```hcl
service_agent_roles = {
  for agent in local._sa : (agent.name) => {
    role      = agent.role
    iam_email = agent.iam_email
  }
  if alltrue([
    var.service_agents_config.grant_default_roles,
    agent.role != null,
    !agent.skip_iam,
    !contains(var.service_agents_config.skip_default_role_service_agents, agent.name)
  ])
}
```

**Example usage:**

```hcl
module "project" {
  source = "github.com/GoogleCloudPlatform/cloud-foundation-fabric//modules/project?ref=v56.0.0"

  name = "my-project"

  service_agents_config = {
    skip_default_role_service_agents = ["vertex-moss-ft"]
  }
}
```

## Benefits

- Keeps `grant_default_roles = true` for all other service agents
- Handles lazy-created service agents gracefully
- Minimal code change
- Gives consumers a safe escape hatch for edge cases

## Additional context

**Related issues:**
- #3900 - Similar problem with non-existing aiplatform service agents after upgrade
- #3893 - Precedent for skipping specific service agents (global skip via `skip_iam`)

This feature request is for cases where consumers need a **runtime/project-level exclusion** instead of a global skip.

We implemented this workaround by vendorizing Fabric v50 with this exact patch. It successfully changed the plan from failing to clean (0 to add, 0 to change, 0 to destroy).

Happy to submit a PR if maintainers agree with the approach.

juliocc:
Allow the project module to selectively skip automatic default IAM role grants for specific service agents.

This addresses a common issue where some APIs are enabled, but some of their associated service agents (such as `vertex-moss-ft`, `aiplatform-pie`, or `vertex-sandbox`) are created lazily by GCP only when their respective features are first used. Because Fabric tries to grant default roles to all known service agents immediately upon API activation, `terraform apply` fails with a 400 error if those lazy agents do not exist yet.

While some agents are skipped globally, this change provides a flexible, project-level escape hatch to bypass errors for new or untracked lazy agents without having to disable all default role grants globally.

The division of labor is designed as follows:
1. **Static Skips (Global):** Managed via `tools/build_service_agents.py` and marked as `skip_iam: true` in `service-agents.yaml` for known platform-level lazy agents.
2. **Dynamic Skips (Runtime):** Managed via the new `service_agents_config.skip_iam` set variable for project-specific overrides.

Closes #3998
Addresses #3900, #3788

---
**Checklist**

### DIFF
--- a/modules/project/README.md
+++ b/modules/project/README.md
@@ -13,6 +13,7 @@ This module implements the creation and management of one GCP project including 
   - [Additive IAM](#additive-iam)
   - [Service Agents](#service-agents)
     - [Cloudservices Editor Role](#cloudservices-editor-role)
+    - [Skipping Service Agent IAM Grants](#skipping-service-agent-iam-grants)
     - [Service Agent Aliases](#service-agent-aliases)
 - [Shared VPC](#shared-vpc)
 - [Organization Policies](#organization-policies)
@@ -269,6 +270,28 @@ The complete list of Google Cloud service agents, including their names, default
 #### Cloudservices Editor Role
 
 The `cloudservices` service agent is granted `roles/editor` by default, making it easy to accidentally remove this binding when managing the editor role authoritatively. In those cases, the module auto-injects the `cloudservices` service agent to preserve the binding. This behaviour is disabled when the `service_agents_config.grant_service_agent_editor` variable is set to `false`.
+
+#### Skipping Service Agent IAM Grants
+
+In some cases, you might want to prevent the module from automatically granting default roles to specific service agents (for example, if the service agent is created lazily by GCP and does not exist yet). You can do this by listing the agent names in `service_agents_config.skip_iam`:
+
+```hcl
+module "project" {
+  source          = "./fabric/modules/project"
+  billing_account = var.billing_account_id
+  name            = "project"
+  parent          = var.folder_id
+  prefix          = var.prefix
+  services = [
+    "container.googleapis.com",
+    "run.googleapis.com"
+  ]
+  service_agents_config = {
+    skip_iam = ["serverless-robot-prod"]
+  }
+}
+# tftest modules=1 resources=7 inventory=service-agents-skip.yaml
+```
 
 #### Service Agent Aliases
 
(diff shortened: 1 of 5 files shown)


ludoo (approved):

ludoo (commented):
We probably also want this in the YAMLs/project factory, etc eventually

PR #4007, opened 2026-06-01
Support `service_agents_config.skip_iam` in project-factory and fast stages

juliocc:
Expose the new `service_agents_config` variable in the `project-factory` module and update the project JSON schemas in both the `project-factory` and `fast` stages to support the `skip_iam` option introduced in #4005.
This enables factory consumers to skip IAM role assignment for specific service agents when creating projects via the project factory.

---
**Checklist**

### DIFF
--- a/modules/project-factory/projects.tf
+++ b/modules/project-factory/projects.tf
@@ -162,10 +162,15 @@ module "projects" {
   logging_sinks         = try(each.value.logging_sinks, {})
   notification_channels = try(each.value.notification_channels, null)
   quotas                = each.value.quotas
+  # Most service agent permissions must be granted in this first pass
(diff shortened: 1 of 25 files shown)


modules/project-factory/projects.tf
```diff
@@ -244,8 +249,10 @@ module "projects-iam" {
   ))
   pam_entitlements = try(each.value.pam_entitlements, {})
   service_agents_config = {
-    create_primary_agents = false
-    grant_default_roles   = false
```
ludoo, line 248: IMHO we need to keep this, we should never grant roles in this step

ludoo (approved):

PR #4009, opened 2026-06-02
Enable managed connection pooling in modules/alloydb

btuckVZ:
Enables the `connection_pool_config` section for the `google_alloydb_instance` resource.

---
**Checklist**

### DIFF
--- a/modules/alloydb/variables.tf
+++ b/modules/alloydb/variables.tf
@@ -87,6 +87,12 @@ variable "cluster_name" {
   nullable    = false
 }
 
+variable "connection_pool_flags" {
+  description = "Map FLAG_NAME=>VALUE for managed connection pooling. Remove the \"connection-pooling-\" prefix and use underscores instead of dashes in the name. For example, \"connection-pooling-pool-mode\" would be \"pool_mode\"."
+  type        = map(string)
+  default     = null
+}
+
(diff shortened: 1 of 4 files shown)


modules/alloydb/variables.tf
```diff
@@ -87,6 +87,12 @@ variable "cluster_name" {
   nullable    = false
 }
 
+variable "connection_pool_flags" {
+  description = "Map FLAG_NAME=>VALUE for managed connection pooling. Remove the \"connection-pooling-\" prefix and use underscores instead of dashes in the name. For example, \"connection-pooling-pool-mode\" would be \"pool_mode\"."
```
juliocc: > Remove the \"connection-pooling-\" prefix and use underscores instead of dashes in the name. For example, \"connection-pooling-pool-mode\" would be \"pool_mode\"."

This is the kind of thing that should probably go in a validation block

juliocc (changes requested):

modules/alloydb/variables.tf
btuckVZ: > This is the kind of thing that should probably go in a validation block

Makes sense to me, just pushed a commit to add it.

juliocc (approved):

PR #4017, opened 2026-06-10
Add net-lb-proxy-int-cross-region module and tests

### LINKED ISSUE #3922: [FR]: Add net-lb-proxy-int-cross-region module
### The feature request

In some scenarios (e.g. Disaster Recovery) it could be useful to use the following architecture:

Image

Sources:
- https://docs.cloud.google.com/load-balancing/docs/tcp/internal-proxy#cross-region

### Proposed solution

The net-lb-app-int-cross-region module could be simplified to obtain a new module named net-lb-proxy-int-cross-region.

### Additional context

_No response_

sruffilli:
First stab - fixes #3922


juliocc (approved):
2026, otherwise <3

PR #4019, opened 2026-06-10
feat(0-org-setup): make predefined iam_principals groups configurable via var.groups

IrfanAnsari:
**What:** add a groups variable to 0-org-setup controlling which predefined group names are derived into iam_principals from the org domain. Default= the current 7, so fully backward-compatible.

**Why:** consumers that only bind gcp-organization-admins currently carry 6 unused group:…@domain entries in the published iam_principals map / downstream globals. This lets them trim to the groups they actually use (e.g. groups = ["gcp-organization-admins"]).

**Tests:** existing cases pass unchanged (backward-compat); new groups tftest case proves the trim. terraform fmt, tfdoc, and the stage tests pass.

### DIFF
--- a/fast/stages/0-org-setup/README.md
+++ b/fast/stages/0-org-setup/README.md
@@ -900,7 +900,6 @@ Define values for the `var.environments` variable in a tfvars file.
 | [cicd-workflows-preconditions.tf](./cicd-workflows-preconditions.tf) | None |  | <code>terraform_data</code> |
 | [cicd-workflows.tf](./cicd-workflows.tf) | None | <code>iam-service-account</code> | <code>google_storage_bucket_object</code> · <code>local_file</code> |
 | [factory.tf](./factory.tf) | None | <code>net-vpc-factory</code> · <code>project-factory</code> |  |
-| [identity-providers-defs.tf](./identity-providers-defs.tf) | None |  |  |
 | [imports.tf](./imports.tf) | None |  |  |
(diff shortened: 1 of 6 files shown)


ludoo:
Thanks for the PR, I personally am a bit ambivalent about this and would instead prefer to remove support for the old inferred groups for three reasons

- we have now fully moved to `defaults.yaml` and we should not add low level configuration via variables which should only support context injection and dataset/factory choices
- the old assumption that a landing zone requires a specific set of personas (groups) does not hold true anymore, FAST now supports wildly different configurations and use cases which use different principals and patterns
- we are now seeing FAST used in environments (sovereign clouds) where groups and Cloud Identity do not even exist,  and we fully support those as first class use cases

Context+principals is the right modern approach for FAST which gives users the required amount of flexibility, and avoids "magic" (infer x groups and inject them in context). Configuring groups in the defaults YAML is one line per group and is well documented.

I will send a PR to remove support for "magic groups" and let's see what @juliocc and @sruffilli think about this so we close the circle.

juliocc:
+1 to remove. The groups variable is vestigial remnant of earlier FAST versions.

IrfanAnsari:
Makes total sense — thanks for the thoughtful response, and I'm fully on board. The variables-vs-defaults.yaml boundary is the right line, and removing the inference is cleaner than making it configurable (it also removes the unused templated @domain principals that prompted this PR in the first place — the published iam_principals map carried 6 groups nothing bound).

IrfanAnsari:
I am happy to close this as no longer required.

PR #4023, opened 2026-06-12
Consistent types for_each net-vpc-firewall in custom_rules dynamic blocks

IrfanAnsari:
The allow/deny dynamic blocks used
  `for_each = each.value.action == "X" ? each.value.rules : {}`. The branches have
  inconsistent types (a populated object vs the empty object `{}`), so a plan fails
  with "Inconsistent conditional result types" when each.value.rules is
  heterogeneous — i.e. some rule entries set `ports` and some don't. The typed
  ingress_rules/egress_rules path coerces entries to one type and hides it; factory
  rules (yamldecode, untyped) don't, so it surfaces there. It's plan-time, so
  `validate` misses it.

  Use a filtered for-expression instead — always a map of consistent type, identical
  output:
    for_each = { for k, v in each.value.rules : k => v if each.value.action == "X" }

  Adds a factory example with a mixed icmp/tcp rule that reproduces it.

---
**Checklist**

### DIFF
--- a/modules/net-vpc-firewall/README.md
+++ b/modules/net-vpc-firewall/README.md
@@ -194,7 +194,7 @@ module "firewall" {
   }
   default_rules_config = { disabled = true }
 }
-# tftest modules=1 resources=3 files=lbs,cidrs inventory=factory.yaml
+# tftest modules=1 resources=4 files=lbs,cidrs inventory=factory.yaml
 ```
 
 ```yaml
@@ -220,6 +220,16 @@ ingress:
         ports:
           - 80
           - 443
+  allow-admin-mixed:
+    description: Allow ICMP plus SSH/RDP from the admin range (mixed protocols).
+    source_ranges:
+      - 10.0.0.0/8
+    rules:
+      - protocol: icmp
+      - protocol: tcp
+        ports:
+          - 22
+          - 3389
 egress:
   block-telnet:
     description: block outbound telnet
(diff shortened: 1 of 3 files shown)


IrfanAnsari:
Thanks for the review! I actually tried `: []` first, but it hits the **same** `Inconsistent conditional result types` error. `each.value.rules` here isn't the raw `list(object(…))` from the variable — a few lines up the module rebuilds it as an index-keyed map:

```hcl
rules = { for k, v in rule.rules : k => v }   # main.tf
```

So by the time the `allow`/`deny` blocks run, `each.value.rules` is a **map/object**, not a list. That's why the original `: {}` failed (object vs empty-object), and `: []` fails the same way (object vs empty-list):

```
Error: Inconsistent conditional result types
  181:     for_each = each.value.action == "DENY" ? each.value.rules : []
  │ each.value.rules is object with 2 attributes
```

(reproduced via a factory rule mixing `icmp` (no ports) and `tcp` (ports) — the heterogeneous case that triggers the bug.)

The `for` expression works precisely because it **isn't** a conditional — it's a single expression that always evaluates to a map (empty when the predicate is false), so there are no two branches to type-unify.

One related note: the `ports = try(rule.value.ports, [])` in the `content` block is also required — in the heterogeneous case the `icmp` entry has no `ports` attribute, so a bare `rule.value.ports` would error.

Happy to switch to another form if you'd prefer one, but both `: []` and `: {}` trip on the map-typed `each.value.rules`. Open to suggestions!

sruffilli (approved):

PR #4025, opened 2026-06-12
Add server TLS policy support to net-lb-app-int-cross-region

LF3551:
This adds a passthrough for `server_tls_policy` on the cross-region internal Application Load Balancer HTTPS proxy configuration.

The field lets users attach a Network Security `ServerTlsPolicy` to the target HTTPS proxy, which is required for frontend mTLS on cross-region internal Application Load Balancers. The module does not create the trust config or server TLS policy; it only allows an existing policy to be attached.

Google Cloud mTLS docs: https://cloud.google.com/load-balancing/docs/mtls

Checks run:
- `terraform fmt -recursive modules/net-lb-app-int-cross-region`
- `.venv/bin/python tools/tfdoc.py modules/net-lb-app-int-cross-region`
- `.venv/bin/pytest tests/modules/net_lb_app_int_cross_region`

### DIFF
--- a/modules/net-lb-app-int-cross-region/README.md
+++ b/modules/net-lb-app-int-cross-region/README.md
@@ -870,9 +870,9 @@ When deploying changes to load balancer configuration please refer to [net-lb-ap
 
 | name | description | type | required | default |
 |---|---|:---:|:---:|:---:|
-| [name](variables.tf#L88) | Load balancer name. | <code>string</code> | ✓ |  |
-| [project_id](variables.tf#L169) | Project id. | <code>string</code> | ✓ |  |
-| [vpc_config](variables.tf#L211) | VPC-level configuration. | <code>object&#40;&#123;&#8230;&#125;&#41;</code> | ✓ |  |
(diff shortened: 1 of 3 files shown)


modules/net-lb-app-int-cross-region/README.md
sruffilli, line 1: I would add an example to the readme, both for documentation purpose and to trigger a test.

LF3551: Added a ### mTLS with Server TLS Policy example to the README demonstrating server_tls_policy usage. The example includes a # tftest directive, so it's picked up by the test suite automatically.

sruffilli (commented):
PTAL at the new merge conflict as the LB modules went through a variable interface refactor. Other than that LGTM, happy to approve once merged/rebased successfully

LF3551:
Rebased on latest master and resolved the conflict. tfdoc regenerated, mTLS example test passes. PTAL.

sruffilli (approved):

PR #4027, opened 2026-06-15
Standardize net-lb family interfaces and features

ludoo:
This PR completes the implementation of the L7 load balancer interface standardization as per [ADR 20260615](adrs/20260615-net-lb-interface-standardization.md).

Key changes:
- Resolved missing context lookups (zone, region, IP address) in `net-lb-app-ext`, `net-lb-app-int`, and `net-lb-app-int-cross-region` modules for NEGs and instance groups.
- Switched to robust try-based lookups to handle optional null values in configuration schemas without failing.
- Updated test configurations and regenerated inventories for affected module-level tests and README examples.
- Verified all tests pass and ran linting checks (check docs, yaml lint, boilerplate).
- Added a "Testing" section to the ADR detailing the required context variables and tfvars used for validation.
- **Fixed `apigee` and `api-gateway` recipes** to use the standardized `group` key instead of the old `backend` key in `net-lb-app-ext` module calls.

E2E live tests executed (apply/destroy) in playground environment:
- `net-lb-ext`: MIG Scenario, Context Scenario
- `net-lb-proxy-int`: MIG Scenario
- `net-lb-proxy-int-cross-region`: Multi-region MIGs (verified existing context support works with standardized format)
- `net-lb-app-int`: MIG with Actions
- `net-lb-app-int-cross-region`: Multi-region MIGs
- `net-lb-app-ext`: Cloud Run Scenario, MIG Scenario
- `net-lb-app-ext-regional`: MIG Scenario

---
**Breaking Changes**

```upgrade-note
`modules/net-lb-proxy-int`: Removed connection tracking and failover configurations as they are not supported by the underlying resources.
```
```upgrade-note
`modules/net-lb-app-int`: Removed failover configurations.
```
```upgrade-note
`modules/net-lb-app-ext`: Renamed the `backend` key to `group` inside `backend_service_configs` to standardize the backend interface. Removed failover configurations.
```
```upgrade-note
`modules/net-lb-app-ext-regional`: Renamed the `backend` key to `group` inside `backend_service_configs`. Replaced the `vpc` variable with `vpc_config` to match other regional modules. Removed failover configurations.
```

### DIFF
--- a/adrs/20260615-net-lb-interface-standardization.md
+++ b/adrs/20260615-net-lb-interface-standardization.md
@@ -1,0 +1,198 @@
+# Standardizing Interfaces and Features in the `net-lb` Module Family
+
+**authors:** Antigravity (AI Assistant)
+**date:** June 15, 2026
+
+## Status
+
+Proposed
+
+## Context
+
+The `net-lb` family in Cloud Foundation Fabric (CFF) consists of 8 modules covering different regional, global, and cross-region load balancer types (L4 and L7, internal and external):
+
+| Module | LB Type | Scope | Protocol |
+| :--- | :--- | :--- | :--- |
+| [`net-lb-int`](file:///home/ludomagno/dev/tf-playground/cloud-foundation-fabric/modules/net-lb-int) | Internal Passthrough NLB | Regional | L4 (TCP/UDP) |
+| [`net-lb-ext`](file:///home/ludomagno/dev/tf-playground/cloud-foundation-fabric/modules/net-lb-ext) | External Passthrough NLB | Regional | L4 (TCP/UDP) |
+| [`net-lb-proxy-int`](file:///home/ludomagno/dev/tf-playground/cloud-foundation-fabric/modules/net-lb-proxy-int) | Internal Proxy LB | Regional | L4 (TCP) |
+| [`net-lb-proxy-int-cross-region`](file:///home/ludomagno/dev/tf-playground/cloud-foundation-fabric/modules/net-lb-proxy-int-cross-region) | Internal Proxy LB | Cross-Region | L4 (TCP) |
+| [`net-lb-app-int`](file:///home/ludomagno/dev/tf-playground/cloud-foundation-fabric/modules/net-lb-app-int) | Internal Application LB | Regional | L7 (HTTP/S) |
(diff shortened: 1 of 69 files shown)


adrs/20260615-net-lb-interface-standardization.md
```diff
@@ -0,0 +1,198 @@
+# Standardizing Interfaces and Features in the `net-lb` Module Family
+
+**authors:** Antigravity (AI Assistant)
```
juliocc, line 3: and you :)

juliocc (approved):

sruffilli:
I love the initiative, and I have a proposal on top: can we either document the upgrade path or (my fav option, albeit unpopular) provide some kind of upgrade script and/or prompt to ensure a smooth transition? I understand this would be a first and potentially deserve its own ADR, but while we're here....

ludoo:
> I love the initiative, and I have a proposal on top: can we either document the upgrade path or (my fav option, albeit unpopular) provide some kind of upgrade script and/or prompt to ensure a smooth transition? I understand this would be a first and potentially deserve its own ADR, but while we're here....

You mean upgrade path for module changes? IMHO it's what we do with breaking changes in PRs. Usually it's a matter of slightly renaming variables. Let's discuss if there's a simple way of doing it better.

sruffilli:
> You mean upgrade path for module changes? IMHO it's what we do with breaking changes in PRs. Usually it's a matter of slightly renaming variables. Let's discuss if there's a simple way of doing it better.

What I'd like as a user is some guidance/automation that actually adjusts my code to adapt to the new interface - if I'm not mistaken we've been providing guidance (I - for one - might have provided loose guidance in the past), but never went beyond `moved` blocks to smoothen the experience, which wouldn't apply here.

ludoo:
Well, isn't this what agents are for? :)

sruffilli:
> Well, isn't this what agents are for? :)

Hence my suggestion for a prompt above :) or at least some guidance to use one and how.

ludoo:
Ahhhh ok, then yes absolutely. Try and think where that could live, maybe in the changelog itself, in a new specific block in PR descriptions which we extract like we do for breaking changes?

LucaPrete:
LGTM.
Important note: LBs are really tricky. We should make sure we test e2e ALL examples to avoid our/AI hallucinations.

ludoo:
I know Luca, we were discussing the same with Simo and it's exactly the pattern I followed when I originally developed those modules. The supportability matrix is crazy and needs actual verification via apply.

PR #4048, opened 2026-07-06
Converting iam principals to deny policies format

leszek-grudzien:
Adding an automatic conversion of a user/group/service account principal injected via FAST interpolation (using "$iam_principals:") to format required by IAM Deny policies. 

In the current version, when one tries to configure principals in either denied_principals or exception_principals block using iam_principals: interpolation directly eg.
denied_principals:
  - $iam_principals:gcp-organization-admins
  - $iam_principals:service_accounts/iac-0/iac-org-rw
exception_principals:
  - $iam_principals:super-admin

they get resolved to classic IAM principals (group:gcp-organization-admin@domain / serviceAccount:iac-org-rw@project-id.iam.gserviceaccount.com / user:super-admin@domain) which are not accepted by IAM Deny policies (see details [here](https://docs.cloud.google.com/iam/docs/principal-identifiers#deny)).  An attempt to deploy such a policy results in 400 Bad request ... "The principal specified by (group:gcp-organization-admins@domain) is in an incorrect format." error. In fact even the automatic tests that we have for that part accept incorrect values (that's why this PR adapts them as well).

Potential workaround is to define again the same values for users/groups/service accounts via context eg.:
iam_principals:
  gcp-organization-admins-uri: principalSet://goog/group/gcp-organization-admin@domain

and use them in deny policies. But that means: 1) doubled effort 2) less concise configuration, using 2 different flavors.

The idea of this PR is to provide automatic conversion between formats for the simplest and most common cases: a user / a group / a service account (Service agent). The logic uses a prefix (before colon) from a principal (after interpolation) to do the lookup of a respective template which is then parametrized.

For other more complicated cases (as eg. WIF pools / all SAs in a given project/folder or org) IAM Allow policies already use the same syntax as Deny Policies do so there is no need to convert them.

---
**Checklist**


modules/folder/deny-policies.tf
```diff
@@ -28,7 +39,15 @@ resource "google_iam_deny_policy" "default" {
       description = rule.value.description
       deny_rule {
         denied_principals = [
-          for p in rule.value.denied_principals : lookup(local.ctx.iam_principals, p, p)
+          for pf in [
```
ludoo, line 42: this feels very hackish, let's discuss the original issue and flesh out alternate approaches; please document the issue clearly in the PR description with examples

leszek-grudzien: Ok, I've just enhanced the description above.

ludoo:
Leszek, also had a chat with @juliocc and we both agree on not implementing this, as it adds "magic" and opacity which is an antipattern for our modules. Happy to chat offline if you want to explore different solutions.

PR #4068, opened 2026-07-24
Add HTTP keep alive timeout attribute to net-lb-app-ext-regional module

darvabus:
…ute for HTTP and HTTPS proxy config variables. This to allow modification of this parameter for Regional Ext App LB

---
**Checklist**

- [X ] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)

### DIFF
--- a/modules/net-lb-app-ext-regional/main.tf
+++ b/modules/net-lb-app-ext-regional/main.tf
@@ -75,6 +75,7 @@ resource "google_compute_region_target_http_proxy" "default" {
   region      = local.region
   name        = coalesce(var.http_proxy_config.name, var.name)
   description = var.http_proxy_config.description
+  http_keep_alive_timeout_sec      = var.https_proxy_config.http_keepalive_timeout
   url_map     = google_compute_region_url_map.default.id
 }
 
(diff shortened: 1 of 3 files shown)


modules/net-lb-app-ext-regional/main.tf
```diff
@@ -75,6 +75,7 @@ resource "google_compute_region_target_http_proxy" "default" {
   region      = local.region
   name        = coalesce(var.http_proxy_config.name, var.name)
   description = var.http_proxy_config.description
+  http_keep_alive_timeout_sec      = var.https_proxy_config.http_keepalive_timeout
```
juliocc: It seems this field is also relevant to global Global L7 XLM. Can you confirm and update the global module too if so?

juliocc: It also seems this field cannot be updated:

> For a regional external Application Load Balancer, you cannot update the keepalive timeout parameter of a regional target HTTP(S) proxy directly. To update the keepalive timeout parameter of a regional target proxy, you need to do the following

A note or warning somewhere might be useful.

juliocc (changes requested):

modules/net-lb-app-ext-regional/main.tf
darvabus: Hello,
Thanks for your feedback.

Regarding the Global L7 XLM, the field is already present in the module, hence I didn't have to modify it.

Regarding the modification, I follow what you mean. 
Where in the README could I add this, if you can help me assess the best place in the document ?

Thank you.

juliocc (approved):

PR #4081, opened 2026-07-29
Fix project ID mismatch in logging bucket sink IAM conditions

### LINKED ISSUE #4030: [bug] folder log sinks fail with log_bucket_not_found due to Project ID vs Project Number mismatch in IAM CEL condition
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Other

### The problem

When creating folder-level logging sinks targeting a log bucket in a different project using the CFF folder module, log routing fails with the error log_bucket_not_found.

This is caused by a latent mismatch in the automatic project-level IAM binding (google_project_iam_member.bucket_sinks_binding) generated by the module:

The module generates a CEL condition that matches the target resource name against the Project ID (string).
However, Google Cloud's IAM CEL condition evaluator evaluates resource.name using the numeric Project Number.
As a result, the condition evaluates to false, the permission is denied, and the sink is unable to write logs to the target bucket.

### Release version

v56.1.0

### Terraform version

Terraform v1.14.8

### Steps to reproduce

Instaniate CFF folder module with a folder log sink pointing to a log bucket:
hcl

```hcl
module "folder" {
  source        = "git::https://github.com/GoogleCloudPlatform/cloud-foundation-fabric//modules/folder"
  id            = "folders/123456789"
  folder_create = false
  logging_sinks = {
    my-sink = {
      destination = "logging.googleapis.com/projects/my-observe-project/locations/us-central1/buckets/my-log-bucket"
      type        = "logging"
    }
  }
}
```
Apply the HCL.
Check the folder's logging sinks in the GCP Console or run gcloud logging read "resource.type=\"logging_sink\" AND severity=ERROR" --folder=123456789. You will find the routing error log_bucket_not_found.

### Expected behavior

Folder's gcp-sa-logging agent should be able to write logs to log sink logging bucket and shouldnt get IAM issue.

### Actual behavior

In [modules/folder/logging.tf](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/modules/folder/logging.tf), the project-level IAM binding for logging bucket sinks is defined as:

```hcl
resource "google_project_iam_member" "bucket_sinks_binding" {
  for_each = local.sink_bindings["logging"]
  project  = split("/", each.value.destination)[1]
  role     = "roles/logging.bucketWriter"
  member   = google_logging_folder_sink.sink[each.key].writer_identity
  condition {
    title       = "${each.key} bucket writer"
    description = "Grants bucketWriter to serviceAccount:${google_logging_folder_sink.sink[each.key].writer_identity} used by log sink ${each.key} on folders/${local.folder_id}"
    expression  = "resource.name.endsWith('projects/${split("/", each.value.destination)[1]}/locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
  }
}
```
The Bug:
The expression uses split("/", each.value.destination)[1], which extracts the Project ID string (e.g., hhs-observe-p-mxty) because the destination attribute is passed in from Terraform resource IDs.

This generates a CEL condition: resource.name.endsWith('projects/PROJECT_ID/locations/LOCATION/buckets/BUCKET_ID')

During live log routing, GCP IAM evaluates this condition. However, GCP's IAM policy engine populates resource.name using the numeric Project Number (e.g., 650484525631) instead of the Project ID.

Because projects/650484525631/... does not end with projects/PROJECT_ID/..., the condition evaluates to false, causing log routing to fail silently or log a log_bucket_not_found configuration error.

### Additional context

Proposed Fix
The CEL expression should be updated to be project-number-independent or robust enough to support both formats.

Since the target bucket's location and bucket ID are the only parts needed to uniquely identify the destination bucket on the project, the expression can be simplified using endsWith to match only the trailing path:

hcl

expression = "resource.name.endsWith('locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
Alternatively, the module can resolve the project number from the project resource (if available in context) or allow it to be passed in, but matching on the trailing location and bucket ID path is the cleanest and most robust project-number-independent solution.

sruffilli:
Fixes #4030.

### Problem
When creating logging sinks targeting a Cloud Logging bucket in a project using the `billing-account`, `folder`, `organization`, or `project` modules, log routing can fail with `log_bucket_not_found` or permission denied errors.

In `google_project_iam_member.bucket_sinks_binding`, the CEL condition expression was previously defined as:
```hcl
expression = "resource.name.endsWith('${each.value.destination}')"
```
When `destination` is configured with a project ID string (e.g., `projects/my-project/locations/europe-west8/buckets/audit-logs`), this generated the CEL expression:
```
resource.name.endsWith('projects/my-project/locations/europe-west8/buckets/audit-logs')
```
However, Google Cloud IAM evaluates `resource.name` at runtime using the numeric Project Number (e.g., `projects/123456789012/locations/europe-west8/buckets/audit-logs`). Because the runtime resource name starts with `projects/123456789012/...` instead of `projects/my-project/...`, the `.endsWith()` condition evaluated to `false`, denying `roles/logging.bucketWriter` to the sink's service account.

### Solution
Since `google_project_iam_member.bucket_sinks_binding` is already explicitly attached to the target project (`project = split("/", each.value.destination)[1]`), a Cloud Logging bucket within that project is uniquely identified by its location and bucket ID.

We updated the CEL condition expression across `billing-account`, `folder`, `organization`, and `project` modules to match on the trailing location and bucket path:
```hcl
expression = "resource.name.endsWith('locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
```
This produces the project-number-independent expression:
```
resource.name.endsWith('locations/europe-west8/buckets/audit-logs')
```
which evaluates correctly regardless of whether `destination` was configured with a string Project ID or numeric Project Number.

#### 1. Live Sandbox E2E Deployment
We performed an end-to-end sandbox deployment test against a live GCP test project:
- Deployed a test Cloud Logging bucket (`locations/global/buckets/test-bucket`) using `modules/logging-bucket`.
- Deployed a logging sink and conditional IAM binding using `modules/project`.
- Verified that `terraform apply` succeeded cleanly and that GCP IAM accepted the updated condition expression on `roles/logging.bucketWriter` without errors.
- Verified cleanly destroying all resources via `terraform destroy`.

#### 2. IAM Policy Troubleshooter Verification (`policytroubleshooter.googleapis.com`)
To verify the exact runtime evaluation behavior of Google Cloud's IAM policy engine, we deployed two `roles/logging.bucketWriter` conditional bindings on a test project and evaluated them using the official IAM Policy Troubleshooter API (`gcloud policy-troubleshoot iam`) against the runtime target resource name using a numeric Project Number (`//logging.googleapis.com/projects/123456789012/locations/global/buckets/test-bucket`):

- **Old Condition** (`resource.name.endsWith('projects/my-project/locations/global/buckets/test-bucket')`):
  - Result: **`NOT_GRANTED` (`false`)** — Denied because `projects/123456789012/...` does not match the string project ID prefix.
- **New Condition** (`resource.name.endsWith('locations/global/buckets/test-bucket')`):
  - Result: **`GRANTED` (`true`)** — Successfully matched the trailing path and granted `logging.buckets.write`.

#### 3. Automated Test Suite
- Regenerated and verified all module-level plan inventories (`tftest.yaml`) and example inventories across `modules/billing-account`, `modules/folder`, `modules/organization`, and `modules/project`.
- Verified all documentation tables (`check_documentation.py`), copyright boilerplate (`check_boilerplate.py`), and Terraform formatting (`terraform fmt`).

### DIFF
--- a/modules/billing-account/logging.tf
+++ b/modules/billing-account/logging.tf
@@ -42,6 +42,10 @@ locals {
       }
     }
   )
+  sink_bucket_expressions = {
+    for name, sink in local.sink_bindings["logging"] :
+    name => "resource.name.endsWith('locations/${split("/", sink.destination)[3]}/buckets/${split("/", sink.destination)[5]}')"
+  }
 }
 
 resource "google_logging_billing_account_sink" "sink" {
@@ -124,7 +128,7 @@ resource "google_project_iam_member" "bucket_sinks_binding" {
   condition {
     title       = "${each.key} bucket writer"
     description = "Grants bucketWriter to ${google_logging_billing_account_sink.sink[each.key].writer_identity} used by log sink ${each.key} on billing account ${var.id}"
-    expression  = "resource.name.endsWith('locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
+    expression  = local.sink_bucket_expressions[each.key]
   }
 }
 
--- a/modules/billing-account/logging.tf
+++ b/modules/billing-account/logging.tf
@@ -124,7 +124,7 @@ resource "google_project_iam_member" "bucket_sinks_binding" {
   condition {
     title       = "${each.key} bucket writer"
     description = "Grants bucketWriter to ${google_logging_billing_account_sink.sink[each.key].writer_identity} used by log sink ${each.key} on billing account ${var.id}"
-    expression  = "resource.name.endsWith('${each.value.destination}')"
+    expression  = "resource.name.endsWith('locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
   }
 }
 
--- a/modules/folder/logging.tf
+++ b/modules/folder/logging.tf
@@ -167,7 +167,7 @@ resource "google_project_iam_member" "bucket_sinks_binding" {
   condition {
     title       = "${each.key} bucket writer"
     description = "Grants bucketWriter to ${google_logging_folder_sink.sink[each.key].writer_identity} used by log sink ${each.key} on ${local.folder_id}"
-    expression  = "resource.name.endsWith('${each.value.destination}')"
+    expression  = "resource.name.endsWith('locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
   }
 }
 
--- a/modules/folder/logging.tf
+++ b/modules/folder/logging.tf
@@ -54,6 +54,10 @@ locals {
       name => sink if sink.iam && sink.type == type
     }
   }
+  sink_bucket_expressions = {
+    for name, sink in local.sink_bindings["logging"] :
+    name => "resource.name.endsWith('locations/${split("/", sink.destination)[3]}/buckets/${split("/", sink.destination)[5]}')"
+  }
 }
 
 resource "google_logging_folder_settings" "default" {
@@ -167,7 +171,7 @@ resource "google_project_iam_member" "bucket_sinks_binding" {
   condition {
     title       = "${each.key} bucket writer"
     description = "Grants bucketWriter to ${google_logging_folder_sink.sink[each.key].writer_identity} used by log sink ${each.key} on ${local.folder_id}"
-    expression  = "resource.name.endsWith('locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
+    expression  = local.sink_bucket_expressions[each.key]
   }
 }
 
--- a/modules/organization/logging.tf
+++ b/modules/organization/logging.tf
@@ -54,6 +54,10 @@ locals {
       name => sink if sink.iam && sink.type == type
     }
   }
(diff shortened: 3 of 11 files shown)


modules/billing-account/logging.tf
```diff
@@ -124,7 +124,7 @@ resource "google_project_iam_member" "bucket_sinks_binding" {
   condition {
     title       = "${each.key} bucket writer"
     description = "Grants bucketWriter to ${google_logging_billing_account_sink.sink[each.key].writer_identity} used by log sink ${each.key} on billing account ${var.id}"
-    expression  = "resource.name.endsWith('${each.value.destination}')"
+    expression  = "resource.name.endsWith('locations/${split("/", each.value.destination)[3]}/buckets/${split("/", each.value.destination)[5]}')"
```
ludoo: please do this in a local

ludoo (approved):

PR #4082, opened 2026-07-29
Add contributing skill for end-to-end CFF contribution workflows

sruffilli:
### Rationale
Adds `skills/contributing/SKILL.md` to guide AI agents through end-to-end contribution workflows in Cloud Foundation Fabric. The skill supports two entry modes:
- **Mode A (Issue Triage & Fix)**: For assigned or reported GitHub Issues.
- **Mode B (Proactive Development / PR Prep)**: For features, refactoring, or existing branches.

### DIFF
--- a/skills/contributing/SKILL.md
+++ b/skills/contributing/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: contributing
-description: End-to-end workflow for contributing to Cloud Foundation Fabric: triaging GitHub issues, proactive feature development, validating with tests, and submitting sanitized Pull Requests.
+description: "End-to-end workflow for contributing to Cloud Foundation Fabric: triaging GitHub issues, proactive feature development, validating with tests, and submitting sanitized Pull Requests."
 ---
 
 # CFF Contribution Flow Skill
--- a/skills/contributing/SKILL.md
+++ b/skills/contributing/SKILL.md
@@ -1,13 +1,26 @@
 ---
 name: contributing
-description: "End-to-end workflow for contributing to Cloud Foundation Fabric: triaging GitHub issues, proactive feature development, validating with tests, and submitting sanitized Pull Requests."
+description: "End-to-end workflow for contributing to Cloud Foundation Fabric: triaging GitHub issues, proactive feature development, validating with tests and Policy Troubleshooter, and submitting sanitized Pull Requests. Use when addressing a Fabric GitHub issue, developing a module or FAST stage change, or preparing a branch for a pull request."
 ---
 
 # Fabric Contribution Flow Skill
 
 This skill defines the end-to-end workflow for contributing to Cloud Foundation Fabric. It supports two entry modes:
 - **Mode A: Issue Triage & Bug Fix**: You are addressing an assigned or reported GitHub Issue. Start at **Step 1**.
-- **Mode B: Proactive Development & PR Prep**: You are actively building a feature, refactoring code, or preparing an existing branch for a Pull Request. Jump directly to **Step 2**.
+- **Mode B: Proactive Development & PR Prep**: You are actively building a feature, refactoring code, or preparing an existing branch for a Pull Request. Jump directly to the **Design Sign-off Gate** before Step 2.
+
+If it is unclear which mode applies, ask the user before proceeding.
+
+## Human-in-the-Loop Gates
+
+Gate on steps that are hard to reverse, costly, or where your judgment is likely to diverge from a maintainer's. Keep mechanical, reversible steps (linting, tests, doc and inventory regeneration) autonomous. Gates are **blocking**: if you are running non-interactively and cannot get an answer, stop — never assume approval.
+
+| Gate | When | What the human decides |
+| :--- | :--- | :--- |
+| **Triage Disposition** | End of Step 1 (Mode A) | Whether the issue is in scope and worth pursuing: proceed, reject, or escalate. |
+| **Design Sign-off** | Before Step 2 | Approves the proposed design and scope. Mandatory for new features and interface changes; skippable for trivial fixes where the design is self-evident. |
+| **E2E Opt-in** | Step 5 | Whether to run live cloud verification. Providing a sandbox project ID **is** the consent to deploy to it. |
+| **PR Approval** | Step 6 | Reviews the final sanitized PR body before submission. |
 
 ---
 
@@ -16,14 +29,20 @@ This skill defines the end-to-end workflow for contributing to Cloud Foundation 
 ```mermaid
 graph TD
     M1[Mode A: Issue Triage] --> A[1. Triage Issue]
-    M2[Mode B: Proactive Dev] --> B[2. Develop Fix / Feature]
-    A --> B
-    B --> C[3. Run Tests & Lint]
-    C --> G[3b. Live Verification: E2E & Policy Troubleshooter]
-    G --> D[4. Run Code Review]
-    D --> E[5. Align & Fix Feedback]
-    E -->|If Changes Needed| C
-    E -->|If Approved| F[6. Submit Sanitized PR]
+    A --> GT{Gate: Triage Disposition}
+    GT -->|Human approves| GD{Gate: Design Sign-off}
+    GT -->|Human rejects| X[Stop / Report Back]
+    M2[Mode B: Proactive Dev] --> GD
+    GD --> B[2. Develop Fix / Feature]
+    B --> C[3. Tests, Lint & Inventories]
+    C --> D[4. Pre-Submission Self-Review]
+    D -->|Issues found| B
+    D -->|Clean| GE{Gate: E2E Opt-in}
+    GE -->|Project ID provided| G[5. Live E2E Verification]
+    GE -->|Skipped| GS{Gate: PR Approval}
+    G -->|Failures| B
+    G -->|Verified| GS
+    GS --> F[6. Commit & Submit Sanitized PR]
 ```
 
 ### Step 1: Triage the Issue (Mode A Only)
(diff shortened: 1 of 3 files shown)


skills/contributing/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+1.  **Retrieve Issue Details**: Use the GitHub CLI to view the issue context.
+    ```bash
+    gh issue view <issue-number>
+    ```
+
+2.  **Explore the Codebase**: Identify the target module (`modules/<module_name>`) or FAST stage (`fast/stages/<stage_name>`) that requires modification.
+
+3.  **Evaluate Fit & Scope**: Assess whether the issue is relevant for Fabric. Ensure it aligns with Fabric's core philosophy (modules should be lean, composable, and represent a single resource type context). Confirm the change has a sufficiently large scope and represents a generic, valuable addition to the module or FAST stage rather than a highly specific, one-off customization.
+
+4.  **Read Provider Documentation**: If the issue involves Google Cloud resources, retrieve and read the latest version of the documentation for the involved GCP resources or Terraform provider resource/datasource to ensure accurate implementation of its attributes, behaviors, and constraints.
```
juliocc: hint on the provider docs?

e.g. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_address

skills/contributing/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+    > [!WARNING]
+    > Do NOT rely solely on the proposed solution, examples, or partial specifications provided in the issue description. Always retrieve and review the complete documentation schema from the official provider registry.
+    >
+    > **Workaround for Registry Page JS/Redirection Errors**:
+    > If the registry page fails to load properly with `read_url_content` (e.g. returns "Please enable Javascript" or truncates due to HTML formatting issues), find the source markdown file on GitHub and fetch its raw content using `curl` to a temporary file:
+    > ```bash
+    > curl -s https://raw.githubusercontent.com/hashicorp/terraform-provider-google/main/website/docs/r/workstations_workstation_config.html.markdown -o scratch/workstation_config_doc.md
+    > ```
+    > You can then search and view it locally to identify all supported arguments and blocks.
+    > Use this information to make an informed decision on which arguments to support. While 100% coverage is not always necessary or desirable, make a deliberate choice to include all arguments that are useful and align with Fabric's composability and simplicity goals, rather than missing them by omission.
```
juliocc, line 72: If you want to go wild, you could point to the magic modules source too. Probably too much IMO

skills/contributing/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+    > You MUST strictly follow the design principles and coding conventions defined in [CONTRIBUTING.md](../../CONTRIBUTING.md) and [GEMINI.md](../../GEMINI.md) when designing variables, modules, and factories.
+
+2.  **Create a Git Branch**: Create a new git branch named after your username and the feature/fix (e.g., `<username>/<feature-name>`, such as `johndoe/feature-name`).
+    ```bash
+    git checkout -b <username>/<feature-name>
+    ```
+
+3.  **Apply Fabric/FAST Design Conventions**:
+    *   **Context Interpolation**: If context support is relevant and needed for the module (e.g., to support resolving symbolic references like `"$project_ids:myprj"`), add a `context` variable block and implement `ctx`/`ctx_p` locals in `main.tf`. Do not add it blindly if the module does not benefit from symbolic interpolation.
+    *   **Compact Variables**: Leverage objects with `optional()` attributes and default values to keep user interfaces clean.
```
juliocc: prefer optional() without defaults to allow to provider to use whatever default value it defines

skills/contributing/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+          validation {
+            condition     = contains(["STANDARD", "MULTI_REGIONAL", "REGIONAL", "NEARLINE", "COLDLINE", "ARCHIVE"], var.storage_class)
+            error_message = "Storage class must be one of STANDARD, MULTI_REGIONAL, REGIONAL, NEARLINE, COLDLINE, ARCHIVE."
+          }
+        }
+        ```
+
+4.  **Reference Implementations ("Golden Paths")**:
+    *   [compute-vm](../../modules/compute-vm): Ideal reference for compact variable design, optional objects with defaults, and map variables.
+    *   [project](../../modules/project): Best reference for logical entity encapsulation (IAM, log sinks, Shared VPC configurations), complex locals transformation, and context-based interpolation.
```
juliocc, line 121: ADRs are also a good source of knowledge

skills/contributing/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+
+5.  **Maintain Code Consistency**:
+    *   Always keep variables in `variables.tf` and outputs in `outputs.tf` in strict alphabetical order.
```
juliocc: for larger modules, variables and outputs can be grouped logically (e.g. `variables-logging.tf`)

skills/contributing/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+    *   Limit line length to 79 characters (relaxed for long attribute names and descriptions).
+    *   **JSON Schema & Factory Alignment**: When a change is made to a module that implements a factory (e.g., `cloud-workstations`) or is the base for one (e.g., `net-vpc-factory` to `2-networking`), the schemas MUST be updated (including their `.schema.md` documentation files), and the factory code should be updated to mirror the changed variable surface. Ensure you regenerate the schema documentation by running:
+        ```bash
+        .venv/bin/python3 tools/schema_docs.py
+        ```
```
juliocc, line 137: also update `tools/duplicate_diff.py` when appropriate.

skills/contributing/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+    ```bash
+    source .venv/bin/activate
+    mkdir -p /tmp/tfcache
+    TF_PLUGIN_CACHE_DIR=/tmp/tfcache pytest tests/modules/<module-name>
+    ```
+
+4.  **Regenerate Test Inventories**: If module-level tests (`tftest.yaml`) or README example inventories fail due to intentional plan output changes, regenerate them automatically using `generate_plan_summary.py` (ensure you activate `.venv` first):
+    ```bash
+    # For module-level tftest.yaml inventories:
+    source .venv/bin/activate
```
juliocc: I'd encourage using `uv run`. All these scripts have deps liste inline

skills/contributing/SKILL.md
sruffilli: Thought about it, risk is not being aligned with the published version

skills/contributing/SKILL.md
sruffilli: Thought about referencing them, but that would force us to keep the skill in sync. I might generically point to their existance, but wouldn't trust the agent to go read them as needed. wdyt?

skills/contributing/SKILL.md
sruffilli: ++

skills/contributing/SKILL.md
sruffilli: ++

skills/contributing/SKILL.md
sruffilli: <3

ludoo:
*Agent-assisted review — claims below were verified against the tree (`tools/`, `.github/workflows/`, `default-versions.tf`, sibling skills) rather than just the diff.*

## Verdict

The skill's **content** is well-researched and clearly battle-tested, but its **process design** has a handful of defects that matter for a workflow meant to be followed literally by non-judging executors — most notably: the flow has no commit/push step, the inventory-regeneration guidance is circular, the live E2E step runs before cheap static review (and looks mandatory in the diagram), and the git workflow silently assumes maintainer push access. None of these require restructuring; combined with @juliocc's comments, this is one focused revision away from a genuinely strong skill.

## What it gets right

- **Two-mode entry** is the correct decomposition — issue-driven and proactive contribution genuinely differ only in the triage prefix, and the skill models that correctly.
- **Step 1.4's warning** — don't rely on the issue's proposed solution, fetch the full provider schema, make deliberate inclusion choices — is excellent agent hygiene against anchoring on user-supplied partial specs. The "don't miss arguments by omission" framing is subtle and right.
- **The review → fix → retest loop** correctly models iteration instead of pretending review is terminal.
- **Policy Troubleshooter verification with numeric project numbers** encodes non-obvious operational knowledge (IAM conditions evaluate `resource.name` against project *numbers*) that an agent could not derive and would silently get wrong.
- **The backtick/`--body` pitfall** and body-file approach reflect a real, common agent failure mode.
- Golden paths (`compute-vm`, `project`) match actual repo exemplar quality; the `gcs` `storage_class` snippet is quoted accurately; both `generate_plan_summary.py` invocations match the real CLI; the PR-title guidance matches CONTRIBUTING.md.

---

## 1. Architecture

**1a. Step 4 duplicates an existing repo subsystem — with conflicting rules.** The repo already has an automated PR review pipeline (`.github/workflows/pr-review.yml` → `tools/pr_review.py`, label-triggered), which posts comments headed `### Automated PR Review 🤖`. Step 4 defines a second, divergent one: same header (agent output becomes indistinguishable from the CI bot's), contradictory style (`pr_review.py`'s prompt says *"Do not use emojis"*; the template mandates ✅/⚠️/❗/💡), and two review-criteria lists that will drift. Suggest either referencing the canonical rubric or explicitly framing Step 4 as a *pre-flight self-check* with a different header. The "CRITICAL TESTING RULE" (inventories must change when resources change) is valuable and arguably belongs in the shared guidelines both consume.

**1b. Self-review by the authoring agent is structurally weak.** The same context that wrote the code reviews it. It has value as a forced checklist pass, but the ~25-line output template optimizes presentation over detection; a terse checklist would catch as much at lower context cost.

**1c. Monolithic structure vs. sibling conventions.** `fabric-builder` and `fast/prerequisites` use progressive disclosure (`references/` folders); this skill loads 229 lines whole. Step 3b (~40 lines) and the review template are ideal `references/` candidates. The frontmatter description also lacks the "Use when…" trigger clause both siblings have, which affects when agents actually activate the skill (and it differs slightly from the `skills/README.md` entry, which adds "and Policy Troubleshooter").

## 2. Flow logic

**2a. There is no commit or push step — anywhere.** The flow is develop → test → E2E → review → `gh pr create`. Nothing tells the agent to commit or push; atomic-commit conventions go unenforced because the concept doesn't appear. An agent following the skill literally cannot complete it.

**2b. The only human gate is at the very end.** Step 5 is "(After User Approval)", but that's the first mandated checkpoint — after design, code, live cloud deployment with `-auto-approve`, and self-review. A design/scope confirmation gate belongs between Steps 1 and 2. `fast/prerequisites` in the same tree is aggressively human-in-the-loop; this skill is fire-and-forget until the finish line.

**2c. Triage has no rejection path.** Step 1.3 evaluates fit and scope, but the flowchart is `A --> B` unconditionally. If the issue fails the fit test there's no defined action (comment on issue? propose closing? escalate?) — the evaluation is performative.

**2d. Mode B contradicts its own entry condition.** Mode B covers "preparing an existing branch", yet Step 2.2 unconditionally runs `git checkout -b`. There's also no "sync master first" or rebase-before-submit guidance in either mode.

**2e. Expensive validation before cheap validation.** Live E2E (Step 3b) runs before static self-review (Step 4); the `E --> C` loop can re-deploy to cloud each iteration. Order should be lint → test → review → E2E once the diff is stable. Compounding: the mermaid places 3b on the mandatory path (`C --> G --> D`) while the heading says "Optional / Recommended" — an agent following the diagram deploys to GCP by default. (The diagram also drifts numerically: nodes "5. Align & Fix" / "6. Submit" vs. body Steps 4.3 / 5.)

**2f. Inventory regeneration launders bugs into the baseline.** Step 3.4: on failure "due to intentional plan output changes", regenerate with `--save`. The agent itself judges intentionality, and `--save` overwrites the assertion baseline — after which the test passes by construction, and Step 4's testing rule (inventories changed? yes) passes too. Circular validation. Suggest requiring: *diff the regenerated inventory against the previous version, verify every changed line maps to the intended feature, and report the inventory diff to the user.*

## 3. Assumptions

| Assumption | Reality | Severity |
| :--- | :--- | :--- |
| Contributor has push access (`checkout -b` on origin + `gh pr create`) | CONTRIBUTING.md's default path for contributors is a fork; direct branches are for regulars by request. The fork flow is absent | Medium |
| Provider docs from the provider repo's `main` branch are authoritative | Repo pins `google >= 7.40.0, < 8.0.0`; docs from `main` may describe unreleased attributes the pinned provider rejects at plan time. Fetch the tag matching the pinned version | Medium |
| `.venv` exists and is populated | Documented path is `uv run` (PEP 723 inline deps); nothing creates `.venv` (already flagged by @juliocc — confirmed: bare `python3 tools/generate_plan_summary.py` fails on missing `click`) | Medium |
| `gcloud` is authed with perms + billing on the sandbox project | Asks for the project ID (good) but never verifies auth or shows a plan before `apply -auto-approve` | Medium |
| PII appears only in the PR body | E2E leaves real project IDs in `scratch/`, which is **not gitignored and doesn't exist**; sanitization is scoped only to Step 5. Commits and sandbox files are uncovered | Medium |
| Agent can self-classify Mode A vs B | No dispatch question for ambiguous entries; cheap fix: "if unclear, ask" | Low |
| Agent has `read_url_content` | Gemini-CLI-specific tool name in an otherwise agent-agnostic skill | Low |
| Issue lifecycle ends at PR creation | No label guidance; worth one sentence that changelog is label-driven so agents don't touch `CHANGELOG.md` | Low |

## 4. Accuracy nits

- **Step 5's heredoc contradicts the repo's own agent rules.** `cat << 'EOF' > /tmp/pr-body.txt` vs. GEMINI.md's *"CRITICAL: NEVER use shell redirection… to create files"*. An agent loaded with both gets conflicting critical instructions — suggest "create the body file with your file-writing tool, then `gh pr create --body-file`".
- **`scratch/` should be gitignored in this PR** (or the workaround should target `/tmp`), since both the provider-doc `curl` and the E2E sandbox write inside the repo.
- **`<module-name>` vs `<module_name>`**: test dirs use underscores (`tests/modules/net_vpn_ha/`), module dirs hyphens; the skill mixes both placeholders, so `pytest tests/modules/<module-name>` with a hyphenated name silently collects nothing.
- @juliocc's `uv run`, `optional()`-without-defaults, ADRs, and grouped-variables comments all check out and should be applied; small catch when applying: the file is `tools/duplicate-diff.py` (hyphen), not `duplicate_diff.py`.

skills/contributing/SKILL.md
juliocc: I'd at least mention the directory and let the agent decide what to use. There's good stuff in there, like the IAM ADR explaining the full iam* interface.

ludoo (approved):

PR #4109, opened 2026-08-25
Fix unknown IAM principals when organization logging settings are created

sruffilli:
Adding an org-level log sink to `0-org-setup` can produce dozens of spurious `update in-place` diffs on folder/project IAM bindings and PAM entitlements, with every existing principal rendered as `(known after apply)`.

### Root cause

`org_logging_identities` in `fast/stages/0-org-setup/organization.tf` gated its map keys on computed attributes:

```hcl
org_logging_identities = merge(
  module.organization[0].logging_identities.kms == null ? {} : {...},
  module.organization[0].logging_identities.logging == null ? {} : {...}
)
```

`logging_identities.logging` is `google_logging_organization_settings.default[0].logging_service_account_id`, which is unknown while that resource is being created. The ternary **condition** is therefore unknown, so the ternary returns a *wholly unknown map* — not a map containing unknown values.

That map is merged into the `iam_principals` context passed to `module.factory` (`factory.tf`), so every `lookup(local.ctx.iam_principals, m, m)` across `modules/project-factory` (folder IAM, project IAM, bucket/KMS/pubsub/log-bucket IAM, PAM entitlements) returns unknown.

The trigger is the organization `logging:` block going from **absent to present** — `logging_settings` and `logging.sinks` are both driven off that one key, so `count` on the settings resource flips 0→1. It does not depend on which sink is defined. Orgs that have always had the OOTB sink never see it, which makes it easy to miss.

Minimal repro of the mechanism, with an *unrelated* principal lookup:

```
+ buggy_lookup_unrelated_principal = (known after apply)
+ fixed_lookup_unrelated_principal = "group:admins@example.com"
```

### Fix

Derive the map keys from the yaml configuration, which is known at plan time:
- `organization/logging/sinks` — gated on presence of the `logging:` block
- `organization/logging/kms` — gated on `logging.kms_key_name` (schema-backed in `schemas/organization.schema.json`)

Also drop `if v != null` from the `iam_principals` comprehension in `factory.tf`; values are non-null by construction now, and that filter would make the map unknown again.

### Effect

Stage-0 plan with the `customizations` dataset (which defines org-level sinks):

| | resolved principals | unknown |
|---|---|---|
| before | 22 | 8 |
| after | 30 | 0 |

The 8 that flip are `google_project_iam_binding.authoritative` and `google_storage_bucket_iam_binding.authoritative` entries — the same class users report.

### Tests

Added `members` assertions to `tests/fast/stages/s0_org_setup/customizations.yaml`. These fail before the change:

```
AssertionError: ...authoritative["roles/cloudbuild.builds.editor"].members
is not a valid address in the plan
```

Note the inventory previously had *no* `members` key on those bindings at all — the bug was silently baked into the fixtures.

Full stage-0 suite passes (`hardened`, `simple`, `starter-gcd`, `customizations`, `caa`).

### Reviewer note

The KMS identity is now gated on `kms_key_name` rather than on the computed `kms_service_account_id`. The provider docs do not state whether that attribute is populated when no CMEK key is configured, so this preserves the *intent* of the original null check but is a slight behaviour change if the provider returns it unconditionally. No in-repo dataset sets `logging.kms_key_name`, so it is uncovered either way — worth a second opinion.

### Workaround for affected users

Until this lands, importing the settings resource makes the value known at plan time and removes the churn:

```hcl
import {
  to = module.organization[0].google_logging_organization_settings.default[0]
  id = "organizations/ORG_ID/settings"
}
```

---
**Checklist**

### DIFF
--- a/fast/stages/0-org-setup/organization.tf
+++ b/fast/stages/0-org-setup/organization.tf
@@ -15,6 +15,7 @@
  */
 
 locals {
+  _org_logging_settings = lookup(local.organization, "logging", null)
   ctx_condition_vars = {
     custom_roles = merge(
       local.ctx.custom_roles,
@@ -56,14 +57,6 @@ locals {
   org_access_levels = {
     for k, v in module.organization[0].access_levels : k => v.id
   }
-  # private local: logging settings as defined in yaml, used to derive the
-  # shape of the identities map below from plan-time known values
-  _org_logging_settings = lookup(local.organization, "logging", null)
-  # keys must only depend on the yaml configuration: gating them on the
-  # computed identity attributes makes the ternary condition itself unknown
-  # while the logging settings resource is being created, which turns the
-  # whole map unknown and poisons every principal interpolation downstream
-  # (folder/project IAM, PAM entitlements, ...) in the project factory
   org_logging_identities = merge(
     try(local._org_logging_settings.kms_key_name, null) == null ? {} : {
       "organization/logging/kms" = module.organization[0].logging_identities.kms
--- a/fast/stages/0-org-setup/organization.tf
+++ b/fast/stages/0-org-setup/organization.tf
@@ -56,11 +56,19 @@ locals {
   org_access_levels = {
     for k, v in module.organization[0].access_levels : k => v.id
   }
+  # private local: logging settings as defined in yaml, used to derive the
+  # shape of the identities map below from plan-time known values
+  _org_logging_settings = lookup(local.organization, "logging", null)
+  # keys must only depend on the yaml configuration: gating them on the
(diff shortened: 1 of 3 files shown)


fast/stages/0-org-setup/organization.tf
```diff
@@ -56,11 +56,19 @@ locals {
   org_access_levels = {
     for k, v in module.organization[0].access_levels : k => v.id
   }
+  # private local: logging settings as defined in yaml, used to derive the
+  # shape of the identities map below from plan-time known values
+  _org_logging_settings = lookup(local.organization, "logging", null)
```
ludoo: underscored local goes on top

juliocc (approved):

PR #4122, opened 2026-08-27
Scopes-only Import Manifest Grammar With Per-scope Types

sruffilli:
## Summary

Replaces the fabric-importer manifest grammar with a **scopes-only** form: `scopes:` is required and **every scope carries its own `types:` list**. The retired top-level `scope:` / `types:` grammar is refused with migration instructions.

This makes per-scope type declarations first-class — different resource types under different subtrees (org governance at org/folder level, the connectivity stack under the networking folder, KMS under security, project IAM only in named workload projects) — which the previous global type list could not express. Choosing a single grammar (rather than adding per-scope types alongside the global list) makes three whole classes of silent-denominator failures *unrepresentable* instead of merely validated: there is no inheritance to mix, no global list to go stale, and no all-or-nothing rule to enforce.

## Fail-closed validation (exit 1, before any API call)

- a scope without a `types:` list, or with `types: []`, is refused — a scope that collects nothing may not say so quietly;
- a type whose `levels` cannot intersect its scope's `levels` is refused as a dead declaration (entries listing `unknown` are exempt);
- a duplicate `type:` within one scope's list is refused, with the scope named;
- the retired grammar is refused with a worked migration snippet.

## Changes

**`scripts/inventory.py`** (frozen — reviewed change): per-scope resolution of levels, leaf-IAM opt-ins and native enumerators; entries carry a merged sorted `scopes` attribution (never last-wins under overlapping scopes); `_meta.scopes[]` records per-scope `declared_types` / `zero_yield_types`; new stderr warning for a type that yields zero in one scope while non-zero elsewhere — the shape the aggregate table cannot show.

**`scripts/manifest_init.py` / `scripts/manifest_from_state.py`**: emit the new grammar; the state-driven generator filters each scope's list with exactly the validator's rule, so it can never emit a manifest `inventory.py` refuses.

**Examples**: all migrated; new `examples/import-manifest.multi-domain.yaml` is the worked four-scope reference; `org-foundation.yaml` rewritten as the single-scope reference with comment-block trims.

**Tests**: 211 → 225. Fixture migration plus `TestPerScopeTypes`: legacy rejection, missing/empty/dead/duplicate lists, per-scope sweep isolation (CAI calls asserted per root), per-scope `enumerate:` and `iam: true` isolation, scope-attribution merging, per-scope yield tables, and shipped examples now validated through the real parser.

**Docs**: SKILL.md step-0 rules and step-5 per-scope reporting duties; README schema section, new *Different depth per subtree* section, `inventory.py` exit codes (previously undocumented); cookbook emission section rewrite plus an explicit emission-vs-types contrast (emission inherits defaults because it is denominator-neutral; types never inherit because they decide the denominator), new *Per-scope type declaration* subsection; operating-contract convergence bounded **per scope**; inferring-manifests examples match real generator output. Also fixes the pre-existing cookbook contradiction on numeric-vs-alphanumeric project ids in include/exclude.

## Validation

- `pytest tests/test_scripts.py`: **225 passed**
- `yamllint` on all examples: clean
- `check_boilerplate.py` on all touched files: clean
- `address_map.py` + AddressMap sync tests: clean
- `yapf` (repo style) applied to all touched Python
- live smoke: `inventory.py collect` on the multi-domain example validates and proceeds to enumeration (fails only on placeholder-org credentials, as expected)

### DIFF
--- a/skills/fabric-importer/scripts/inventory.py
+++ b/skills/fabric-importer/scripts/inventory.py
@@ -1259,9 +1259,10 @@ def _verify_split_parity(assets, declared_types, sibling_map, scope,
   })
 
 
-def validate_manifest_types(types):
+def validate_manifest_types(types, where='manifest'):
   """Fails closed on manifest mistakes that silently shrink the
-  denominator.
+  denominator. Called once per scope: each scope carries its own
+  `types:` list, and `where` names the scope in every message.
 
   - An unrecognised `levels` value (e.g. `org` for `organization`) used
     to drop every entry of that type with no output at all — the exact
@@ -1274,16 +1275,16 @@ def validate_manifest_types(types):
   for t in types:
     tt = t.get('type')
     if not tt:
-      raise SystemExit('manifest type entry without a `type:` field')
+      raise SystemExit(f'{where} type entry without a `type:` field')
     if tt in seen:
       raise SystemExit(
-          f'manifest declares type {tt!r} more than once; duplicate '
+          f'{where} declares type {tt!r} more than once; duplicate '
           'entries silently narrow the denominator (last one wins) — '
           'merge them into a single entry')
     seen.add(tt)
     if tt == PAM_GRANT_TYPE:
       raise SystemExit(
-          f'manifest type {tt!r} is machine-managed runtime state: active '
+          f'{where} type {tt!r} is machine-managed runtime state: active '
           'PAM grants are enumerated automatically whenever IAM is '
           'collected, and their temporary bindings are stripped from the '
           'denominator (never mapped, never waived — see '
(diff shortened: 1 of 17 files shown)


skills/fabric-importer/scripts/inventory.py
```diff
@@ -1274,45 +1275,61 @@ def validate_manifest_types(types):
...
+  `types:` grammar is refused with migration instructions rather than
+  a bare unknown-key error — silently accepting half of it would
+  change what the denominator means.
+  """
+  legacy = [k for k in ('scope', 'types') if k in manifest]
+  if legacy:
+    raise SystemExit(
+        f'manifest declares retired top-level key(s) {legacy}. The '
+        'grammar is scopes-only and every scope carries its own '
+        "'types:' list:\n"
```
juliocc, line 1321: :(

why not support both? Or give me an easy migration

juliocc (approved):

skills/fabric-importer/scripts/inventory.py
sruffilli: ++ - added

PR #4128, opened 2026-08-29
Add IAP tunnel IAM to compute-vm module

singhal0306:
compute/osAdminLogin or compute/osLogin roles doesn't work alone for internal only VM's, Thats where they need iap-tunnel-access

---
**Checklist**


sruffilli:
There's a number of structural issues to tackle before considering merging, but besides that we should discuss whether this feature should be in scope for the module. 
@ludoo @juliocc wdyt?

singhal0306:
> There's a number of structural issues to tackle before considering merging, but besides that we should discuss whether this feature should be in scope for the module. @ludoo @juliocc wdyt?

These IAP roles have one to one co-relation with instances.

ludoo:
Let's have a quick discussion, I did not look at the shape of the code yet, but while a bit on the boundary vs what our modules typically do, I see the use case and rationale for having this here. My main concern is getting the variable naming right. @sruffilli I don't get what you were saying above. :)

And we should revisit the iam interface in this module, I don't remember why it only has `iam` and not the full interface, it's been too long. I'll think about it for a bit and come back tomorrow with a proposal. Thanks for flagging this and taking to the time to send a PR.

singhal0306:
> Let's have a quick discussion, I did not look at the shape of the code yet, but while a bit on the boundary vs what our modules typically do, I see the use case and rationale for having this here. My main concern is getting the variable naming right. @sruffilli I don't get what you were saying above. :)
> 
> And we should revisit the iam interface in this module, I don't remember why it only has `iam` and not the full interface, it's been too long. I'll think about it for a bit and come back tomorrow with a proposal. Thanks for flagging this and taking to the time to send a PR.

I appreciate that @ludoo

singhal0306:
@ludoo Is there any update?

juliocc:
added the full iam interface.

juliocc (approved):

ludoo (approved):

PR #4139, opened 2026-09-03
fix: explicitly configure disabled IAP

singhal0306:
This PR updates the IAP configuration in the backend service module to explicitly render the `iap` block with `enabled = false` when IAP is not configured.

### Changes
- Updated the dynamic `iap` block to always render when `iap_config` is `null`.
- Explicitly sets `enabled = false` when IAP is not configured.
- Prevents Terraform from continuously detecting the provider-returned `enabled = false` as a removal of the `iap` block.

### Impact
This resolves the recurring Terraform plan diff:

```text
- iap {
    - enabled = false -> null
  }
```
---
**Checklist**

### DIFF
--- a/modules/net-lb-app-ext-regional/backend-service.tf
+++ b/modules/net-lb-app-ext-regional/backend-service.tf
@@ -194,14 +194,11 @@ resource "google_compute_region_backend_service" "default" {
     }
   }
 
-  dynamic "iap" {
-    for_each = each.value.iap_config == null ? [] : [each.value.iap_config]
-    content {
-      enabled                     = true
-      oauth2_client_id            = try(iap.value.oauth2_client_id, null)
-      oauth2_client_secret        = try(iap.value.oauth2_client_secret, null)
-      oauth2_client_secret_sha256 = try(iap.value.oauth2_client_secret_sha256, null)
-    }
+  iap {
+    enabled                     = each.value.iap_config != null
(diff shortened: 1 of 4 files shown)


modules/net-lb-app-ext-regional/backend-service.tf
```diff
@@ -195,9 +195,9 @@ resource "google_compute_region_backend_service" "default" {
   }
 
   dynamic "iap" {
-    for_each = each.value.iap_config == null ? [] : [each.value.iap_config]
+    for_each = [try(each.value.iap_config, null)]
     content {
-      enabled                     = true
+      enabled                     = try(iap.value != null, false)
```
juliocc: Something is wrong here

```hcl
[try(each.value.iap_config, null)]
```

Always produces a list with one element. If you're going that way, why even bother with the `dynamic` block?

singhal0306: Yeah, i overlooked that..

It was just, i was solving this problem..
     - iap {
          - enabled                         = false -> null
          - oauth2_client_id_wo             = (write-only attribute) -> null
          - oauth2_client_secret_sha256     = (sensitive value) -> null
          - oauth2_client_secret_wo         = (write-only attribute) -> null
            # (4 unchanged attributes hidden)
        }

juliocc (approved):

juliocc (approved):

PR #4142, opened 2026-09-04
Add net-cloud-armor module

emailabhigarg:
This PR introduces the `modules/net-cloud-armor` module for provisioning and managing Google Cloud Armor security policies.

### Overview & Key Features

- **Preconfigured WAF & Threat Intelligence**:
  - Direct rule convenience helpers for OWASP Core Rule Set (`sqli-v33-stable`, `xss-v33-stable`, `rce-v33-stable`, etc.) via `preconfigured_waf`.
  - Native support for Google Threat Intelligence feeds (`iplist-tor-exit-nodes`, `iplist-known-malicious-ips`, `iplist-vpn-providers`, etc.) via `threat_intel_feed`.
- **IP & Custom Expression Rules**:
  - Source IP CIDR allow/deny matching (such as GCP load balancer health check probe allowlists `130.211.0.0/22`, `35.191.0.0/16`).
  - Custom CEL expression matching (such as geo-blocking via `origin.region_code != 'US'`).
- **Rate Limiting & Throttling**:
  - IP-based and session-based (`HTTP_COOKIE`, `HTTP_HEADER`) rate limiting with configurable thresholds, conform/exceed actions (`throttle`, `deny(429)`), and ban duration.
- **Layer 7 DDoS Defense & Advanced Inspection**:
  - Adaptive Protection configuration for Layer 7 DDoS defense.
  - JSON body parsing (`STANDARD`) and custom content-type inspection.
- **Rules Factory**:
  - Data-driven configuration allowing users to inject rules from YAML files or folders via `factories_config`.
  - JSON Schema (Draft-07) defined in `schemas/rules.schema.json` with companion documentation generated by `tools/schema_docs.py` in `schemas/rules.schema.md`.
- **Dual-Engine CI Parity**:
  - Tested and verified under both Terraform and OpenTofu.

---

**Checklist**

I acknowledge that I have:
- [X] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [X] Ran `terraform fmt` on all modified files
- [X] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [X] Made sure all relevant tests pass (all 11 CFF verification steps passed)

PR #4144, opened 2026-09-07
Add net-cloud-armor module

sruffilli:
# modules/net-cloud-armor

This PR adds the `modules/net-cloud-armor` module. It provides a single interface to manage both global and regional [Cloud Armor security policies](https://cloud.google.com/armor/docs/security-policy-overview) and their rules.

This builds supersedes the initial work in #4142 (credit to @emailabhigarg), extending coverage to the entire Cloud Armor feature set.

### Key Features & Design

- **Unified global and regional support:** Set a `region` to switch between regional and global scope (similar to `net-firewall-policy`). Supports all policy types:
  - `CLOUD_ARMOR` (backend security)
  - `CLOUD_ARMOR_EDGE` (Cloud CDN edge security)
  - `CLOUD_ARMOR_NETWORK` (network edge, including `network_match`, `user_defined_fields`, and `ddos_protection`)
  - `CLOUD_ARMOR_INTERNAL_SERVICE` (internal load balancers)
- **Individual rule resources:** Uses standalone `google_compute_[region_]security_policy_rule` resources with `for_each` keyed by rule name. This prevents unexpected resource recreation, keeps Terraform state addresses stable, and allows clean additions, updates, or deletions. The default rule (priority 2147483647) is configured separately via `default_rule_config` and cannot be defined inside `rules`.
- **Early validation at plan time:** Checks whether features match the chosen scope and policy type before apply time (e.g., catching that redirects, custom headers, and Adaptive Protection only work on global policies, or that `network_match` requires `CLOUD_ARMOR_NETWORK`). You get clear validation errors during `terraform plan` instead of API failures during `terraform apply`.
- **Comprehensive rule options:** Supports preconfigured WAF exclusions, rate limiting (including `enforce_on_key_configs` and redirect-on-exceed options), reCAPTCHA `expr_options`, request header actions, and redirects.
- **Rules factory:** Supports loading rules from YAML files via `factories_config.rules_file_path`, complete with JSON schema validation and generated documentation. Rules defined in YAML and rules passed via variables use the exact same format and normalization.
- **Included recipes:**
  - `recipe-glb-waf`: Global Application Load Balancer + Cloud Run + CDN backend bucket with both backend WAF and edge policies.
  - `recipe-regional-waf`: A single regional policy shared across an external and an internal regional Application Load Balancer.
  - Cloud Run invoker permissions can be configured using `invoker_members` (defaults to `["allUsers"]`) to work around domain-restricted sharing organization policies.

### Load Balancer Integration

Policies created with this module can be attached directly to the existing `security_policy` and `edge_security_policy` attributes in the Fabric load balancer modules. This PR also adds `security_policy` support to `net-lb-ext`.

---

### Testing

**Unit & Plan Tests:**
- 7 README examples backed by plan inventories (covering global, rate limiting, regional, network edge, edge, factory).
- Schema validation for factory YAML files.
- 2 recipe plan tests.
- 24 negative test cases tested manually to verify the plan-time validation rules.

**End-to-End Live Testing:**
All scenarios were deployed, tested, and cleanly destroyed in a sandbox GCP project:

| Scenario | Apply | API read-back | Behavioral | Mutations | Destroy |
|---|:---:|:---:|:---:|:---:|:---:|
| Global policy on global ALB (WAF, rate limiting, custom headers, preview mode, default rule) | ✓ | ✓ | ✓ | ✓ | ✓ |
| Edge policy on CDN backend bucket | ✓ | ✓ | ✓ | ✓ | ✓ |
| Regional policy on regional external ALB | ✓ | ✓ | ✓ | ✓ | ✓ |
| Same regional policy shared with an internal ALB | ✓ | ✓ | ✓ | ✓ | ✓ |
| Network edge policy (`CLOUD_ARMOR_NETWORK`) | ✓ | ✓ | skipped | ✓ | ✓ |
| Factory rules | ✓ | ✓ | ✓ | ✓ | ✓ |

- **Traffic & behavior tests:** Verified actual enforcement with `curl`: blocked SQL injection and XSS attacks (HTTP 403), rate-limit throttling (HTTP 429), IP allow/deny rules, and preview mode pass-through. Log entries were verified in Cloud Logging under `jsonPayload.enforcedSecurityPolicy` and `enforcedEdgeSecurityPolicy` to confirm correct rule matching and action attribution.
- **Lifecycle tests:** Confirmed rules could be modified, added, and removed concurrently without unexpected replacements or state churn. Verified that changes to the default rule apply in place, re-plans produce zero diffs, and `terraform destroy` tears down all resources cleanly without requiring manual detachment.

---

### Documentation correction (follow-up commit)

While running the E2E suite for the stacked PR #4145 in a second sandbox project, two README claims turned out to be wrong against the live API. Both are corrected in `Correct edge and network policy constraints in net-cloud-armor docs`:

1. **Edge policy expressions depend on the attachment target.** The README listed `origin.asn`, `origin.ip`, `origin.region_code`, `request.headers`, `request.method`, `request.path`, `request.query` and `request.scheme` as the evaluable attributes for `CLOUD_ARMOR_EDGE`. That is the Media CDN surface. A policy attached to a **backend bucket** of a global external ALB only accepts `origin.ip` (via `src_ip_ranges` or `inIpRange`) and `origin.region_code`. Everything else is rejected at *rule creation* — not at attach time — with:

   ```
   Invalid value for field 'resource'. Expression supported only for Media CDN edge policies.
   ```

   Probed attribute by attribute against the live API:

   | Expression | Rule create | Attach to backend bucket |
   |---|:---:|:---:|
   | `inIpRange(origin.ip, '203.0.113.0/24')` | accepted | accepted |
   | `origin.region_code == 'IT'` | accepted | accepted |
   | `request.path.startsWith('/x')` | rejected | n/a |
   | `request.headers['x-test'] == 'block'` | rejected | n/a |
   | `request.method == 'POST'` | rejected | n/a |
   | `origin.asn == 65000` | rejected | n/a |

   No plan-time validation was added for this: the module cannot see the attachment target, so a validation would wrongly reject legitimate Media CDN policies. The constraint is documented instead.

2. **`CLOUD_ARMOR_NETWORK` requires Enterprise to create, not just to define rules on.** The README said Enterprise was needed for *defining custom rules*. On a project below that tier the API rejects the policy resource itself:

   ```
   Invalid value for field 'resource.type': 'CLOUD_ARMOR_NETWORK'.
   Network Security Policies are not supported as part of the current Cloud Armor service tier.
   ```

   The wording now separates the two requirements: Enterprise to create the policy, and regional advanced network DDoS protection additionally to attach it.


modules/net-cloud-armor/README.md
```diff
@@ -0,0 +1,375 @@
+# Cloud Armor Security Policy
+
+This module provisions and configures [Cloud Armor security policies](https://cloud.google.com/armor/docs/security-policy-overview) along with their rules. It handles both global and regional policies through a unified interface, validating at plan time that all requested features are compatible with the selected policy scope and type.
+
+Attach the policy to load balancers using the `id` output, which can be passed to the `security_policy` (or `edge_security_policy`) attribute of backend services or buckets managed by the `net-lb-app-ext`, `net-lb-app-ext-regional` and `net-lb-app-int` modules. Cross-region internal Application Load Balancers, internal proxy Network Load Balancers, and internal passthrough Network Load Balancers do not support Cloud Armor, so their corresponding modules omit the attribute.
```
LucaPrete, line 5: I'm not 100% sure other LBs are not supported at all. For sure ext passthrough LB supports it.

modules/net-cloud-armor/README.md
```diff
@@ -0,0 +1,375 @@
...
+- [Files](#files)
+- [Variables](#variables)
+- [Outputs](#outputs)
+<!-- END TOC -->
+
+## Examples
+
+### Global backend policy
+
+Global backend security policy for the global external Application Load Balancer, featuring preconfigured WAF rules, rule exclusions, an IP allowlist, and Adaptive Protection.
```
LucaPrete, line 28: I'm not sure adaptive protection is in the standard cloud armor subscription, but rather only in the enterprise. It might be worth not activating it by default

sruffilli: Adaptive Protection is available on Cloud Armor Standard, with limited features. I agree that enabling it by default sets the wrong expectation for most users - defaulting this to false.

modules/net-cloud-armor/README.md
sruffilli: You're right - the problem here was that net-lb-ext module did not support security_policy. I added it in this PR.

LucaPrete (approved):

PR #4145, opened 2026-09-07
Add security policy context support to load balancer modules

sruffilli:
Follow-up to #4144 (stacked on its branch), completing the composition between Cloud Armor policies and the load balancer modules.

`modules/net-cloud-armor` can create the policies, but attaching one to a load balancer meant passing a fully qualified policy id around by hand. This PR makes the attachment resolvable through the standard `context` mechanism, so factories and FAST stages can reference policies symbolically.

### Changes

- Add a `security_policies` key to the `context` variable of `net-lb-app-ext`, `net-lb-app-ext-regional` and `net-lb-app-int`, resolving the existing `security_policy` (backend services) and `edge_security_policy` (backend buckets) attributes. A policy can now be referenced as `$security_policies:waf` instead of a full resource path.
- Resolve `zone` through `context.locations` in `net-lb-ext` unmanaged instance groups (`groups.tf`). This was the one field in that module still bypassing context.
- Document context interpolation in the `net-lb-ext` README, and the context-based attachment path in the `net-cloud-armor` README.

The `security_policy` attribute on `net-lb-ext` itself, and its `security_policies` context key, moved to #4144 — the base branch now carries them, so this PR no longer contains that work.

### Testing

**Unit & plan tests:** 36 module tests and 97 README example tests pass across the five affected modules. Context resolution is asserted in the plan inventories for all four load balancer modules (`tests/modules/net_lb_app_ext/examples/context.yaml`, `net_lb_app_ext_regional/context.yaml`, `net_lb_app_int/examples/context.yaml`, `net_lb_ext/context.yaml`), each checking that the symbolic reference resolves to the concrete policy path.

**End-to-end live testing:** deployed to a sandbox project, verified, and destroyed. One VPC with a proxy-only subnet, nginx backends behind unmanaged instance groups, a GCS backend bucket, and five Cloud Armor policies attached across four load balancers.

*Read-back* — every attachment confirmed against the service API with `gcloud compute backend-services describe` / `backend-buckets describe`, not just Terraform state:

| Module | Attribute | Policy type | Resolved and attached |
|---|---|---|:---:|
| `net-lb-app-ext` | `security_policy` | global `CLOUD_ARMOR` | ✓ |
| `net-lb-app-ext` | `edge_security_policy` (backend bucket) | global `CLOUD_ARMOR_EDGE` | ✓ |
| `net-lb-app-ext-regional` | `security_policy` | regional `CLOUD_ARMOR` | ✓ |
| `net-lb-app-int` | `security_policy` | regional `CLOUD_ARMOR` | ✓ |
| `net-lb-ext` | `security_policy` | regional `CLOUD_ARMOR_NETWORK` | see below |

*Behavioral* — traffic sent through each load balancer to confirm the policies actually enforce, rather than merely being attached:

| Path | Global ALB | Regional ALB | Internal ALB |
|---|:---:|:---:|:---:|
| `GET /` | 200 | 200 | 200 |
| `GET /blocked` (deny rule) | 403 | 403 | 403 |
| `GET /api/` (unmatched) | 200 | 200 | 200 |
| `GET /` with `x-armor-test: block` | 403 | — | — |

The internal ALB was tested from a client VM inside the VPC over IAP SSH.

For the backend bucket edge policy, the denied source range was flipped from a TEST-NET-3 placeholder to the tester's `/32` and the request re-sent: the response changed from 412 to 403 after roughly two minutes of propagation, while `GET /` through the backend service stayed 200. The 412 comes from an organization policy enforcing `storage.publicAccessPrevention` on the bucket; since Cloud Armor evaluates before the request reaches GCS, the 412 → 403 transition still isolates the edge policy cleanly and shows it enforcing independently of the backend service policy.

*`net-lb-ext` limitation* — the sandbox project sits below Cloud Armor Enterprise, so a `CLOUD_ARMOR_NETWORK` policy cannot be created there at all and the intended attach could not be exercised. To confirm the module still resolves and forwards the value, a deliberately wrong-type policy was attached; the API rejection echoed the fully resolved path before failing on the type check:

```
Error 400: Invalid value for field 'resource':
'{ "securityPolicy": "projects/my-project/regions/my-region/securityPolicies/waf-regional..." }'.
Security policy ... is not a network security policy.
```

So the context lookup demonstrably reaches the API; only the service tier blocks the real attachment. Plan-level resolution for this path is covered by `tests/modules/net_lb_ext/context.yaml`.

No breaking changes: `security_policies` is a new optional context key and every existing configuration keeps working unchanged.


LucaPrete (approved):

LucaPrete (commented):
It would be nice to document for passthrough LBs as well how to use context in the readme (and maybe test directly there?)

sruffilli:
> It would be nice to document for passthrough LBs as well how to use context in the readme (and maybe test directly there?)

Done! Do you also mind to PTAL at #4144 😬 ?

PR #4150, opened 2026-09-08
Forward dropped bucket and log bucket attributes in project-factory

### LINKED ISSUE #4147: [FR]: `project-factory`: bucket attributes including `public_access_prevention` are silently dropped
Buckets declared under a project's `buckets:` key are wired to `modules/gcs`, but `modules/project-factory/projects-buckets.tf` forwards a closed subset of that module's inputs. `public_access_prevention` is one of the omissions, so a bucket that has an explicit value for it cannot be represented in project YAML and cannot be imported without permanent drift.

**Evidence**

`projects-buckets.tf:17-64` builds the bucket list from a fixed set of`lookup()`s and `:66-116` passes exactly those to `../gcs`.
`public_access_prevention` appears nowhere under `modules/project-factory/`; it is declared at `modules/gcs/variables.tf:306` and applied at `modules/gcs/main.tf:73`.

Also not forwarded: `autoclass`, `cors`, `default_event_based_hold`, `enable_hierarchical_namespace`, `ip_filter`,  `notification_config`, `objects_to_upload`, `requester_pays`, `rpo`, `website`.

`$defs.bucket` in `modules/project-factory/schemas/project.schema.json` is `additionalProperties: false` and omits these too.

**Impact**

An existing bucket with an explicit `public_access_prevention` value cannot be imported into project-factory without perpetual plan drift on that attribute, forcing it to stay a standalone `modules/gcs` instance.

### Proposed solution

Add `public_access_prevention` (and preferably the other missing `gcs` inputs) to `local.projects_buckets` and the `module "buckets"` call in `projects-buckets.tf`, and to `$defs.bucket` in the project schema — remembering the five duplicated copies listed in `tools/duplicate-diff.py:102-106`.

### Additional context

_No response_

### LINKED ISSUE #4149: project-factory: log_buckets.description is documented and read, not used
### Affected components

- [x] Fabric modules
- [ ] FAST
- [ ] Other

### Summary

The project schema documents description on a log bucket and projects-log-buckets.tf reads it, but the module "log-buckets" block never passes it on. A description set in project YAML validates, produces no error, and has no effect.

### Evidence

modules/project-factory/projects-log-buckets.tf:24 computes the value:

```hcl
  description  = lookup(opts, "description", "Terraform-managed.")
```

The module "log-buckets" block (:42-67) forwards parent, name, location, kms_key_name, context, retention and log_analytics — not description.

modules/logging-bucket supports it: variables.tf:36-40 (default = null), applied to all four parent-type resources in main.tf (:55, :72, :90, :99).

The attribute is part of the documented input contract: $defs.log_bucket in modules/project-factory/schemas/project.schema.json lists description. This is what distinguishes it from a plain feature gap — the schema advertises it, the glue reads it, and the value is dropped.

### Release version

latest head

### Terraform version

1.15

### Reproduction

```yaml
  # data/projects/logs.yaml
  name: logs
  project_reuse:
    use_data_source: false
    attributes: {name: logs, number: 1234567890, services_enabled: []}
  log_buckets:
    audit-logs:
      location: global
      retention: 365
      description: Centralized log bucket for app logs
```

terraform plan (provider hashicorp/google 7.x, as pinned by default-versions.tf):

```
  + resource "google_logging_project_bucket_config" "bucket" {
      + bucket_id      = "audit-logs"
      + description    = (known after apply)
      + location       = "global"
      + project        = "logs"
      + retention_days = 365
    }
```

### Impact

- Log bucket descriptions cannot be set or changed through project YAML.
- A description in a project file is accepted silently, so the YAML misrepresents the deployed configuration.

This does not cause drift on adopted/imported buckets: description is Optional + Computed in the provider, so a null config value leaves the live description untouched. The problem is purely that the factory cannot manage the value.

### Expected behavior

-

### Actual behavior

-

### Suggested fix

In projects-log-buckets.tf:

```hcl
  # local
  description  = lookup(opts, "description", null)

  # module "log-buckets"
  description  = each.value.description
```

The default must change from "Terraform-managed." to null. Forwarding the current default would write a description onto every existing factory-managed log bucket that has none today, producing an unexpected in-place update on the next apply.

A test inventory should assert description on the planned google_logging_project_bucket_config so the attribute cannot be dropped again unnoticed.

sruffilli:
Fixes #4147, fixes #4149.

### Context & Problem

In `modules/project-factory`, the input lists passed to `modules/gcs` and `modules/logging-bucket`have drifted from the interfaces of the wrapped modules. Several attributes advertised in the project schema and validated by the glue code were being silently discarded instead of forwarded to the underlying resources.

---

### Root Cause: Discarded Attributes

* **GCS Buckets:**
  * **Dropped attributes:** `autoclass`, `cors`, `default_event_based_hold`, `enable_hierarchical_namespace`, `ip_filter`, `notification_config`, `public_access_prevention`, `requester_pays`, `rpo`, and `website`.
  * **Impact:** Existing buckets with an explicit `public_access_prevention` could not be represented in project YAML, causing permanent plan drift upon import.
* **Log Buckets:**
  * **Dropped `description`:** Evaluated at `projects-log-buckets.tf:24` but never passed to the module. Setting a description in project YAML validated without errors but had no effect.
  * **Other missing attributes:** `locked`, `tag_bindings`, and `views`.

---

#### 1. Input Forwarding & Variable Parity
* **Forward missing inputs:** Updated `projects-buckets.tf` and `projects-log-buckets.tf` to pass all missing arguments to the wrapped modules.
* **Variable parity:** Added these attributes to the typed `projects` variable so they are accessible directly without using factory YAML files.
* **Log bucket context:** Added `tag_values` and `tag_vars` to the log bucket context merge so newly added `tag_bindings` resolve symbolically.
* **Safe description default:** Changed log bucket `description` default from `"Terraform-managed."` to `null`. Forwarding the previous default would have forced a description onto every existing factory-managed log bucket that currently lacks one.

#### 2. Schema Adjustments
* **Bucket schema (`$defs.bucket`):** Added `iam_by_principals` (the glue code already forwarded it, but schema validation was rejecting it via `additionalProperties: false`).
* **Automation buckets:** Kept their intentionally narrow surface, adding only `public_access_prevention` (also added to the folder schema for folder-level automation buckets).

---

#### Plan-Level Testing
* The `Tests` README fixture now exercises every newly supported attribute.
* The regenerated test inventory asserts these attributes on planned resources.
* The inventory diff is strictly additive (no pre-existing resources were modified).
* Full test suites pass: `project-factory` (19 tests) and `tests/fast` (25 tests).

#### Live End-to-End (E2E) Testing
Resource configuration was verified via direct GCP service API read-back (bypassing Terraform state):
* **GCS read-back verified:** `publicAccessPrevention: enforced`, `requesterPays: true`, `rpo: DEFAULT`, `defaultEventBasedHold: true`, `hierarchicalNamespace.enabled: false`, plus nested structures for `ipFilter`, `cors`, `website`, and `customPlacementConfig`.
* **Logging read-back verified:** `gcloud logging buckets describe` returned `description: E2E log bucket description.`, with log views and `retentionDays: 365`.

**API Validation Feedback Loop:**
Two apply failures confirmed attributes were successfully delivered to GCP APIs:
1. GCP rejected `rpo: DEFAULT` on a `REGION` bucket (fixed in fixture by switching to a dual-region bucket, making the README example apply-valid rather than merely plan-valid).
2. `ip_filter` locked out the operator IP during read-back (recovered via temporary `storage.buckets.exemptFromIpFilter` custom role, which has since been removed).

> *Note:* Behavioral verification beyond API read-back was not performed. All sandbox test resources were destroyed.

---

### Breaking Changes

```upgrade-note
`modules/project-factory`: the `description` attribute is removed from bucket definitions (`buckets` and `automation.bucket`, in both the project and folder schemas). It never had any effect, because `google_storage_bucket` has no description field, so the schema was advertising an attribute the GCS API cannot store.

No plan or apply breaks and no resource changes: the factory does no runtime schema validation, so a leftover `description` key in project YAML is simply ignored (verified against a real plan). The key is only rejected by schema validation in editors and in CI, so drop it when convenient. Eighteen in-repo datasets set one and are cleaned up in this PR.

Separately, log bucket `description` is now actually applied. A `log_buckets` entry that already sets one will show a one-time in-place update adding it to `google_logging_project_bucket_config`.
```

### DIFF
--- a/fast/project-templates/data-mongodb/project.yaml
+++ b/fast/project-templates/data-mongodb/project.yaml
@@ -38,7 +38,6 @@ automation:
     rw:
       description: Read/write automation service account for MongoDB.
   bucket:
-    description: Terraform state bucket for apt registries.
     # this reuses the existing stage state bucket and creates a folder in it
     name: iac-stage-state
     create: false
--- a/fast/project-templates/managed-kafka/project.yaml
+++ b/fast/project-templates/managed-kafka/project.yaml
@@ -40,7 +40,6 @@ automation:
     rw:
       description: Read/write automation service account for Managed Kafka.
   bucket:
-    description: Terraform state bucket for Managed Kafka.
     # this reuses the existing stage state bucket and creates a folder in it
     name: iac-stage-state
     create: false
--- a/fast/project-templates/secops-anonymization-pipeline/project.yaml
+++ b/fast/project-templates/secops-anonymization-pipeline/project.yaml
@@ -37,7 +37,6 @@ automation:
     rw:
       description: Read/write automation service account for secops pipeline.
   bucket:
-    description: Terraform state bucket for apt registries.
     # this reuses the existing stage state bucket and creates a folder in it
     name: iac-stage-state
     create: false
--- a/fast/stages/0-org-setup/datasets/classic-gcd/projects/core/iac-0.yaml
+++ b/fast/stages/0-org-setup/datasets/classic-gcd/projects/core/iac-0.yaml
@@ -94,7 +94,6 @@ data_access_logs:
 buckets:
   # Terraform state bucket for this stage
   iac-org-state:
-    description: Terraform state for the org-level automation.
     versioning: true
     iam:
       roles/storage.admin:
@@ -103,7 +102,6 @@ buckets:
         - $iam_principals:service_accounts/iac-0/iac-org-ro
   # Terraform state bucket for additional FAST stages
   iac-stage-state:
-    description: Terraform state for stage automation.
     versioning: true
     managed_folders:
       1-vpcsc:
@@ -132,7 +130,6 @@ buckets:
             - $iam_principals:service_accounts/iac-0/iac-pf-ro
   # Terraform state bucket for FAST outputs
   iac-outputs:
-    description: Terraform state for the org-level automation.
     versioning: true
     iam:
       roles/storage.admin:
--- a/fast/stages/0-org-setup/datasets/classic/projects/core/iac-0.yaml
+++ b/fast/stages/0-org-setup/datasets/classic/projects/core/iac-0.yaml
@@ -99,7 +99,6 @@ data_access_logs:
 buckets:
   # Terraform state bucket for this stage
   iac-org-state:
-    description: Terraform state for the org-level automation.
     versioning: true
     iam:
       roles/storage.admin:
@@ -108,7 +107,6 @@ buckets:
         - $iam_principals:service_accounts/iac-0/iac-org-ro
   # Terraform state bucket for additional FAST stages
   iac-stage-state:
-    description: Terraform state for stage automation.
     versioning: true
     managed_folders:
       1-vpcsc:
@@ -137,7 +135,6 @@ buckets:
             - $iam_principals:service_accounts/iac-0/iac-pf-ro
   # Terraform state bucket for FAST outputs
   iac-outputs:
-    description: Terraform state for the org-level automation.
     versioning: true
     iam:
       roles/storage.admin:
--- a/fast/stages/0-org-setup/datasets/hardened/projects/core/iac-0.yaml
+++ b/fast/stages/0-org-setup/datasets/hardened/projects/core/iac-0.yaml
@@ -132,7 +132,6 @@ org_policies:
 buckets:
   # Terraform state bucket for this stage
   iac-org-state:
-    description: Terraform state for the org-level automation.
     versioning: true
     encryption_key: $kms_keys:iac-0/ew1/storage
     lifecycle_rules:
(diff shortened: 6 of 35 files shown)


sruffilli:
> ### Automated PR Review 🤖
> Please update `modules/project-factory/variables-projects.tf` to include the new attributes for `log_buckets`, `automation.bucket`, and `iam_by_principals` in `buckets`.

Done - and went beyond.

> ### 2. Optional: `folder-factory` Schema
> I noticed that `folder.schema.json` was updated in the FAST stages and in `project-factory`. If the `modules/folder-factory` module also manages buckets, you might want to check if its schema (`modules/folder-factory/schemas/folder.schema.json`) and variables need the same `description` removal and `public_access_prevention` addition. This is optional and can be done in a separate PR if preferred.

Hallucination

juliocc (approved):

PR #4158, opened 2026-09-10
Add deletion_policy support to modules/pubsub subscriptions

nathancolhoun-signifyd:
## What

Exposes the `google_pubsub_subscription` `deletion_policy` argument (`DELETE` / `PREVENT` / `ABANDON`) as an optional field on the `subscriptions` variable, and passes it through to the resource.

## Why

Subscription `filter` is immutable in GCP, so editing one forces a destroy/create. Because a subscription's IAM policy is a property *of* the subscription, GCP deletes the policy along with it — while Terraform plans no change to this module's IAM bindings, since they identify the subscription by a config-deterministic name that a filter edit doesn't change.

The net effect is a subscription that comes back with an empty IAM policy, no error surfaced, and state still asserting the bindings are healthy. Convergence needs a second apply. We hit this in production: a one-line filter change silently removed a subscriber's `roles/pubsub.subscriber` binding and the consumer crash-looped on `PERMISSION_DENIED`.

`deletion_policy = "PREVENT"` lets callers make that destroy fail loudly instead of silently succeeding.

A `lifecycle` block would also address it, but `deletion_policy` is preferable as a module interface: it's a resource *argument*, so it can be set per subscription by the caller, whereas `prevent_destroy` and `replace_triggered_by` are meta-arguments that can only be hardcoded inside the module.

## Notes

- Purely additive and optional; defaults to `null`, so existing behaviour is unchanged.
- `deletion_policy` is available from provider `v7.40.0`, satisfying this module's `>= 7.40.0, < 8.0.0` constraint.
- `tools/tfdoc.py modules/pubsub` produces no diff — `subscriptions` is the last variable in the file so its line anchor is unchanged, and complex types render collapsed.
- `pytest tests/modules/pubsub` passes.
- I deliberately did **not** add `PREVENT` to a README example: the Subscriptions example is tagged `e2e`, and a `PREVENT` subscription would leave the e2e run unable to tear itself down. Happy to add an example using `DELETE`, or a non-e2e example, if you'd prefer one for documentation.

### DIFF
--- a/fast/stages/0-org-setup/schemas/project.schema.json
+++ b/fast/stages/0-org-setup/schemas/project.schema.json
@@ -2531,6 +2531,14 @@
                 "ack_deadline_seconds": {
                   "type": "number"
                 },
+                "deletion_policy": {
+                  "type": "string",
+                  "enum": [
+                    "ABANDON",
+                    "DELETE",
+                    "PREVENT"
+                  ]
+                },
                 "enable_exactly_once_delivery": {
                   "type": "boolean"
                 },
--- a/fast/stages/0-org-setup/schemas/project.schema.md
+++ b/fast/stages/0-org-setup/schemas/project.schema.md
@@ -734,6 +734,8 @@
     - **`^[a-zA-Z0-9_-]+$`**: *object*
       <br>*additional properties: false*
       - **ack_deadline_seconds**: *number*
+      - **deletion_policy**: *string*
+        <br>*enum: ['ABANDON', 'DELETE', 'PREVENT']*
       - **enable_exactly_once_delivery**: *boolean*
       - **enable_message_ordering**: *boolean*
       - **expiration_policy_ttl**: *string*
--- a/fast/stages/2-networking/schemas/project.schema.json
+++ b/fast/stages/2-networking/schemas/project.schema.json
@@ -2531,6 +2531,14 @@
                 "ack_deadline_seconds": {
                   "type": "number"
                 },
+                "deletion_policy": {
+                  "type": "string",
+                  "enum": [
+                    "ABANDON",
+                    "DELETE",
+                    "PREVENT"
+                  ]
+                },
                 "enable_exactly_once_delivery": {
                   "type": "boolean"
                 },
--- a/fast/stages/2-networking/schemas/project.schema.md
+++ b/fast/stages/2-networking/schemas/project.schema.md
@@ -734,6 +734,8 @@
     - **`^[a-zA-Z0-9_-]+$`**: *object*
       <br>*additional properties: false*
       - **ack_deadline_seconds**: *number*
+      - **deletion_policy**: *string*
+        <br>*enum: ['ABANDON', 'DELETE', 'PREVENT']*
       - **enable_exactly_once_delivery**: *boolean*
       - **enable_message_ordering**: *boolean*
       - **expiration_policy_ttl**: *string*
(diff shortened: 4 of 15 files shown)


sruffilli (dismissed):
Thanks for the PR! A few things to address:

## 1. Add a validation block for the enum

Our convention is that variables mirroring a provider enum carry a `validation` block; see `modules/project`, `modules/firestore` and `modules/secret-manager`, which all validate their own `deletion_policy`.

This case is stronger than the usual convention argument, because a wrong value silently degrades to `DELETE` for this resource (!!).

## 2. Propagate the change to project-factory and its schemas

`modules/project-factory` re-declares this variable surface and consumes `modules/pubsub`, so as it stands the new attribute is unreachable from the project factory, and factory YAML using it fails schema validation because the subscription object is `"additionalProperties": false`. Please also update:

- `modules/project-factory/variables-projects.tf`: add `deletion_policy = optional(string)` to the `pubsub_topics.*.subscriptions` object (around L561).
- `modules/project-factory/schemas/project.schema.json`: add `deletion_policy` to `$defs.pubsub_topic.properties.subscriptions`, ideally with an `enum`.
- That schema is a duplicate group in `tools/duplicate-diff.py`, so the same edit must be mirrored in `fast/stages/{0-org-setup,2-networking,2-project-factory,2-security}/schemas/project.schema.json`, or CI will fail on the duplicate check.
- Regenerate the schema docs afterwards with `uv run tools/schema_docs.py`.

## 3. Add test coverage for the new attribute

Nothing currently exercises the field. The easiest way is to use the new attributes in one of the README.md examples.

nathancolhoun-signifyd:
Thanks for the thorough review — all three items are addressed.

**1. Validation block.** Added to the `subscriptions` variable, following the `modules/secret-manager` idiom since this is also a map-nested field:

```hcl
validation {
  condition = alltrue([
    for k, v in var.subscriptions : contains(
      ["ABANDON", "DELETE", "PREVENT"], coalesce(v.deletion_policy, "DELETE")
    )
  ])
  error_message = "Subscription deletion policy must be one of 'ABANDON', 'DELETE' or 'PREVENT'."
}
```

Your point about a wrong value silently degrading to `DELETE` is exactly why this matters more than convention here — that failure mode is invisible. Verified both directions: `"NOPE"` now fails at plan time with the error above, `"PREVENT"` validates clean.

**2. project-factory and schemas.** Propagated:

- `modules/project-factory/variables-projects.tf` — `deletion_policy = optional(string)` on the `pubsub_topics.*.subscriptions` object
- `modules/project-factory/schemas/project.schema.json` — added to `$defs.pubsub_topic.properties.subscriptions` with an `enum` of the three values
- Mirrored to all four `fast/stages/{0-org-setup,2-networking,2-project-factory,2-security}/schemas/project.schema.json`; `tools/duplicate-diff.py` passes clean
- Schema docs regenerated with `tools/schema_docs.py` (the five `project.schema.md` files)

**3. Test coverage.** Added a *Subscription deletion protection* section to the module README with a worked example, and generated its inventory via `tools/generate_plan_summary.py`. The inventory records `deletion_policy: PREVENT` on the planned subscription, so the attribute is genuinely exercised rather than just declared.

One deliberate choice worth flagging: **that example is not tagged `e2e`.** A `PREVENT` subscription cannot be destroyed, so an end-to-end run would be unable to tear itself down and would leak the resource. It is still plan-tested against the inventory. If you would rather have e2e coverage, I am happy to switch the example to `DELETE` and add a separate non-e2e block for `PREVENT` — just let me know which you prefer.

Local checks: `terraform fmt` clean, `tools/duplicate-diff.py` clean, `pytest tests/modules/pubsub` 1 passed, `pytest tests/schemas` 1 passed, `pytest tests/examples -k project-factory` 17 passed.

juliocc (approved):

nathancolhoun-signifyd:
@sruffilli can you review recent changes please?

PR #4166, opened 2026-09-11
Add support for reading secret in WFIF pool from a local file

kovagoadam:
Add support for reading secret in WFIF pool from a local file
---
**Checklist**


ludoo (commented):
I don't think we should do it, as file would then need to always be present AFAIR, we should wait for a write only attribute and do the same thing we do in the secret manager module

juliocc:
reading from a file doesn't fix anything. The real fix is write only attributes, which is part why I wanted to upgrade the provider. WFIF pools now have support for a write only field for the client secret

PR #3692, opened 2026-02-01
Use `LEGACY_DATAPATH` when `enable_features.dataplane_v2` is false

singhal0306:
---
**Checklist**

I applicable, I acknowledge that I have:
- [x] Read the [contributing guide](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md)
- [x] Ran `terraform fmt` on all modified files
- [x] Regenerated the relevant README.md files using [`tools/tfdoc.py`](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric/blob/master/CONTRIBUTING.md#fabric-tools)
- [x] Made sure all relevant tests pass

PR #4171, opened 2026-09-14
Add datapath_provider override to gke-cluster-standard

iAbhishek91:
## Summary

PR #3692 ("Use `LEGACY_DATAPATH` when `enable_features.dataplane_v2` is
false", merged 2026-02-02, first released in v52.1.0) changed the
`datapath_provider` value written for clusters with
`enable_features.dataplane_v2 = false`:

```diff
   datapath_provider = (
     var.enable_features.dataplane_v2
     ? "ADVANCED_DATAPATH"
-    : "DATAPATH_PROVIDER_UNSPECIFIED"
+    : "LEGACY_DATAPATH"
   )
```

**This is not backward compatible.** `datapath_provider` is a ForceNew
field on `google_container_cluster`. Any cluster provisioned before
#3692 with `dataplane_v2 = false` has `DATAPATH_PROVIDER_UNSPECIFIED`
in state. After upgrading past v52.1.0, Terraform plans a change to
`LEGACY_DATAPATH` for that field, which forces the cluster to be
destroyed and recreated. #3692 shipped with no upgrade note and no
migration path, so this currently blocks upgrading the module for any
consumer with `dataplane_v2` explicitly disabled.

Clusters that never set `dataplane_v2` (the default is `true`, i.e.
`ADVANCED_DATAPATH`) are unaffected.

## What this change does

Adds a new optional field, `enable_features.datapath_provider`
(string, default `null`), that overrides the value derived from
`dataplane_v2`:

```hcl
datapath_provider = coalesce(
  var.enable_features.datapath_provider,
  var.enable_features.dataplane_v2 ? "ADVANCED_DATAPATH" : "LEGACY_DATAPATH"
)
```

- Default behavior (field unset) is unchanged for every existing user,
  including the #3692 behavior for new clusters.
- Consumers hitting the forced-recreate issue can set
  `enable_features.datapath_provider = "DATAPATH_PROVIDER_UNSPECIFIED"`
  to pin the pre-#3692 value and upgrade the module without a
  disruptive recreate.
- A validation restricts the field to the three valid API enum values
  (`DATAPATH_PROVIDER_UNSPECIFIED`, `LEGACY_DATAPATH`,
  `ADVANCED_DATAPATH`) or `null`.

## Test plan

- [x] `terraform validate` passes for `modules/gke-cluster-standard`
- [x] `terraform fmt` clean
- [x] README regenerated via `tools/tfdoc.py`
- [ ] Verify on a cluster with `dataplane_v2 = false` that setting
      `datapath_provider = "DATAPATH_PROVIDER_UNSPECIFIED"` produces a
      clean plan (no forced replacement) against existing state


juliocc:
Perhaps it's easier if we revert #3692? Usually we don't use or allow _UNSPECIFIED values for enums. You can usually achieve the same effect setting the field to null.

@singhal0306, what was the reasoning behind #3692?

iAbhishek91:
I completely agree with you @juliocc . @singhal0306 if you could please help us revert or make it backward compatible it would be really appreciated. Currently we are kind of soft blocked on upgrade :(

juliocc:
Actually #3692 is correct and I really don't like introducing a second field to bypass dataplane_v2

@ludoo @sruffilli wdyt?

iAbhishek91:
@juliocc Thanks, but the question is how do we deal with this, its a breaking change for anyone not using `dataplane_v2`. Basically we have two options, recreate all production clusters or don't upgrade. Is there something else we could do, maybe? please suggest.

iAbhishek91:
@juliocc @ludoo @sruffilli any suggestions on the above please? Thanks. We are kind of soft block if there is a way to get it backward compatible.

singhal0306:
> ```diff
> DATAPATH_PROVIDER_UNSPECIFIED
> ```

@iAbhishek91, I wasn’t aware of when DATAPATH_PROVIDER_UNSPECIFIED is used. Since I started working on the Terraform documentation and the Fabric repository, I’ve only come across two options: LEGACY_DATAPATH and ADVANCED_DATAPATH as per terraform official documentation. Based on that understanding, I raised the PR, especially since the change was affecting my cluster.

For now, I have a plan that should help unblock this. I would request you to maintain a local fork of the Fabric repository and push your changes there. You can keep rebasing your fork with the upstream Fabric repository so that you stay in sync.

This way, you’ll have whatever additional changes you need on your side without directly blocking or altering the upstream Fabric repository. Whenever you feel the changes are appropriate for upstream, you can raise a PR against the Fabric repository.

singhal0306:
Additionally, I apologize for getting involved this late. I was occupied with some very important work at my organization, specifically preparing for and conducting a DR drill to replicate the complete infrastructure to the DR site as part of an active-passive DR plan.

iAbhishek91:
@singhal0306 appreciate that, but that doesn't help. Maintaining local fork is not a help its pushing maintenance overhead of this code because we merged something which isn't backward compatible. I would classify this as a bug as it breaks existing system and try to either revert or fix it.

Also I don't understand why @juliocc says "I really don't like introducing a second field to bypass dataplane_v2" there is not much technical explanation. 

I guess I cant do much here, I will probably close this MR and manage it ourselves. Thanks everyone!

singhal0306:
> @singhal0306 appreciate that, but that doesn't help. Maintaining local fork is not a help its pushing maintenance overhead of this code because we merged something which isn't backward compatible. I would classify this as a bug as it breaks existing system and try to either revert or fix it.
> 
> Also I don't understand why @juliocc says "I really don't like introducing a second field to bypass dataplane_v2" there is not much technical explanation.
> 
> I guess I cant do much here, I will probably close this MR and manage it ourselves. Thanks everyone!

As far as I understand the thing is that the Terraform resource doesnot accept `DATAPATH_PROVIDER_UNSPECIFIED` in  datapath_provider attrubute anymore.

singhal0306:
@iAbhishek91 I understand your problem now..
image

As soon i explored the providers terraform repo https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/container/resource_container_cluster.go 

The value is still acceptable in the cluster resource DATAPATH_PROVIDER_UNSPECIFIED, I just believed on the terraform documentation by the provider itself..

image

iAbhishek91:
I would consider Google Docs as source of truth https://docs.cloud.google.com/php/docs/reference/cloud-container/latest/V1.DatapathProvider and clearly it contradicts terraform docs. Although the API under the hood works as per Google Docs.

